PR #22, opened 2017-03-16
add a code of conduct

kayiwa:
If we will be collaborating with other organizations I'd like to have this set
up from the get go.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


kayiwa:
since we anticipate working with other orgs I feel we need to add this.

escowles:
👍  we should talk about how we can collaborate with other people — I know there are a few other folks working on Ansible playbooks for Hydra, so it would be great to work together where it makes sense.

kayiwa:
This is one of my regrets of missing C4L. There are many people doing Ansible playbooks/roles and I could have done a B0F there. I know Alicia for DCE and Hardy from UCLA had one (none of whom will be at LDCX)

PR #27, opened 2017-04-25
Add mount for lib-gissystem

### LINKED ISSUE #26: Create mount on plum for geo data
Create an nfs mount to lib-gissystem:gbl to use a source for initial ingest of scanned maps and vector datasets. See Also: https://github.com/pulibrary/plum/issues/1216

eliotjordan:
Closes #26

### DIFF
--- a/roles/pulibrary.plum/defaults/main.yml
+++ b/roles/pulibrary.plum/defaults/main.yml
@@ -7,3 +7,4 @@ plum_environment: 'production'
 plum_fedora_mount: 'hydra'
 simple_tiles_version: 0.6.0
 simple_tiles_url: https://github.com/propublica/simple-tiles/archive/v{{simple_tiles_version}}.tar.gz
+lib_gissystem_src: 128.112.201.250:/gbl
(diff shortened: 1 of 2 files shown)


eliotjordan:
@kayiwa Does this look like the right way to add a read-only nfs mount to plum and the plum worker boxes? There is no password, and I long ago discovered that I had to use the ip instead of the FQDN to get this to work. Windows 😞 .

kayiwa:
said differently... while the OS would boot without problems if it doesn't resolve an additional

```
mount -t nfs
```
step will be required every time there will be a power on, just to be sure.

eliotjordan:
@kayiwa I think I addressed your concerns with this. If not, let me know. FWIW, I don't expect that this will be a permanent change to the configuration. We only need to mount this share while we import our existing collection of geo objects.

eliotjordan:
@kayiwa ^ I meant to say that i force-pushed some changes to this commit.

kayiwa:
apologies for this. I could have sworn I merged it. :-(

PR #30, opened 2017-05-02
point to PUL Library oracle java 8 role

kayiwa:
this uses PUL's oracle java 8 (does the version of Java matter here?)


escowles:
Any version of Java 8 is fine here — it looks like our role and the existing ansiblebit one both wind up using the same `ppa:webupd8team/java` repository to fetch the Java packages, so the version that actually gets installed should be the same either way.

eliotjordan:
Ok to merge?

PR #32, opened 2017-05-02
Update variables in plum_config template

### LINKED ISSUE #31: Variables in plum_config template are incorrect
{{plum_iiif_url}} should be {{vault_plum_iiif_url}} etc...

https://github.com/pulibrary/princeton_ansible/blob/master/roles/pulibrary.plum/templates/plum_config#L2

eliotjordan:
Closes #31


tpendragon:
These reference these variables on purpose which are then pointed to the vault versions at https://github.com/pulibrary/princeton_ansible/blob/master/group_vars/all/vars.yml#L25-L35. It's an ansible best practice as per http://docs.ansible.com/ansible/playbooks_best_practices.html#variables-and-vaults. @eliotjordan says he got an error message when trying to deploy as-is that I've never seen before.

So, I think we should hold off on this until we know what's going on.

eliotjordan:
Not sure what was happening here. I think we can close this PR.

PR #33, opened 2017-05-03
Graylog server

kayiwa:
This creates a graylog node that can be added to our cluster. The end point will need to be n+1 and completed using the load balancer.


jpstroop:
I can merge, but I can't say I understand everything here. Is there anyone else who might do a better job?

kayiwa:
Asked on #operations channel.

roles/pulibrary.graylog2/defaults/main.yml
```diff
@@ -0,0 +1,146 @@
+---
+# General
+graylog_is_master:            'True'
+graylog_node_id_file:         '/etc/graylog/server/node-id'
+graylog_password_secret:      'somethingmorecomplex'
+graylog_root_username:        'admin'
+graylog_root_password_sha2:   'donotusethisandusesha28415bde908c873fc4ab448a918'
```
tpendragon: Why isn't the real version of this in a vault file?

kayiwa:
@tpendragon running this playbook against an endpoint you enter on the `hosts` file which has the user `pulsys` will install a graylog node. The node will need to be added to the mongodb, elasticsearch, and graylog cluster to work. The ugly details are in the Documentation on the Google Drive. I was able to stand one up on aws with 

```
ansible-playbook --limit 1.2.3.4 graylog.yml -b -k --key-file=/Users/username/.aws/aws_keypair.pem -u ubuntu
```

where 1.2.3.4 is the AWS endpoint.

roles/pulibrary.graylog2/defaults/main.yml
kayiwa: it is now.

jpstroop:
Is this ready?

PR #35, opened 2017-06-08
System user

kayiwa:
We have a couple of roles that expect a user not named `deploy` this change allows us to do so. 

Variables can be passed like so

```
ansible-playbook --limit 1.2.3.5 -e system_user=deploy plum.yml -b
```


kayiwa:
ping? ;-)

PR #39, opened 2017-06-20
Fix deploy user definition

tpendragon:


kayiwa:
so if I try to install a user named "arcgis" or "geoserver" using

```
ansible-playbook --limit 1.2.3.4 geoserver.yml -b
```
this won't work because it will install the `deploy` user that this change adds.

PR #41, opened 2017-06-26
Add samba role and configure for geoserver playbook

### LINKED ISSUE #40: Geoserver should have Samba share that Plum can mount
- [ ] Create smb share that for Plum to mount
- [ ] The directory should either be a directory within the GeoServer data dir or a directory in GeoServer should be symlinked to it.

eliotjordan:
Closes #40


kayiwa:
still unsure (this is definitely a nit) is we want to isolate the `vars` in the geoserver.yml I'd rather see them now or at a future date end up in `group_vars/geoserver.yml`

PR #57, opened 2017-09-07
First pass at DSS playbook.

kevinreiss:
@kayiwa Is this headed in the right direction? I'm not %100 on how to reference the Maria DB cluster in this recipe.


kayiwa:
Hey @kevinreiss look at an example from the archivesspace branch. Install mariadb unless the `db_host` variable is pointing to the cluster. 
https://github.com/pulibrary/princeton_ansible/blob/archivesspace/archivesspace.yml

PR #62, opened 2017-10-10
WIP - Pas

kevinreiss:


Vagrantfile
```diff
@@ -42,7 +42,7 @@ Vagrant.configure(2) do |config|
     # ansible.verbose = 'vvv'
 
     ansible.groups = {
-      "figgy" => ["default"]
+      "vagrant" => ["default"]
```
kayiwa: Not need but we probably want to make work differently

Vagrantfile
```diff
@@ -59,12 +59,12 @@ Vagrant.configure(2) do |config|
...
     }
 
-    ansible.playbook = "figgy.yml"
+    ansible.playbook = "pas.yml"
 
     # update start_at_task and re-run `vagrant provison` if your configuration scripts fail on a particular task
     # and you want to restart the provisioning at the step where the failure occurred
     # ansible.start_at_task = "packages | update apt cache"
-    config.ssh.private_key_path = "pulsys_rsa_key"
+    config.ssh.private_key_path = "~/.ssh/pulsys_rsa_key"
```
kayiwa: this file is not managed by git. No need to change the path. Look at the `.gitignore` file

Vagrantfile
```diff
@@ -59,12 +59,12 @@ Vagrant.configure(2) do |config|
       postgresql_is_local: true
     }
 
-    ansible.playbook = "figgy.yml"
+    ansible.playbook = "pas.yml"
```
kayiwa: not needed... but I've created an issue for this

group_vars/pas.yml
```diff
@@ -0,0 +1,3 @@
+pas_db_name: '{{vault_pas_db_name}}'
```
kayiwa: add the generated password to `group_vars/all/vault.yml`

Vagrantfile
kevinreiss: Not sure what you want here.

roles/pulibrary.pas/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  npm: name=grunt-cli global=yes
+  become: true
+
+- name: install sass
+  gem:
+    name: sass
+    state: latest
+    user_install: no
+
+# Didn't clone https://github.com/pulibrary/pas.git
```
kevinreiss: I couldn't get the clone to work. The repo is pretty large. It hung for a long time on https. When I tried an SSH style URL I got:

```fatal: [default]: FAILED! => {"failed": true, "msg": "to use the 'ssh' connection type with passwords, you must install the sshpass program"}```

Comment (kevinreiss) [reply roles/pulibrary.pas/tasks/main.yml]: It is a private repo. But I would have expected https would have just given me the username/password dialog.

File: Vagrantfile (hunk 1 above)
Comment (kayiwa) [reply Vagrantfile]: leave it as in. No change needed to the file.

File: roles/pulibrary.pas/tasks/main.yml (hunk 5 above)
Comment (kayiwa) [reply roles/pulibrary.pas/tasks/main.yml]: will it always be private? in psuedo code
```
- name: add github ssh key
    copy: >
      src=files/id_rsa.github
      dest=/home/pulsys/.ssh/id_rsa.github
      owner=pulsys
      group=root
      mode=0600
  
  - name: configure ssh to use ansible key for github.com
    template: >
      src=templates/ssh_config.j2
      dest=/home/pulsys/.ssh/config
      owner=pulsys
      group=pulsys
      mode=0644
      
  - name: clone a private repository
    git: >
      repo=ssh://git@github.com/pulibrary/pas.git
      key_file=/home/pulsys/.ssh/id_rsa.github
      dest=/path/to/pas
```

File: roles/pulibrary.pas/defaults/main.yml [hunk 6]
```diff
@@ -4,6 +4,9 @@ pas_db_user: 'pas'
 apache_app_path: '/opt/pas'
 pas_db_name: 'pas'
 pas_db_password: '{{ pas_password }}'
+pas_server_name: '{{ application_server | default("localhost") }}'
```
kevinreiss: @kayiwa The variables are some application level config details that seemed best automate as  part of the build. I son et them to default to the defaults for vagrant development. Apparently the craftcms app needs to be told the server it is on and have two different URLs hardcoded to what the public URL of the site will be.

kevinreiss:
pas mount point fails with error 

```TASK [pulibrary.pas : Create mount for pas shared files] ***********************
failed: [default] (item=pas) => {"failed": true, "item": "pas", "msg": "Error mounting /mnt/pas: mount: wrong fs type, bad option, bad superblock on //diglibdata1.princeton.edu/pas,\n       missing codepage or helper program, or other error\n       (for several filesystems (e.g. nfs, cifs) you might\n       need a /sbin/mount.<type> helper program)\n\n       In some cases useful info is found in syslog - try\n       dmesg | tail or so.\n"}
```

kevinreiss:
@kayiwa the vault addition are a username/password for the pas database. Should those be moved somewhere else and be stored in an encrypted file in the pas role instead?

kayiwa:
The `group_vars/all/vault.yml` file *is* encryted.

kevinreiss:
I know, I was asking if an application level detail like that should reside in a different encrypted file within the role defined for pas rather than the vault file shared by all roles.

kayiwa:
ansible looks at `group_vars/all` first. you are welcome to encrypt and place in the pas role but you will have to enter the password more than once.

kevinreiss:
I see, I was just looking to avoid the merge conflict. It can stay where it is at now if you are okay with it.

kayiwa:
use `ansible_vault-merge.sh` it should help you merge into `vault.yml`

PR #73, opened 2017-10-24
(WIP) ignore Vagrant file

kayiwa:
The Vagrantfile is becoming unwieldy as the projects start to differ.
The documents will now have a symbolic untracked link to files stored in
the `Vagrant` directory

this will be the first of many documenting roles that I will do over the next couple weeks. The biggest change worth looking at is we will place Vagrantfiles in the Vagrant directory and use symbolic links to them. Please hold on merging this *for now* but thanks for letting me know if this works.


escowles:
👍  This should make it easier to have different Vagrantfiles for different projects.

Vagrant/laeVagrantfile
```diff
@@ -39,14 +39,12 @@ Vagrant.configure(2) do |config|
       rails_env: "production",
       application_host: 'localhost:8484',
       passenger_server_name: 'localhost',
-      figgy_host_name: 'localhost',
```
kayiwa: hey @escowles does an lae Vagrantfile need figgy references?

escowles: @kayiwa Yep! It pulls JSON-LD and IIIF manifests from Figgy to do a full reindex.

kayiwa: thanks!

kayiwa:
Thanks for merging this @escowles The big "breaking" change is the need to do a

```
ln -s /path/to/projectVagrantfile /path/toclonedrepo/Vagrantfile
```

kayiwa:
Okay the bug in ansible 2.4 is fixed by this. 

https://github.com/ansible/ansible/pull/31493/commits/0c3975e48430917e79d213bbee252cda17deea19

Thanks for looking into this.

kayiwa:
comments? OK? @escowles thanks for merging.

apache.yml
```diff
@@ -0,0 +1,8 @@
+---
```
escowles: Do we need both `apache.yml` and `apache2.yml`?

kayiwa:
good catch. Fixed that.

PR #85, opened 2017-11-04
this allows remote IPs to show up in logs

kayiwa:
not sure it belongs here because if you uncomment it will break the
ability to use on vagrant set up


kayiwa:
ping @eliotjordan

eliotjordan:
Uncommenting to enable this behavior seems fine to me. I'm happy to merge if you're happy with this.

PR #106, opened 2017-12-14
DPUL Production Ansible-ized.

tpendragon:


Vagrant/dpulVagrantfile
```diff
@@ -0,0 +1,68 @@
...
+
+Vagrant.configure(2) do |config|
+  # Vagrant configuration options are fully documented at
+  # https://docs.vagrantup.com.
+
+  config.vm.box = "princeton_box"
+  config.vm.network "private_network", type: "dhcp"
+  config.vm.network "forwarded_port", guest: 80, host: 8484 # apache web server in production mode
+  config.vm.network "forwarded_port", guest: 8983, host: 9893  # solr
+  config.vm.network "forwarded_port", guest: 8080, host: 2424 # tomcat (fedora & solr) in production mode
```
escowles: We're not running Tomcat, right? And we should probably change the comment above to say "nginx" instead of "apache".

PR #112, opened 2017-12-19
Bibdata role

eliotjordan:
- Adds bibdata playbook and role for setting up server running marc_liberation.
- Adds oracle.client role to install Oracle Instant Client for communicating with Voyager database.
- Notable elements
  - Bibata uses a vault specific to the playbook. It does not use the global vault, except for a few shared variables.
  - Staging deployments install a local mariadb instance, while production deployments target the mariadb cluster.
  - READMEs for easy deployment and development.

Closes https://github.com/pulibrary/marc_liberation/issues/282


Vagrant/bibdataVagrantfile
```diff
@@ -0,0 +1,58 @@
...
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  # Vagrant configuration options are fully documented at
+  # https://docs.vagrantup.com.
+
+  config.vm.box = "princeton_box"
+  config.vm.network "private_network", type: "dhcp"
+  config.vm.network "forwarded_port", guest: 80, host: 8484 # apache web server in production mode
+  config.vm.network "forwarded_port", guest: 8983, host: 9893  # solr
```
escowles: Do we need this? I didn't think bibdata used Solr...

eliotjordan: Nope. This is extraneous.

escowles:
This looks good to me, but I wasn't able to get vagrant to spin it up (probably a local issue unrelated to this PR).

eliotjordan:
ping @kayiwa

kayiwa:
let's hope it really doesn't take 3 days to get a license from Oracle!

PR #130, opened 2018-01-08
(WIP) Ezproxy installer

kayiwa:


kevinreiss:
Need to set the hostname value in config.txt to what the public facing ezproxy hostname will be after the file is copied over from the template placed in local_files/ezproxy/config.txt.tmp.

PR #145, opened 2018-01-18
Add workers to bibdata

### LINKED ISSUE #144: Make worker role for bibdata + Updates
- Worker, staging roles need sidekiq
- Worker role needs to share data dir via samba
- Prod role needs to mount data dir via samba

eliotjordan:
Closes #144 

Adds bibdata worker host and shares it's data directory via samba with web servers.


bibdata.yml
```diff
@@ -1,20 +1,59 @@
...
+  vars:
+    bibdata_db_host: localhost
+    passenger_server_name: bibdata-staging.*
+    samba_status: server
   roles:
     - role: pulibrary.ruby
     - role: pulibrary.deploy-user
     - role: pulibrary.passenger
     - role: pulibrary.redis
-    - {role: pulibrary.mariadb, when: bibdata_db_host == 'localhost'}
```
kayiwa: assuming that we never install bibdata locally yes?

eliotjordan: This was there to install mariadb only staging machines. But it's no longer required because there is special config just for staging servers now.

PR #165, opened 2018-02-01
Build a Drupal App Server

kayiwa:
this installs php, composer and drush. I'm a bit out of my depth in my thinking on what would be needed to set up drupal. My thinking currently is

install things as the are with 

php
composer
drush
(then create a generic "hello world" drupal role)
modify the role above with another r/w role that can be public or hidden. 

OK?


kevinreiss:
@kayiwa getting a fatal error using this branch

```
fatal: [default]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "gid": "1001",
            "name": "deploy",
            "state": "present",
            "system": false
        }
    },
    "msg": "groupadd: GID '1001' already exists\n",
    "name": "deploy"
}```

A bit strange since the deploy user task is the first one invoked as far as I can tell in the verbose ansible output.

kayiwa:
hey @kevinreiss on the VM the pulsys user is going to be. When deploy user role was created we didn't have the princeton_box from packer. As a temporary fix assign it a different UID and I will modify the vmimages to create pulsys with 1002

kayiwa:
Hey @kevinreiss pull and try after this commit https://github.com/pulibrary/princeton_ansible/pull/165/commits/7d536ef4ae5f8ad37d44f54a12c96c5951b13780

PR #177, opened 2018-02-15
Updates the default Solr release to 6.6.2 for pulibrary.solr and Figgy and uses an alternative mirror

### LINKED ISSUE #175: Ensure that the latest release of Solr 6.x (6.6.2) is used for the pulibrary.solr Role and the Figgy Playbook
This is in part due to the relative absence of mirrors which offer archival releases earlier than the last stable releases for Solr 6.x.  For example, please see:
http://mirror.cc.columbia.edu/pub/software/apache/lucene/solr/

Additionally, this was originally discovered in response to finding that 
https://lib-solr-mirror.princeton.edu/dist seems to fail to resolve while http://lib-solr-mirror.princeton.edu/dist does.  Further, http://lib-solr-mirror.princeton.edu/dist/lucene/solr/6.2.1/solr-6.2.1.tgz may return a 404 response (possibly due to network firewall restrictions).

jrgriffiniii:
Resolves #175

### DIFF
--- a/roles/pulibrary.solr/defaults/main.yml
+++ b/roles/pulibrary.solr/defaults/main.yml
@@ -4,8 +4,8 @@ solr_workspace: /root
 solr_create_user: true
 solr_user: solr
 
-solr_version: "6.6.0"
-solr_mirror: "https://lib-solr-mirror.princeton.edu/dist"
+solr_version: "6.6.2"
+solr_mirror: "http://mirrors.sonic.net/apache"
 solr_remove_cruft: false
 
 solr_service_manage: true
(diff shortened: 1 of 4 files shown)


roles/pulibrary.solr/defaults/main.yml
```diff
@@ -4,8 +4,8 @@ solr_workspace: /root
 solr_create_user: true
 solr_user: solr
 
-solr_version: "6.6.0"
-solr_mirror: "https://lib-solr-mirror.princeton.edu/dist"
```
kayiwa: just remove the s and it should work no?

kayiwa: http://lib-solr-mirror.princeton.edu/dist/lucene/solr/6.6.2/index.html

hackmastera:
we decided in slack to merge 190 instead. 404 errors should be resolved but do report errors to #operations if they occur.

Issue #186, opened 2018-02-23
Ruby should be built (rather than installed using a system package)


jrgriffiniii:
As proposed by @hackmastera , Ruby should be compiled using a source code distribution and built on the local environment.  The Science History Institute has developed an Ansible Role in response to these needs: https://github.com/curationexperts/ansible-hydra/blob/master/roles/ruby/tasks/main.yml

hackmastera:
well, we paid DCE to do it for us ;)

PR #188, opened 2018-02-27
Adding some additional code comments and using /opt/ruby as the installation directory

jrgriffiniii:
Thank you very much for this @hackmastera - after a review and testing your solution for #186, I can only offer some additional comments, and would please propose that `/opt/install` be changed to `/opt/ruby` for the `install_path`.

Also, is it necessary to ensure that `software-properties-common` be removed?  Most of the included packages appear to be for a system installation of Python, glib, and for CA certificate management:
https://packages.ubuntu.com/xenial/software-properties-common

### DIFF
--- a/roles/pulibrary.ruby/defaults/main.yml
+++ b/roles/pulibrary.ruby/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
-install_path: /opt/install
+install_path: /opt/ruby
 ruby_ver: '2.4.3'
 ruby_major_minor_ver: '2.4'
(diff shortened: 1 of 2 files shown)


hackmastera:
awesome; this looks great. I only think we should leave a note somewhere about why we aren't doing `gem update --system`. I worry that some day we'll wish we could update system gems. but maybe by then we'll be on ruby 4.5 and it will be okay to do it again because we won't need to downgrade bundler?

I will add a note somewhere. thanks for running through this and for the improvements @jrgriffiniii 

I will probably also remove the "downgrade bundler" role.

PR #199, opened 2018-03-13
Ensuring that `bibdata` staging environment is supported

### LINKED ISSUE #191: Update the bibdata Playbook for new environment variables and config. templates
https://github.com/pulibrary/marc_liberation/pull/372 will require that three additional configuration files and one environment variable be handled during the provisioning of `marc_liberation` server environments.

jrgriffiniii:
Resolves #191 

Updating the Playbook and Roles to support bibdata cron tasks (for seeding the ARK cache) and admin. NetIDs


bibdata.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: bibdata
+- hosts: bibdata_staging
```
escowles: Do we want to rename this playbook to `bibdata_staging.yml`?  I'm not sure if we reached agreement on the naming conventions, but I would be in favor of including the environment name in each filename.

jrgriffiniii:
This is prepared for another review.

PR #200, opened 2018-03-13
WIP Postgres10- merge after new hardware is set up

kayiwa:
update to postgres 10 for our db cluster. This will not automatically set up replication (which is not needed by default)


roles/pulibrary.postgresql/templates/pg_hba.conf.j2
```diff
@@ -0,0 +1,59 @@
...
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     md5
+# IPv4 local connections:
+host    all             all             127.0.0.1/32            md5
+# IPv6 local connections:
+host    all             all             ::1/128                 md5
+# figgy connections
+host    all             all             {{ figgy0 }}        md5
```
jrgriffiniii: If connections are host-based and permitted from any hosts using the `md5` trust mode, should `hostssl` be explored as a possible alternative?

group_vars/postgresql/postgresql.yml
```diff
@@ -0,0 +1,32 @@
+---
```
jrgriffiniii: For the `postgresql` Playbook, the location of this file is fine.  However, when testing this, I found that other Playbooks which use the `pulibrary.postgresql` Role failed to access these or Vault-encrypted variables.

roles/pulibrary.postgresql/templates/pg_hba.conf.j2
```diff
@@ -0,0 +1,59 @@
...
+host    all             all             {{ figgy1 }}       md5
```
tpendragon: Can we just do a loop over a variable which is like `postgresql_allowed_hosts` or something, instead of having to edit two files every time we add a host?

group_vars/postgresql/postgresql.yml
tpendragon: Ideally they shouldn't need access to these variables, since they're not in the postgresql group.

PR #219, opened 2018-03-19
DPUL Staging

tpendragon:


ansible-vault-merge
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
```
jrgriffiniii: 👍 I had introduced this locally

PR #232, opened 2018-03-29
Add geoserver worker to pulmap and figgy

### LINKED ISSUE #231: Add geoserver sync worker to pulmap role
Part of https://github.com/pulibrary/figgy/issues/961

### LINKED ISSUE #233: Update figgy vars for geoserver

eliotjordan:
Closes #231 and closes #233


roles/pulibrary.pulmap/meta/main.yml
```diff
@@ -1,3 +0,0 @@
----
-dependencies:
```
jrgriffiniii: This dependency is what ensures that environment variables are set for the `deploy` user.  I would please recommend that this Role be kept, or that `rails-app` be used instead.

eliotjordan: `rails-app` role added.

PR #243, opened 2018-04-11
Fixes SSH `authorized_keys` and the Datadog Role for Orangelight, DPUL, and Figgy Playbooks

jrgriffiniii:


roles/pulibrary.deploy-user/tasks/main.yml
```diff
@@ -30,6 +30,13 @@
               line="for f in ~/app_configs/*; do source $f; done"
               insertbefore=BOF
 - name: Allow deploy user to SSH
+  lineinfile: >
+              dest=/etc/ssh/sshd_config
+              state=present
+              backrefs=yes
+              regexp='^#AuthorizedKeysFile(.*?)$'
+              line="AuthorizedKeysFile\1"
+- name: Allow "authorized_key" files
```
escowles: Shouldn't this name and the one on line 32 be swapped?

PR #265, opened 2018-05-10
Removing worker role from production webservers

escowles:


escowles:
See https://github.com/pulibrary/figgy/issues/1223

PR #274, opened 2018-06-13
[Don't Merge Yet] we are switching from ask pass to file

kayiwa:
decrypted all the vault files and re-encrypted them using a password
file which will be stored as a hidden file in the users home dir

For those on `keybase` contents of `~/.vault_pass.txt` can be shared there.


eliotjordan:
We're waiting on this until we can pass around the vault password file in person.

jrgriffiniii:
I've proposed that this be added to the agenda for the meeting in January.

kayiwa:
I know that @kevinreiss set something up on the schedule @jrgriffiniii

PR #282, opened 2018-06-21
Removes google analytics from orangelight production and staging

christinach:
Removes google analytics from orangelight production and staging.

### DIFF
--- a/group_vars/orangelight.yml
+++ b/group_vars/orangelight.yml
@@ -19,7 +19,6 @@ ol_db_user: '{{ vault_ol_db_user }}'
 ol_db_password: '{{ vault_ol_db_password }}'
 ol_secret_key: '{{vault_ol_secret_key}}'
 ol_solr_url: '{{vault_ol_solr_url}}'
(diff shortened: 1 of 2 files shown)


group_vars/orangelight.yml
```diff
@@ -19,7 +19,6 @@ ol_db_user: '{{ vault_ol_db_user }}'
 ol_db_password: '{{ vault_ol_db_password }}'
 ol_secret_key: '{{vault_ol_secret_key}}'
 ol_solr_url: '{{vault_ol_solr_url}}'
-ol_google_analytics: '{{vault_ol_google_analytics}}'
```
eliotjordan: The `vault_ol_google_analytics` var also needs to be removed from the vault file: https://github.com/pulibrary/princeton_ansible/blob/master/group_vars/all/vault.yml

christinach: We didn't remove it intentionally, in case we want to re-use it.

eliotjordan: https://github.com/pulibrary/princeton_ansible/issues/284

PR #295, opened 2018-07-06
use our our archive mirror

kayiwa:
I think we are also using solr 7.1.0 yes?


escowles:
@kayiwa Yes, we're using 7.1.0 — I just updated several apps today to be in sync with that version, which is what we're using on the lib-solr* machines.

PR #304, opened 2018-08-02
Add yarn role to DSS. 

kevinreiss:


kevinreiss:
@kayiwa could we run this against dss-staging1 if you think it looks okay?

PR #313, opened 2018-09-05
[WIP] Decouples pulibrary.zookeeper role variables from those of pulibrary.solrcloud and resolves warnings for Ansible 2.5.8

jrgriffiniii:
Please note that this was tested using Ansible release 2.5.8 and Python 2.7.15


jrgriffiniii:
Also successfully tested using Ansible release `2.6.3`.  I originally received some warnings related to the creation of temporary directories on the remote hosts, but I could not reproduce these.

PR #314, opened 2018-09-06
Resolves issues for Ansible release 2.6.3 and restructuring the variables for the pulibrary.zookeeper and pulibrary.solrcloud Roles

jrgriffiniii:
Addresses the following:
- Ensuring that the "update host file" task for pul.solrcloud does not depend on hostvars
- Fixing warnings related to Jinja2 templating being used within "when" conditionals
- Ensures that decompressing the Solr installation does not raise any warnings
- Setting Python release 2.7.15 as the preferred version for pyenv
- Decoupling the host groups for Zookeeper from SolrCloud
- Fixing a bug within the myid template which prevented the ensemble server ID from being printed
- Adjusting the Vagrantfile for the pulibrary.zookeeper and pulibrary.solrcloud role variables


Vagrant/solrcloudVagrantfile
```diff
@@ -0,0 +1,98 @@
...
+        }
+    end
+    config.ssh.private_key_path = "pulsys_rsa_key"
+    config.ssh.username = "pulsys"
+  end
+
+  # Create Solr server nodes
+  # https://docs.vagrantup.com/v2/vagrantfile/tips.html
+  (1..1).each do |i|
+    config.vm.define "lib-zk#{i}" do |node|
```
eliotjordan: This section appears to be a duplicate of the zookeeper section above, except with a different comment.

roles/pulibrary.solrcloud/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+## Service options
+
+# Owner
+solr_user: solr
+solr_group: solr
+
+# start on boot
+solr_service_enabled: 'yes'
+# current state: started, stopped
```
eliotjordan: Keep this comment?

roles/pulibrary.solrcloud/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+solr_jetty_response_header_size: 8192
+solr_jetty_send_server_version: 'false'
+solr_jetty_send_date_header: 'false'
+solr_jetty_header_cache_size: 512
+solr_jetty_delay_dispatch_until_content: 'false'
+solr_jetty_http_selectors: -1
+solr_jetty_http_acceptors: -1
+
+# ZOOKEEPER
+zookeeper_hosts: localhost:2181  # Deprecated
```
eliotjordan: Are these deprecated comments to remind us to update in the future, or are we not using these settings?

zookeeper.yml
```diff
@@ -0,0 +1,11 @@
+---
+- hosts: zookeeper
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - site_vars.yml
+  roles:
+    - role: pulibrary.openjdk8
+    - role: pulibrary.samba
```
eliotjordan: None of the variables used to configure samba (samba_shares_root, samba_shares, samba_users, etc...) are setup for these roles. Is this what we want? What are we using samba for here?

roles/pulibrary.solrcloud/defaults/main.yml
jrgriffiniii: I could not determine this myself, and introduced this question in a new issue: https://github.com/pulibrary/princeton_ansible/issues/315

eliotjordan:
One more thing @jrgriffiniii. Do we need samba in the `solrcloud.yml` as well? It's not configured either.

eliotjordan:
@tampakis Says that we might want to create a shared samba mount on each solr server. This requires some setup in the ansible config.

PR #343, opened 2018-09-26
installs php 7.2 for bionic beaver

kayiwa:
defaults to php7.0 for xenial

bionic beaver does not have php7.0 so we default to php7.2. Another PR coming from ansible training.


kayiwa:
already merged!

Issue #345, opened 2018-10-15
Set /proc/sys/vm/overcommit_memory to 1 in bibdata-worker2


christinach:
bibdata-worker2

christinach:
closes https://github.com/pulibrary/marc_liberation/issues/500

PR #346, opened 2018-10-15
Set redis-server overcommit_memory to 1 for bibdata-worker2

christinach:
@kayiwa, would it be a better practice to set the overcommit_memory in the pulibray.redis defaults/main.yml file with a default value and then reference it in any other playbook? If yes, would it be ok to let the default value as 0? From what I've seen so far the servers have it set as 0.


christinach:
closes #345

kayiwa:
OK!

PR #349, opened 2018-10-18
php7 is on EOL

kayiwa:
Software support for PHP7.0 ends in 2018


eliotjordan:
@kayiwa Safe to merge? Safe to deploy?

PR #360, opened 2018-10-23
Update solr urls

eliotjordan:
- Moves OL solr url values out of vault
- Updates OL solr url values


eliotjordan:
Deployed on catalog-staging with good results.

tampakis:
Thanks @eliotjordan!

PR #369, opened 2018-11-12
Ansible awx

kayiwa:
this install the open source version of Ansible Tower


roles/pulibrary.awx/README.md
```diff
@@ -0,0 +1,35 @@
+# Ansible Role: AWX (open source Ansible Tower)
```
jrgriffiniii: 💯 , but I would please suggest linking to the documentation for the default user name and password: https://docs.ansible.com/ansible-tower/latest/html/quickstart/login_superuser.html

kayiwa: Fixed (thank you!) with this commit. https://github.com/pulibrary/princeton_ansible/pull/369/commits/42f48aad9df7dbf2e25faa3400acdced1e6d0c36

PR #370, opened 2018-11-13
Playbooks

kayiwa:
This is from a comment from ansible kindergarten. Having all the playbooks in one directory makes it easier when trying to describe the differences between roles and playbooks. Also makes the repo a tiny bit more tidy.


kayiwa:
Thanks and Nice catch @jrgriffiniii

PR #376, opened 2018-11-20
add webhook intergration

kayiwa:
add an announcement to server-alerts every time an ansible playbook is run.


playbooks/awx.yml
```diff
@@ -9,3 +9,11 @@
   roles:
     - role: ../roles/pulibrary.nodejs
     - role: ../roles/pulibrary.awx
+
+  post_tasks:
+    - name: tell everyone on slack you ran an ansible playbook
+      slack:
```
jrgriffiniii: Are we going to need to specify a required minimum release for Ansible in the README? https://docs.ansible.com/ansible/latest/modules/slack_module.html states that this is currently a part of core.

kayiwa: Our contributions doc `assumes` you are always using the latest stable (I know it isn't always our practice) 
https://github.com/pulibrary/princeton_ansible/blob/master/CONTRIBUTING.md

kayiwa: ping @jrgriffiniii

jrgriffiniii: Sorry, I had overlooked that.  I am still 👍then (that wasn't meant to block approval).

PR #385, opened 2018-11-27
Add a role installing freetds for figgy workers

### LINKED ISSUE #384: add role for freetds
as part of pulibrary/figgy#2139 and pulibrary/figgy#2043

hackmastera:
resolves #384

### DIFF
--- a/playbooks/figgy.yml
+++ b/playbooks/figgy.yml
@@ -7,6 +7,7 @@
   strategy: free
   roles:
     - role: roles/pulibrary.mediainfo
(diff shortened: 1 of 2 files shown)


tpendragon:
@hackmastera Sorry, I messed up. Why isn't this added to the webservers as well as the workers?

hackmastera:
@tpendragon oh you're right it needs to be there to deploy. I was thinking we would never use it from the webservers. will submit follow-up pr!

PR #388, opened 2018-11-28
modify new postgres hosts

kayiwa:
simple update of postgres hosts.


kayiwa:
ping @escowles

escowles:
@kayiwa `postgres3.princeton.edu` doesn't resolve for me — should those be `lib-postgres3.princeton.edu` ?

kayiwa:
sorry about that @escowles

PR #393, opened 2018-12-03
Circleci

kayiwa:
a docker container (for Ubuntu Bionic) that we will use for testing our roles. 

We've added a local connection which will be used to `ansible-playbook --limit circleci` This PR only has a simple date command however.


hackmastera:
Is there an issue to reference here?

I don't really understand how circle ci runs on localhost.

An outline of the proposed architecture would be very helpful!

kayiwa:
I agree @hackmastera will work on that. FWIW `ansible-playbook` runs on localhost

escowles:
I've pushed a commit to fix the naming issue @jrgriffiniii — is this ready to merge?

escowles:
Or do we want to add at least some basic sanity testing of the playbooks and roles, e.g., https://pypi.org/project/ansible-test/ ?

PR #407, opened 2018-12-10
Ensures that redis can be installed on Ubuntu 18.04 by enabling IPv6

### LINKED ISSUE #406: pulibrary.redis Role fails for Ubuntu 18.04 LTS environments
Currently the following errors are logged during the installation:

```
Setting up redis-server (5:4.0.9-1ubuntu0.1) ...
Job for redis-server.service failed because a timeout was exceeded
```

jrgriffiniii:
Resolves #406

### DIFF
--- a/roles/pulibrary.redis/tasks/main.yml
+++ b/roles/pulibrary.redis/tasks/main.yml
@@ -1,4 +1,19 @@
 ---
+# Ensure that IPv6 is enabled for the redis server
+# This will remain necessary until the resolution for https://github.com/antirez/redis/issues/3894 is available upstream
+- sysctl:
+    name: net.ipv6.conf.all.disable_ipv6
+    value: 0
+    state: present
+- sysctl:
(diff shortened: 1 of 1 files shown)


kayiwa:
this is a strange one @jrgriffiniii I think it will be OK but IPv6 is disabled because it is "considered bad network citizenship" on princeton's network.

kayiwa:
See here. https://github.com/pulibrary/vmimages/blob/5fec27b8c65cf31c2a4d8f7dee86cc683dc0461c/ansible/roles/lockdown/tasks/main.yml#L45

jrgriffiniii:
https://github.com/pulibrary/vmimages/pull/31 resolves this, but if this becomes a security concern over IPv6 usage, then perhaps another solution should be offered.

bearnxx:
@jrgriffiniii  how to  resolve the problem ?  the link is all 404 at this page.

PR #410, opened 2018-12-12
playbook that provisions figgy staging

kayiwa:
this is an attempt to fix the notify on slack feature that gives us a
simpler playbook for figgy staging


playbooks/figgy_staging.yml
```diff
@@ -0,0 +1,31 @@
+---
+- hosts: figgy
```
hackmastera: I think we should change it in hosts, too

playbooks/figgy_staging.yml
```diff
@@ -0,0 +1,31 @@
+---
```
hackmastera: delete all this from the figgy playbook

jrgriffiniii:
See #401

playbooks/figgy_staging.yml
hackmastera: rename that playbook `figgy_production`

jrgriffiniii:
I'm still getting the same error:
```
fatal: [figgy-staging1.princeton.edu -> localhost]: FAILED! => {"changed": false, "module_stderr":"sudo: a password is required\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
```

This might be due some issue in my local env.

jrgriffiniii:
I've found the problem, this is the offending line: https://github.com/pulibrary/princeton_ansible/pull/410/files#diff-7c1890847b60b189fec0fed7e1faec44R33

Attempting to escalate privileges on `localhost` requires `sudo` with a password depending upon the local environment where `ansible-playbook` is being invoked.

PR #413, opened 2018-12-13
Remove unused play from figgy_production playbook

### LINKED ISSUE #412: Delete first play in figgy_production.yml
we never run all the machines together, and running the whole playbook would cause each machine to be provisioned twice

hackmastera:
fixes #412


jrgriffiniii:
I believe that `figgy_production_webserver` will be the only play which explicitly disambiguates itself from worker environments by its name.

PR #433, opened 2019-01-23
Oraclejava

kayiwa:
testing this on travis has been a bear when it comes to determining what processor to use. Will continue to investigate our to add integration tests for java.


kayiwa:
ping! OK?

roles/pulibrary.openjdk8/molecule/default/molecule.yml
```diff
@@ -0,0 +1,20 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+lint:
+  name: yamllint
+platforms:
+  - name: instance
+    image: centos:7
```
jrgriffiniii: Why is the CentOS image being used here?

roles/pulibrary.oracle.java8/molecule/default/molecule.yml
```diff
@@ -0,0 +1,20 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+lint:
+  name: yamllint
+platforms:
+  - name: instance
+    image: centos:7
```
jrgriffiniii: Please see the above question regarding CentOS

jrgriffiniii:
Sorry for the delay.  This is excellent, I only had two questions regarding the usage of CentOS Docker images.

Also, would adding any server-specific JDK configuration options be possible a future PR?  If so, I can create an issue following merging this.

roles/pulibrary.oracle.java8/molecule/default/molecule.yml
kayiwa: the way the repo is set up, we never really use this. we rely `molecule/default/molecule.yml` from the repo root.

PR #444, opened 2019-01-31
Oracle client rewrite

kayiwa:
there's a "vaulted" file here. While I've written tests I can't think of a way that works easily with the `tnsnames.ora` file. Thanks for merging this.


kayiwa:
this also updates the password of the vaulted file.

PR #447, opened 2019-02-04
Remove old studio mount from Figgy.

tpendragon:


tpendragon:
The tests failed, but there's no way I caused it. @kayiwa ?

PR #456, opened 2019-02-26
Changes the ENV variables and Rails app. dependencies for the cicognara staging and production Playbooks

jrgriffiniii:
This is issued in response to https://github.com/pulibrary/cicognara-rails/blob/master/config/blacklight.yml#L21


jrgriffiniii:
In order to restrict the number of PRs I issue in order to fix any deployment issues, I'd like to propose to leave this in progress until I've tested this also against the new production server environment.

jrgriffiniii:
This also now addresses additional Rails app. system package dependencies for the server environments, as well as resolves bugs found in the installation and configuration of passenger for NGINX in Ubuntu 18.04 LTS environments.

PR #465, opened 2019-03-28
Nginxplus

kayiwa:
this role conflicts with `pulibrary.apache` for port 80 (inside the docker container) Can you think of a way that these can co-exist @jrgriffiniii ?


escowles:
@kayiwa Can you make the `pulibrary.apache` port configurable, so we can make it not conflict in the docker container, but still be port 80 on production machines?

kayiwa:
> @kayiwa Can you make the `pulibrary.apache` port configurable, so we can make it not conflict in the docker container, but still be port 80 on production machines?

I spent a good amount of time thinking of this before I submitted the PR and I was hoping @jrgriffiniii  will have an instinct that I don't @escowles I have a this fix [here](https://github.com/pulibrary/princeton_ansible/commit/a7939375603ebb0a04c5dd781d17f3fdfd7b3365). It seems rare that this role will be used (for now) to install upstream nginx.

roles/pulibrary.nginxplus/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+galaxy_info:
+  role_name: pulibrary.composer
```
jrgriffiniii: Is this identifying the right Role for the Galaxy Info.?

roles/pulibrary.nginxplus/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+galaxy_info:
+  role_name: pulibrary.nginxplud
```
jrgriffiniii: `s/nginxplud/nginxplus/`

jrgriffiniii:
Travis CI has indicated that there were failures, but https://travis-ci.org/pulibrary/princeton_ansible/builds/527777696?utm_source=github_status&utm_medium=notification leaves me to suspect that this shouldn't introduce any breaking changes.

PR #469, opened 2019-04-09
[WIP] Add subversion client to bibdata-worker

escowles:
* Adds Subversion client needed for accessing PULFA EADs


eliotjordan:
@escowles Is this still relevant? I just rebased against master.

escowles:
@eliotjordan No, we decided on a different strategy (and changed our minds a couple more times after that!).  I'm going to close this.

PR #474, opened 2019-04-10
Lib solr staging

tampakis:
Configures solr cloud staging cluster. Makes solr and zookeeper host names configurable as variables (still assuming that there are three of each).
Also adds "mnt" tag for the network file share backup mount. Tasks related to the backup mount need to be skipped on staging.


playbooks/solrcloud.yml
```diff
@@ -4,6 +4,8 @@
   become: true
   vars_files:
     - ../site_vars.yml
+    - ../group_vars/solrcloud.yml
```
tampakis: I'm not sure why I needed to include this here, but the playbook didn't work without it

escowles: That's weird — the only things in that file that I see are DataDog-related.

escowles: Oh, I was confused between production and staging.  Still weird that you have to include the file, but it's not mixed up between production and staging at least.

PR #475, opened 2019-04-10
Figgy Preservation Configuration

tpendragon:


group_vars/figgy_production.yml
```diff
@@ -121,6 +121,10 @@ rails_app_vars:
     value: '{{figgy_svn_pass}}'
   - name: SVN_URL
     value: '{{figgy_svn_url}}'
+  - name: STORAGE_CREDENTIALS
+    value: '/home/{{deploy_user}}/google_cloud_credentials.json'
```
escowles: Do we want to create a directory to hold required config files, either in the deploy home dir (like `app_configs`) or somewhere else?  My main concern with having this file directly in the home directory is that someone might not realize it was the live config file, and edit/move/delete it.

escowles:
👍 This looks good to me, other than one question about where the config file should live.

group_vars/figgy_production.yml
tpendragon: I can make a directory, I just avoided it so I wouldn't have to make a directory as part of the playbook.

tpendragon:
@escowles Moved it over to a `cloud_config` subdirectory.

PR #479, opened 2019-04-12
Add a readonly user to figgy db

hackmastera:
advances pulibrary/figgy#2830


hackmastera:
Successful test on staging

PR #496, opened 2019-04-29
[WIP] Implements a Playbook for provisioning AtoM installations

jrgriffiniii:
This includes a local installation of MariaDB server and ElasticSearch.

PR #498, opened 2019-04-30
Implements a role for AtoM Gearman workers

jrgriffiniii:
Blocked by #496


jrgriffiniii:
We are no longer interested in exploring AtoM.

PR #499, opened 2019-04-30
Fixing mariadb role

carolyncole:
We made this change during linting, but did not realize it broke 'molecule test'

This PR just reverts the change so that `molecule test` is green

## Error without the fix
Screen Shot 2019-04-30 at 8 13 41 AM

## Passing `molecule test` with the fix
Screen Shot 2019-04-30 at 8 27 24 AM

### DIFF
--- a/roles/pulibrary.mariadbserver/tasks/main.yml
+++ b/roles/pulibrary.mariadbserver/tasks/main.yml
@@ -3,7 +3,11 @@
 - name: Check if database server is installed
   environment:
     LC_MESSAGES: 'C'
-  command: dpkg-query -W -f='${Version}\n' 'mariadb-server' | grep -v '^$'
+  shell: |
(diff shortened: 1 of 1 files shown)


kayiwa:
I got this error
```
 Examining /Users/kayiwa/Documents/dev/pulibrary/princeton_ansible/molecule/default/roles/pulibrary.openjdk8/meta/main.yml of type meta
    [306] Shells that use pipes should set the pipefail option
    /Users/kayiwa/Documents/dev/pulibrary/princeton_ansible/molecule/default/roles/pulibrary.mariadbserver/tasks/main.yml:3
    Task/Handler: Check if database server is installed
```

hackmastera:
This fix does work as intended for the targeted role, so I'm gonna go ahead and merge

PR #503, opened 2019-04-30
Remove steps that only run for Vagrant.

tpendragon:
They complicate the playbooks and sometimes the checks break, resulting
in things (like solr!) being installed on production boxes that
shouldn't be.


hackmastera:
I wanted to see if it would still run on vagrant but now I'm getting a failure on the java role. so idk! Did you try this on vagrant?

PR #509, opened 2019-05-10
adds orangelight worker machine playbook

tampakis:
I ran the playbook against lib-pulindexer3. Should I wait until 1/2 are upgraded to bionic before running there @kayiwa ?


kayiwa:
yes @tampakis Sorry (I'd turned off -accidentally- notifications.

PR #513, opened 2019-05-14
add catalog staging SSL certs

kayiwa:


kayiwa:
this just adds the certs for catalog-staging.

PR #515, opened 2019-05-15
Update Readme to fix role creation instructions

carolyncole:
There was a typo in the instructions, which I found when I tried to run them.
Additionally two steps were missing for setting up the new role to run with the global molecule build.


carolyncole:
Need to fix recomend to.

carolyncole:
@kayiwa this now includes the change for the spelling.

PR #517, opened 2019-05-20
Adding user creation in mariadb

carolyncole:
This creates a database and then creates users for the specified database.
Adds test for database existing with user permissions


kayiwa:
sorry for the delay @carolyncole I was taken away until now.

PR #518, opened 2019-05-21
Adding in remote host information

carolyncole:
The two tasks were running locally, but not on the remote server.  Added login_password and login_host to fix the issue


carolyncole:
@kayiwa I noticed the tests are not testing the options I really think they are.

kayiwa:
gah. I shouldn't work before coffee

PR #519, opened 2019-05-21
steps to add new hosts

kayiwa:
these are the steps needed to add a new virtual server to NGINX Plus.

### DIFF
--- a/roles/pulibrary.nginxplus/ADDHOSTS.md
+++ b/roles/pulibrary.nginxplus/ADDHOSTS.md
@@ -6,15 +6,13 @@ Adding new hosts to Nginx Plus
 ping the #Operations Slack channel. The turn around is instant + time for DNS to
 propagate worldwide.
 
-* Generate SSL keys for the Virtual Server above. This is also another bug that
+* What we will need is the name of the new service. If the service is to be
+  example.princeton.edu, we will then register example-staging1.princeton.edu
+  and example-prod1.princeton.edu, example-prod2.princeton.edu
+
+* Where necessary generate SSL keys for the Virtual Server above. This is also another bug that
   will be fixed with a ping to the #Operations Slack channel. The turn around
   here averages about 12 hours.
-
-* You will need at least one upstream server (for staging) and we encourage at
-  least 2 for production that will host your application. This is the last bug
-  that will still involve the #Operation channel. Registration for a fixed lease
-  IP address is done through the network registration link in step one. The turn
-  around here is also instant.
 
 * Create your nginx configuration files for your service. Examples can be found
   at `roles/pulibrary.nginxplus/files/conf/http` for HTTP based services and
@@ -31,9 +29,3 @@ propagate worldwide.
 
 * Tunnel into the production ADC on port 8080 to see if your new virtual server
   is available
-
-
-[TODO]
-
-* check datadog for upstream services. We have a ticket with both datadog and
-  nginx plus on why the upstream services aren't showing up in our metrics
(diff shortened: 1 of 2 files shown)


roles/pulibrary.nginxplus/ADDHOSTS.md
```diff
@@ -0,0 +1,39 @@
...
+  at `roles/pulibrary.nginxplus/files/conf/http` for HTTP based services and
+  `roles/pulibrary.nginxplus/files/conf/stream` for all others. You will need to
+  use `ansible-vault` to protect any file that will be placed here.
+
+* When the ssl files are received from OIT they will be placed under
+  `roles/pulibrary.nginxplus/files/ssl` and will also need to be protected.
+
+* Once you have your application ready you can then re-run the
+  `playbooks/nginxplus.yml` playbook which will upload the new configuration
+  files and ssl certificates to the production ADC and the hot-standy one and
```
carolyncole: @kayiwa is `hot-standy` the real name?

roles/pulibrary.nginxplus/ADDHOSTS.md
```diff
@@ -0,0 +1,39 @@
+Adding new hosts to Nginx Plus
+==============================
+
+* You will need to register the new Virtual Server at
+  [https://networkregistration.princeton.edu](https://networkregistration.princeton.edu) - for now this is a bug because permission/authority is required from OIT, so
+ping the #Operations Slack channel. The turn around is instant + time for DNS to
```
carolyncole: This statement is a bit confusing to me.  I believe this is two steps 1. register at https://networkregistration.princeton.edu and 2. ping the library slack channel.  I assume we will always have to register the server?  Maybe it would help if there was an example of the information you wanted in the slack channel.

roles/pulibrary.nginxplus/ADDHOSTS.md
```diff
@@ -0,0 +1,39 @@
...
+propagate worldwide.
+
+* Generate SSL keys for the Virtual Server above. This is also another bug that
```
carolyncole: Is this a manual process and the slack ping is the only step.  In the future there will be a non manual step?

roles/pulibrary.nginxplus/ADDHOSTS.md
```diff
@@ -0,0 +1,39 @@
...
+  will be fixed with a ping to the #Operations Slack channel. The turn around
+  here averages about 12 hours.
+
+* You will need at least one upstream server (for staging) and we encourage at
```
carolyncole: This really may be my lack of understanding of the process, but what is the step you want done here?  Could all the operations pings on the slack channel be done in one step?  An example of the information included int eh slack ping here might also help.

roles/pulibrary.nginxplus/ADDHOSTS.md
kayiwa: we have two identical servers with a virtual IP that they share. if one of them goes down [Keepalived ](https://www.keepalived.org/)will make the hot-standby the current active server.

roles/pulibrary.nginxplus/ADDHOSTS.md
kayiwa: yeah, this makes sense. It should be one step.

roles/pulibrary.nginxplus/ADDHOSTS.md
kayiwa: this is a desired goal but it is not under our control.

roles/pulibrary.nginxplus/ADDHOSTS.md
carolyncole: Change `hot-standy` to `hot-standby` (notice the missing b)

kayiwa: oh boy!

PR #520, opened 2019-05-22
add post and get restrictions

kayiwa:
adds POST and GET restrictions to upstream servers.


tampakis:
There's a typo near the end @hc_lib-solr-stafing_upstream should be @hc_lib-solr-staging_upstream

PR #521, opened 2019-05-24
Adding Princeton and Slavery (PAS) role

carolyncole:
This role just sets up the machine to make it ready to run the PAS codebase.   There will be another role that installs the code and the database.

Adding the ability to skip tasks on ci
Adding playbook for pas-staging
Adding update to apache2 for idempotency
Adding listen port for apache2
Adding update to smb for docker


roles/pulibrary.pas/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+  notify: restart apache
+  with_items:
+    - remoteip
+    - proxy
+    - proxy_fcgi
+
+- name: Install "sass" node.js package globally.
+  npm:
+    name: sass
+    global: yes
```
kayiwa: python prefers true

roles/pulibrary.pas/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+- name: Create mount directories
+  file:
+    path: '/mnt/diglibdata1'
+    state: 'directory'
+
+- name: Copy smb credentials
+  copy:
+    src: files/pas.smb.credentials
+    dest: /etc/pas.smb.credentials
+  when: on_ci != 'true'
```
kayiwa: `molecule test` fails with 
```
fatal: [instance]: FAILED! => {"msg": "A vault password or secret must be specified to decrypt /Users/kayiwa/Documents/dev/pulibrary/princeton_ansible/roles/pulibrary.pas/files/pas.smb.credentials"}
```

carolyncole: @kayiwa and I talked offline.  I will move the not ON_CI task to not in container and remove the test

roles/pulibrary.pas/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+- name: Create mount directories
+  file:
+    path: '/mnt/diglibdata1'
+    state: 'directory'
+
+- name: Copy smb credentials
+  copy:
+    src: files/pas.smb.credentials
+    dest: /etc/pas.smb.credentials
+  when: run_not_in_container
```
kayiwa: this will need to be defined in `molecule/default/playbook.yml` as 
```
var:
  - run_not_in_container: false
```
I did and it passed.

PR #523, opened 2019-05-29
adds role to set eastern timezone

tampakis:
Using the local timezone will make it easier to anticipate daylight savings for cron jobs.


roles/pulibrary.timezone-eastern/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Sets eastern timezone for cron jobs
+  command: timedatectl set-timezone America/New_York
```
escowles: Does this set the timezone for cron or the whole system?

tampakis: Yes it does.

PR #539, opened 2019-06-12
Fix pulmap and pulmap-staging playbooks

eliotjordan:
- Point pulmap_staging to lib-solr-staging
- Move passenger variables to group_vars files


eliotjordan:
Did a test provisioning of maps-staging1. Works as expected.

PR #544, opened 2019-06-19
add stream directory

kayiwa:


kayiwa:
this is a backport that I had to rush in to unbreak maria-staging :-(

PR #546, opened 2019-06-21
Adding the ability for the deploy-user to install a git ssh key

carolyncole:
We are going to have multiple projects needing the private git repo stuff that was in pas.  Moving this to the deploy-user role since that seemed like a good logical place.


carolyncole:
Yes that is correct.  Git deploy keys are for each repo.  They can not be shared between git repos, but can be used on as many servers as desired.  Do PAS will have one and pul_library_drupal would have one.

PR #557, opened 2019-07-13
add iiif-staging config and ssl certs

kayiwa:
config file for iiif-staging


tpendragon:
`server_name` for 443 should be `server_name iiif-staging.princeton.edu;`

PR #559, opened 2019-07-15
Shared data

escowles:
* Add a `shared_data` mount for apps to use to share files between servers
* Add a shared uploads directory for DPUL


escowles:
@tpendragon I pushed a couple small fixes — can you re-review?

PR #609, opened 2019-07-27
Rails app bind9

kayiwa:
while trying to moleculize the capistrano role I noticed that the rails app had identical tasks to the bind role.


roles/pulibrary.bind9/molecule/default/tests/test_bind9.py
```diff
@@ -10,3 +10,4 @@ def test_is_bind9_installed(host):
     pkg = host.package('bind9')
 
     assert pkg.is_installed
+    assert host.process.get(user="bind", comm="named")
```
jrgriffiniii: 👍

Issue #592, opened 2019-07-25
add molecule to Capistrano role


carolyncole:
Also add circleci job

PR #610, opened 2019-07-27
592 capistrano role

kayiwa:
but for the nodejs unpredictable fail this fixed [issue 592](https://github.com/pulibrary/princeton_ansible/issues/592)


roles/pulibrary.capistrano/molecule/default/tests/test_capistrano.py
```diff
@@ -0,0 +1,15 @@
...
+    os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
+
+
+def test_for_capistrano_is_directory(host):
+    f = host.file('/opt/rails_app/shared/tmp')
+
+    assert f.exists
+    assert f.is_directory
+    assert f.user == 'deploy'
+    assert f.group == 'deploy'
```
jrgriffiniii: 👍

Issue #603, opened 2019-07-25
Create a mirror for Node.js and Yarn apt. package repositories


jrgriffiniii:
Otherwise, downloads can fail in CircleCI containers.

PR #611, opened 2019-07-28
Keyservers redundancy

kayiwa:
we don't need (right now anyway) a local cache of node. I've added the MIT servers for gpg keys. This closes [603](https://github.com/pulibrary/princeton_ansible/issues/603)


roles/pulibrary.nodejs/tasks/main.yml
```diff
@@ -4,6 +4,13 @@
     id: "{{ nodejs__upstream_key_id }}"
     keyserver: "hkp://pool.sks-keyservers.net"
     state: present
+  ignore_errors: true
+
+- name: Add upstream Yarn APT repository GPG key from MIT
+  apt_key:
+    id: "{{ nodejs__yarn_upstream_key_id }}"
+    state: present
+    keyserver: "hkp://pgp.mit.edu"
```
jrgriffiniii: 💯

jrgriffiniii:
Thank you very much for this @kayiwa

PR #632, opened 2019-07-31
Sneakers workers

kayiwa:
mirrors sidekiq


roles/pulibrary.sneakers_worker/molecule/default/molecule.yml-e
```diff
@@ -0,0 +1,32 @@
+---
```
jrgriffiniii: I think that this might have been committed in error using a `sed -e` invocation in a macOS environment.

roles/pulibrary.sneakers_worker/tasks/main.yml
```diff
@@ -1,11 +1,23 @@
...
+    src: sneakers.service
+    dest: "/etc/systemd/system/{{ sneakers_worker_name }}.service"
   notify: 'restart sneakers worker'
+
 - name: Keep workers running
-  service: "name={{sneakers_worker_name}} enabled=yes state=started"
+  service:
+    name: "{{ sneakers_worker_name }}"
+    enabled: true
+    state: started
```
jrgriffiniii: Can the service state in the container also be tested given this?  If not, I don't think that this should stop anyone from merging this PR.

kayiwa: our container doesn't have systemd so it cannot.

roles/pulibrary.sneakers_worker/molecule/default/molecule.yml-e
kayiwa: indeed it is. my muscle memory doesn't recall using `sed` howeever. It looks like it was generated by `invoke molecularize pulibrary.sneakers_workers`

PR #642, opened 2019-08-01
Bibdata staging nginx conf

kayiwa:
this moves bibdata-staging to NGINX Plus


tampakis:
Two changes:
I don't think lib-bibdata-staging1 is a real server.
```
    server bibdata-staging1.princeton.edu resolve;
```
The server should be open to the public because CI runs against it

PR #646, opened 2019-08-02
updates solrcloud config for solr 7.7.2

tampakis:


roles/pulibrary.solrcloud/tasks/main.yml
```diff
@@ -17,7 +17,7 @@
   tags:
     - mnt
 
-## Mount the share
+## Mount the share - comment this out for solrcloud_staging
```
escowles: I don't understand this — is there a mechanism to disable this for staging that I'm missing?

eliotjordan:
@tampakis Any thoughts on the review comments above? I assume this is not currently deployed, or am I wrong?

tampakis:
@eliotjordan the solrcloud role doesn't work for staging currently because it doesn't have the backup mount set up. That's why I added the comment.

I deployed this when I upgraded the Solr staging cluster

PR #648, opened 2019-08-03
585 nginx plus role

kayiwa:


roles/pulibrary.nginxplus/README.md
```diff
@@ -287,7 +412,7 @@ nginx_status_port: 8080
 # Requires NGINX Plus.
 # Default is false.
 nginx_rest_api_enable: false
```
jrgriffiniii: Is this going to remain the case?  I know that there were discussions focusing upon this as something which might be explored for updating load balancing configurations.

kayiwa: this config in `group_vars/nginxplus/main.yml` overrides this. 

https://github.com/pulibrary/princeton_ansible/blob/97e7c348cf5a08f90b831740ddc6cec8dc8f65e7/group_vars/nginxplus/main.yml#L4

PR #662, opened 2019-08-05
Adding the YAML lint settings for pulibrary.geoserver

jrgriffiniii:
Also addresses the lint violations for pulibrary.geoserver and ensures that the test suites for pulibrary.tomcat8 and pulibrary.geoserver are run on CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,3 +58,9 @@
        - test_role:
            name: "test_libwww"
            role: "pulibrary.libwww"
+       - test_role:
+           name: "test_tomcat8"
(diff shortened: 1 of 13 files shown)


kayiwa:
changed https://github.com/pulibrary/princeton_ansible/blob/1ac303117f72dd256fd10731692a1d2fb04a254f/roles/pulibrary.tomcat8/handlers/main.yml#L3 to use `service` instead of `systemd`. I had a subsequent commit that addresses this on the `651_geoserver_role` branch. It otherwise fails on circle because of systemd

PR #668, opened 2019-08-06
Adding in production

carolyncole:
Allowing deploy to change the ownership
Allowing deploy to restart apache2
Allowing deploy to run commands as www-data
Updating the database to run using mysql commands instead of drush
Removing update.sql since we are doing that in capistrano


roles/pulibrary.libwww/molecule/default/tests/test_libwww.py
```diff
@@ -26,3 +26,16 @@ def test_for_libwww_apache_config(host, line):
     file = host.file("/etc/apache2/sites-available/000-default.conf")
 
     assert file.contains(line)
+
+
+@pytest.mark.parametrize("line", [
+    "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service apache2 restart",
```
kayiwa: 👍

PR #688, opened 2019-08-09
add tigris private key (do not merge and delete when done)

kayiwa:


kayiwa:
file is available on last pass.

PR #691, opened 2019-08-13
Move Cantaloupe cache to big disks.

tpendragon:
Fixes the problem with cache filling up.


roles/pulibrary.cantaloupe/tasks/install.yml
```diff
@@ -29,7 +29,6 @@
     path: "{{ cantaloupe_FilesystemSource_BasicLookupStrategy_path_prefix }}"
     owner: "{{ cantaloupe_user }}"
     group: "{{ cantaloupe_group }}"
-    recurse: yes
```
carolyncole: @tpendragon not sure why you are no longer setting the owner and group recursively when moving the location.

tpendragon: I didn't move the location of this. This directory is -way- too big to recursively check the owner/group on an existing install, so the provision would never finish if recurse was set to yes.

Issue #692, opened 2019-08-13
Create a new machine for the existing recap website


carolyncole:
Move the existing site https://recap.princeton.edu/ off of libphp_prod to a virtual machine behind the load balancer.

Since this is a small site we will only want one server for staging and one server for production.

The thought would be that since there are other servers associated with recap we may want to name the servers recap_www_prod1 instead of recap_prod1.

PR #696, opened 2019-08-14
Adding new hosts for recap

carolyncole:
Added in the http conf files
refs #692

Need the machines and the SSL certs to finish


carolyncole:
@kayiwa I did not use recap.princeton.edu since I assumed we would get everything working and then move that, vs doing it from the start.  I think we want to keep the current recap up until we know the new one works.  I was envisioning adding it later like we did for the main site.  If we should put it in from the start that is ok too.

kayiwa:
I see, so the library-prod model is what you are going for. I will get new certs for that.

PR #697, opened 2019-08-15
Add force settings as a possible option

carolyncole:
Adding the comment out force setting so folks know that it is possible


kevinreiss:
👍

PR #708, opened 2019-08-20
add solr prod config file

kayiwa:


kayiwa:
ping @tampakis wanna do this?

PR #713, opened 2019-08-22
Adding a role for recap

carolyncole:
ran molecule init role -r roles/pulibrary.recap-www
updated to run using the example files
Moved shared items to drupal role
created a capistrano.yml in the drupal role for items needed for capistrano
updated the recap role to have the correct settings.php

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,9 @@
            name: "test_libwww"
            role: "pulibrary.libwww"
        - test_role:
+           name: "test_redis"
+           role: "pulibrary.redis"
+       - test_role:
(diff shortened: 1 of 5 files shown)


playbooks/libwww-prod.yml
```diff
@@ -6,6 +6,7 @@
     - run_not_in_container: true
     - drupal_version_branch: develop
```
kayiwa: running on development branch?

carolyncole: @kayiwa it is not really used now that capistrano is deploying the code, but it is a fair point.  I changed the commit to remove the reference.

PR #740, opened 2019-08-28
737 nginx plus archivesspace prod

kayiwa:


kayiwa:
the URLS (require VPN) for the server are https://aspace-staff.princeton.edu/ (for staff) and https://aspace.princeton.edu/ (for public) 🤷‍♂

Issue #774, opened 2019-09-04
Move recap-prod.princeton.edu to recap.princeton.edu


carolyncole:
We want to put the new server online in-place of the existing old server.

We need to update the NGINX config file to server recap.princeton.edu
Additionally we need to change the network configuration to point at the  nginx load balancer

The goal is to put this online on Monday September 9th, 2019

kayiwa:
will close this after playbook runs.

PR #775, opened 2019-09-04
Adding recap.princeton.edu to the server configuration

carolyncole:
Removed IP restriction

refs #774


carolyncole:
The only change that would have consequences before Monday is the deletion of the IP restrictions.  I think it could be merged, but lets wait until Friday.

carolyncole:
@kayiwa updated the config to redirect recap-prod to recap for http and https.

PR #778, opened 2019-09-05
777 dpul nginxplus conf

kayiwa:


jrgriffiniii:
As this is urgent and should not affect the Molecule test suites, I'm going to proceed with merging given the delays on https://circleci.com/workflow-run/dfbee28e-1948-42d2-be23-8daa25cb3a02

PR #782, opened 2019-09-05
updates load balancer ips for passenger role

tampakis:


jrgriffiniii:
As this is urgent and should not affect the Molecule test suites, I'm going to proceed with merging given the delays on CircleCI.

PR #791, opened 2019-09-08
790 update datadog role

kayiwa:


carolyncole:
@kayiwa unfortunately this will need to be rebased as I merge another PR.  Then it can be merged.

PR #797, opened 2019-09-09
591 molecule cantaloupe

kayiwa:
add's molecule and circleci to this role


roles/pulibrary.cantaloupe/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 cantaloupe_temp_pathname:
-cantaloupe_src_prefix: cantaloupe
+cantaloupe_src_prefix: Cantaloupe
```
jrgriffiniii: Why was this changed when the version of cantaloupe remained the same?  I would have assumed that the name of the source would have only changed with a 3.3.2 or later release.

jrgriffiniii: Please disregard this question.  https://github.com/cantaloupe-project/cantaloupe/releases/download/v4.1.3/Cantaloupe-4.1.3.zip downloads without issuing a 404 response (https://github.com/cantaloupe-project/cantaloupe/releases/download/v4.1.3/cantaloupe-4.1.3.zip is, however, listed as the proper URL on https://github.com/cantaloupe-project/cantaloupe/releases/tag/v4.1.3)

PR #806, opened 2019-09-11
751 molecule freetds 

kayiwa:
this role seems to be targeting xenial only. Is it not needed for bionic?


kayiwa:
ping @hackmastera

PR #830, opened 2019-09-13
588 bibdata role

kayiwa:
rebased this with master and LGTM.


roles/pulibrary.bibdata/molecule/default/playbook.yml
```diff
@@ -1,12 +1,21 @@
 ---
 - name: Converge
   hosts: all
+  pre_tasks:
```
jrgriffiniii: Is this needed for the Percona cluster?

kayiwa: yes the container doesn't get an usable IP address otherwise. :-(

PR #833, opened 2019-09-13
832 figgy pub sub worker

kayiwa:


kayiwa:
add tests @jrgriffiniii

PR #839, opened 2019-09-17
835 svn role

kayiwa:


roles/pulibrary.svn/tasks/main.yml
```diff
@@ -1,2 +1,6 @@
 ---
 # tasks file for roles/pulibrary.svn
+- name: add mod svn
+  apt:
+    name: ["libapache2-mod-svn"]
```
escowles: Aren't we using `svnserve` instead of `mod_dav_svn`?

kayiwa: indeed. was "cargo culting" this documentation.

escowles: 👍

kayiwa:
heya @escowles my hope (sorry I didn't explain that) is that this role would install svn and a separate one would administer it. This is a pattern @carolyncole and I feels allows us to scale. So one can always install subversion

escowles:
@kayiwa Oh, that makes total sense!

PR #842, opened 2019-09-18
835 lib svn role

kayiwa:
this role will manage a subversion server


kayiwa:
I was able to create a user who is vaulted protected on `group_vars/lib-svn/vault.yml` This is ready to go.

PR #847, opened 2019-09-19
Restore php latest

eliotjordan:
- Restores installation of latest PHP while skipping linting errors.
- Adds new drush installation task (adapted from https://github.com/geerlingguy/ansible-role-drush). Composer no longer allows the installation of drush as root. This can be overcome by setting COMPOSER_ALLOW_SUPERUSER=1, however this is definitely not recommended and breaks Ansible idempotentcy.


eliotjordan:
Opening a new PR without the Drush update, because it's working now. 🤷‍♂

PR #849, opened 2019-09-19
changed vault file and added to gitattributes

pmgreen:
Please merge


pmgreen:
> @pmgreen Could you elaborate a little more on what is happening in this PR? What is being updated, etc... Looks like Carolyn fixed the trailing space issue. 👍

We're pairing on this.

carolyncole:
@eliotjordan we were just fixing up the vault file for the lib-jobs machines.  We realized when we looked at the server the file was missing the vault_ in all the variable names.

We tested it out on the sever and the vault file is looking good now.

PR #866, opened 2019-09-20
Allow editing SVN properties

escowles:


kayiwa:
ping @escowles when this goes green wanna merge?

PR #870, opened 2019-09-24
577 patroni postgres role

kayiwa:
This installs a patroni HA cluster


jrgriffiniii:
This is passing for me locally, https://github.com/pulibrary/princeton_ansible/pull/870/checks?check_run_id=234499899 was canceled, so perhaps that is why the build status indicates that it is failing?

roles/pulibrary.patroni/defaults/main.yml
```diff
@@ -0,0 +1,189 @@
...
+
+patroni_dcs: consul
+patroni_dcs_exists: false
+
+patroni_exec_start_pre: "/bin/mkdir -m 2750 -p /var/run/postgresql/{{ patroni_postgresql_version }}-main.pg_stat_tmp"
+
+patroni_replication_username: replicator
+patroni_replication_password: repuserpasswd
+patroni_superuser_username: postgres
+patroni_superuser_password: supersecretpostgrespasswd
```
jrgriffiniii: Should these values be imported from Vault values or are they overridden later?

jrgriffiniii: kayiwa directed me to https://github.com/pulibrary/princeton_ansible/blob/577_patroni_postgres_role/group_vars/patroni/staging.yml, where it is indeed set.

kayiwa: overridden by `group_vars/patroni/` variables (I made an issue) however. https://github.com/pulibrary/princeton_ansible/issues/872

PR #876, opened 2019-09-26
872 patroni playbooks

kayiwa:


roles/pulibrary.consul/tasks/consul_services_registration.yml
```diff
@@ -11,9 +11,9 @@
     token: "{{ consul_acl_agent_token }}"
   notify: restart consul
   changed_when: false
-  no_log: true
   with_items:
     - "{{ consul_services_register }}"
+  ignore_errors: true
```
jrgriffiniii: Will these raise problems if the Consul HTTP API is not responsive?  I apologize if this concern is unwarranted, I am issuing this based upon my understanding of https://learn.hashicorp.com/consul/getting-started/services.html

PR #896, opened 2019-10-03
880 library www staging

kayiwa:
unbreaks the npm install.


roles/pulibrary.libwww/molecule/default/playbook.yml
```diff
@@ -12,5 +23,5 @@
     - ../../../../site_vars.yml
     - ../../defaults/mariadb.yml
```
carolyncole: This references a file that no longer exists.  There are also references to this file in the production playbook that I do not see being removed.

PR #901, opened 2019-10-08
Needed to allow loading of large database dump

carolyncole:
The main library site could not load the database without this additonal persmission


kevinreiss:
👍

PR #904, opened 2019-10-08
Necessary changes to get Figgy Playbook to run.

tpendragon:
@hackmastera Can you look at this? You added the `become` bits, but the role wouldn't run with them


hackmastera:
The role was broken on vagrant without them. Should the `become_user` be parameterized? But it seems like we don't have a valid vagrant environment right now anyway. The figgy role is in a bit of QA limbo right now.

kayiwa:
I'll work on molecularizing the figgy role but as bibdata has shown me it will need someone who understands it to review/merge @hackmastera Merging this though.

PR #909, opened 2019-10-09
892 patroni client role

kayiwa:
lack of systemd on our docker container makes this role impossible to test.

### DIFF
--- a/roles/pulibrary.patroni/tasks/manage_users.yml
+++ b/roles/pulibrary.patroni/tasks/manage_users.yml
@@ -1,0 +1,48 @@
+---
+- name: configure postgres
+  lineinfile:
+    path: "{{ postgres_conf }}"
+    regex: "^{{ item.name }} ="
+    line: "{{ item.name }} = {{ item.value }}"
+    create: true
+  with_items:
+    - "{{ postgres_settings }}"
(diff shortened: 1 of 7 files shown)


jrgriffiniii:
I will attempt to test this on a Vagrant Box.

roles/pulibrary.patroni/tasks/manage_users.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: create /update pg_hba.conf
```
jrgriffiniii: This is `default.yml` in my Vagrant Box, but is successfully created with the desired values.

jrgriffiniii:
I've also noticed that, for Vagrant, https://github.com/pulibrary/princeton_ansible/blob/master/playbooks/patroni-staging.yml#L6 looked for a file `staging.yml` here: https://github.com/pulibrary/princeton_ansible/tree/master/group_vars/patroni.  If this isn't needed, I can merge this.

jrgriffiniii:
I've also added my work for the Vagrantfile here: https://github.com/pulibrary/princeton_ansible/compare/892_patroni_client_role...892_patroni_client_role-vagrant

PR #910, opened 2019-10-09
880 pas staging percona

kayiwa:
dunno how tight the coupling between `pas` and `pas_code` are so I will make these separate PRs.

### DIFF
--- a/group_vars/pas/pas-staging.yml
+++ b/group_vars/pas/pas-staging.yml
@@ -10,8 +10,8 @@ apache:
 
 pas_password: '{{ vault_pas_db_password }}'
 pas_security_key: '{{ vault_pas_security_key }}'
-db_host: '{{ maria_db_cluster_host | default("localhost") }}'
-db_password: '{{ mariadb_mysql_root_password | default("") }}'
+db_host: 'maria-staging.princeton.edu'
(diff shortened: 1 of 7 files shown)


carolyncole:
@kayiwa I think pas_code will break if the pas role sets up a different database.  That said, since I generally run both together it could work. I think it would be best to have them both ready before merging one.

kayiwa:
heya @carolyncole what is the source of pas.zip? I'm hitting a block with said file.
```
ERROR:
(princeton_ansible) pulibrary.pas_code 880_pas_staging_percona % docker exec -it instance bash
root@instance:/# cd /tmp/
root@instance:/tmp# ls
bionic_all.deb  hsperfdata_root  pas.zip  reset0.sql  reset.sql
root@instance:/tmp# unzip pas.zip
Archive:  pas.zip
error [pas.zip]:  missing 205 bytes in zipfile
  (attempting to process anyway)
error [pas.zip]:  attempt to seek before beginning of zipfile
  (please check that you have transferred or created the zipfile in the
  appropriate BINARY mode and that you have compiled UnZip properly)
  (attempting to re-compensate)
  inflating: pas-craft3.sql
  error:  invalid compressed data to inflate
```
which seems corrupt.

PR #920, opened 2019-10-11
we bump up the health checks from Nginx Plus default

kayiwa:
default is 5s and we bump that to 10s
fails parameter requires the server to fail three health checks to be marked as unhealthy (up from the default one).
passes parameter means the server must pass two consecutive checks to be marked as healthy again instead of the default one.


kayiwa:
I renamed it `figgy-prod.conf` for consistency.

PR #929, opened 2019-10-16
Plantains config

kayiwa:
updates pulfa3 to plantain-staging


roles/pulibrary.nginxplus/files/conf/http/plantain-staging.conf
```diff
@@ -0,0 +1,68 @@
+$ANSIBLE_VAULT;1.1;AES256
```
jrgriffiniii: Should `proxy_pass http://pulfa3-staging;` be to `proxy_pass http://plantain-staging;` (given that `upstream` is `plantain-staging`)?

kayiwa: nice catch.

Issue #923, opened 2019-10-11
Create a Role for pulibrary.pulfa


jrgriffiniii:
This should be a new Role which can be used to provision the following PULFA web app. environments:
- production
- staging
- local

Where `local` should be used for Vagrant-based provisioning on a MacBook or personal development environment.

jrgriffiniii:
In order to address `sed`-like operations one is faced with two options:
- https://docs.ansible.com/ansible/latest/modules/replace_module.html#replace-module
- https://docs.ansible.com/ansible/latest/modules/shell_module.html?highlight=shell

There may be some limitations to using `replace`, but it seems to provide enough support for RegExp that this could suffice for porting the `sed` operations found in the Vagrantfile.

jrgriffiniii:
The Ansible documentation on variables can be found here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html

Specifically, the variable precedence follows this scheme:

- command line values (eg “-u user”)
- `role` defaults
- inventory file or script group vars
- inventory `group_vars/all`
- playbook `group_vars/all`
- inventory `group_vars/*`
- playbook `group_vars/*`
- inventory file or script host vars
- inventory `host_vars/*`
- playbook `host_vars/*`
- host facts / cached set_facts
- play `vars`
- play `vars_prompt`
- play `vars_files`
- role vars (defined in `role/vars/main.yml`)
- block vars (only for tasks in block)
- task vars (only for the task)
- include_vars
- set_facts / registered vars
- role (and include_role) params
- include params
- extra vars (always win precedence)

(please see the section https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable)

jrgriffiniii:
@cwulfman I have scaffolded a new role, playbook, and Vagrantfile with https://github.com/pulibrary/princeton_ansible/tree/prototype-pulfa-jrgriffiniii.  I am testing this now, and will offer some more instructions for running the Vagrantfile before proceeding with integrating SVN for checking out the EADs.

jrgriffiniii:
https://github.com/pulibrary/princeton_ansible/commit/43f4ffeff19278229659e7f3eaddf4a407a68724 advances this with Solr and eXistdb on the Vagrant Box.

jrgriffiniii:
This should just duplicate Subversion checkouts between this new Role and that of `plantain`.

acozine:
[...]

PR #933, opened 2019-10-16
Implements a pulibrary.subversion-pulfa Role for plantain_staging

### LINKED ISSUE #931: plantain_staging should use SVN to checkout PULFA EADs into the server environment

jrgriffiniii:
Resolves #931.  Please note that the tests are not run using GitHub Actions given that the role itself requires that Ansible vault credentials be decrypted.


roles/pulibrary.subversion-pulfa/meta/main.yml
```diff
@@ -0,0 +1,16 @@
...
+
+  min_ansible_version: 2.2
+
+  platforms:
+    - name: Ubuntu
+      versions:
+        - 16.04
+        - 18.04
+dependencies:
+  - role: 'pulibrary.subversion'
```
jrgriffiniii: This needs to be corrected.

roles/pulibrary.subversion-pulfa/molecule/default/playbook.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Converge
+  hosts: all
+  roles:
+    - role: pulibrary.subversion
```
jrgriffiniii: This needs to be corrected.

jrgriffiniii:
This needs to be rebased, but might block progress on `plantain-staging`

Vagrant/plantainStagingVagrantfile
```diff
@@ -0,0 +1,61 @@
+# -*- mode: ruby -*-
```
jrgriffiniii: Solr should be installed if this is to remain a viable testing environment.

playbooks/plantain_staging.yml
```diff
@@ -10,6 +10,7 @@
     - role: roles/pulibrary.ruby
     - role: roles/pulibrary.postgresql
     - role: roles/pulibrary.openjdk
+    - role: roles/pulibrary.subversion-pulfa
```
kayiwa: is this role a dependency of plantain? if yes why not add this in the `meta/main.yml` of the `pulibrary.plantain` role to prevent it running two times?

hackmastera:
Since this is a single task @kayiwa and I decided to fold it into the plantain role. This test is tricky because to connect to the live repo we'd need to give the tests access to the vaulted credentials (not sure how to do this) and because the checkout takes a long time due to the size of the repo. It might be possible to mock the svn response with a fixture but this isn't something any of us have tried with molecule. Inclination is to scrap this test for now.

I'll open a new PR that adds the task to plantain.

jrgriffiniii:
This was structured as an independent role given that I was assisting @cwulfman with the development of a Vagrant Box (and Playbook) for deploying PULFA 2.0.

jrgriffiniii:
https://github.com/pulibrary/princeton_ansible/issues/923 was the issue that this was also seeking to advance, but I completely understand the concerns regarding the testing of the Role.  It's a simple task really - should I just duplicate this between the `pulfa` and `plantain` Playbooks?

kayiwa:
> #923 was the issue that this was also seeking to advance, but I completely understand the concerns regarding the testing of the Role. It's a simple task really - should I just duplicate this between the `pulfa` and `plantain` Playbooks @kayiwa and @hackmastera?

I would personally put them as a task in `pulfa` and `plantain` instead of a playbook.

PR #946, opened 2019-10-18
Adding support for Node.js LTS installations for pulibrary.nodejs

jrgriffiniii:
This uses https://github.com/nodesource/distributions/blob/master/README.md#deb and includes the https://deb.nodesource.com/setup_10.x retrieved by `curl` in:

```bash
# Using Ubuntu
curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
sudo apt-get install -y nodejs
```

Without this, staging and production servers cannot have Node.js updated beyond older releases.

### DIFF
--- a/roles/pulibrary.nodejs/files/setup_10.x
+++ b/roles/pulibrary.nodejs/files/setup_10.x
@@ -3,20 +3,20 @@
 # Discussion, issues and change requests at:
 #   https://github.com/nodesource/distributions
 #
-# Script to install the NodeSource Node.js 12.x repo onto a
+# Script to install the NodeSource Node.js 10.x repo onto a
 # Debian or Ubuntu system.
 #
 # Run as root or insert `sudo -E` before `bash`:
 #
-# curl -sL https://deb.nodesource.com/setup_12.x | bash -
+# curl -sL https://deb.nodesource.com/setup_10.x | bash -
 #   or
-# wget -qO- https://deb.nodesource.com/setup_12.x | bash -
+# wget -qO- https://deb.nodesource.com/setup_10.x | bash -
 #
 
 export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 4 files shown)


jrgriffiniii:
This needs to be updated to support the latest LTS release (12.13.0, as of today).

jrgriffiniii:
This has been successfully tested using the `plantain` staging server environment.

carolyncole:
@jrgriffiniii I have been doing this a slightly different way: https://github.com/pulibrary/princeton_ansible/blob/master/roles/pulibrary.approvals/tasks/main.yml#L25-L39
I assume the script you are checking in is the same one I am downloading?

jrgriffiniii:
Yes, I was checking in the script in the off-chance that a malicious script was instead offered for download (more out of principle than any realistic fear).

jrgriffiniii:
The Drupal Molecule test suite is failing with the following output:
```
fatal: [instance]: FAILED! => {"cache_update_time": 1572374136, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o "Dpkg::Options::=--force-confdef" -o "Dpkg::Options::=--force-confold"     install 'nodejs-dev' 'node-gyp' 'npm' 'libssl1.0-dev'' failed: E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages."
```

Given that `create` is run with a new container environment before this suite is executed, I suspect that this is not related.

carolyncole:
@jrgriffiniii sine it is installing node packages for anisble actions and you have just changed the default node version I would guess it is related.  It is possible that step is no longer needed.

jrgriffiniii:
This is deprecated by https://github.com/pulibrary/princeton_ansible/pull/1303 and other Role-specific updates.

PR #952, opened 2019-10-24
Renaming run_not_in_containter

### LINKED ISSUE #643: Rename run_not_in_container
We are finding it hard to understand the wording of run_not_in_container, when out of the immediate context.

We believe that it should be renamed to `run_on_server_only`.

carolyncole:
Renames `run_not_in_containter` to `running_on_server for clarity` on when it should be true and when it should be false.
True if you are running on a server, false if you are running in a container (any molecule playbook).
fixes #643

@hackmastera & @jrgriffiniii I know we discussed `run_on_server_only` but when I read the tasks and vars I though `running_on_server` sounded better.  
For example: `when: run_on_server_only` vs `when: running_on_server`

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ In all the steps below substitue your role name for `example`
       1. Add:
          ```
          vars:
-           - run_not_in_container: false
+           - runnning_on_server: false
          ```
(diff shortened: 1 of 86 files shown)


kayiwa:
woot!

PR #960, opened 2019-10-29
Plantain Solr Configuration

tpendragon:


roles/pulibrary.solr/tasks/main.yml
```diff
@@ -26,12 +26,13 @@
   when: solr_install_path_status.stat.isdir is not defined
   register: solr_download_status
 
-- name: Expand Solr.  # noqa 503
+- name: Expand Solr.  # noqa 503 ANSIBLE0016
```
hackmastera: what is this?

playbooks/plantain_staging.yml
```diff
@@ -14,7 +14,7 @@
     - role: roles/pulibrary.nodejs
     - role: roles/pulibrary.openjdk
     - role: roles/pulibrary.blacklight-app
```
jrgriffiniii: Might this be added as a dependency to the `pulibrary.plantain` role? I would contend that its own dependencies should be run for each `plantain` server provisioning.

group_vars/plantain/plantain_staging.yml
```diff
@@ -45,6 +50,6 @@ rails_app_vars:
   - name: APPLICATION_HOST_PROTOCOL
     value: '{{application_host_protocol}}'
   - name: SOLR_URL
-    value: 'http://lib-solr-staging.princeton.edu:8983/solr/pulfa3-staging'
+    value: "http://localhost:8983/solr/{{ plantain_solr_core }}"
```
jrgriffiniii: These aren't being set for the user `deploy`, I am working to diagnose and fix these now.

roles/pulibrary.solr/tasks/main.yml
kayiwa: I added the `.ansible-lint` which is the pattern used in other roles @hackmastera

PR #967, opened 2019-11-01
upgrade to postgresql-10 postgis-2.4

kayiwa:
unbreaks geoserver.


kayiwa:
ping @eliotjordan OK?

Issue #916, opened 2019-10-10
mudd servers


carolyncole:
We would need two servers lib-mudd-staging & lib-mudd-prod.  These will need ssl certificates.

They will be proxied from https://library-staging.princeton.edu/special-collections/mudd and https://library.princeton.edu/special-collections/mudd.

sdellis:
@carolyncole I would call them lib-mudd-dbs-staging and lib-mudd-dbs-prod since this is simply a front end for a bunch of database tables. I also wonder if the proxy info is correct? I think they should be proxied from `library.princeton.edu/mudd-dbs/`.  Here is an example page for the faculty table where queries get requested using GET to the following address: https://library.princeton.edu/mudd-dbs/faculties?qname=graduate+alumni+index&LNAME=&FNAME=&dept=

kayiwa:
this isn't a big deal but what's the penalty (what's in a name eh?) I already registered the DNS names in the subject (so lib-mudd-staging & lib-mudd-prod) and requested certificates. will the lack of DBS be a problem?

kevinreiss:
the lack of DBS is not a problem.

kayiwa:
ping @sdellis 👆

kevinreiss:
End users will only ever see this proxied through the our main domain so the name is arbitrary.

kayiwa:
VMs are up and running. PR's with certs will go in.

PR #972, opened 2019-11-04
Adding nginx configuration for lib-mudd

carolyncole:
refs #916


carolyncole:
@kayiwa updated to 64k.  Thanks for catching that.

PR #974, opened 2019-11-05
Issue 962 historicperiodicals conf

kayiwa:


kayiwa:
The PR adds veridian based sites. Since they are all running on the same upstream host they will need to be on the same server blocks. We start with historicperiodicals and blue mountain and their SSL certificates.

cwulfman:
@kayiwa , why do you think the bluemountain key(s) are failing?  What do we need to do to resolve that?

kayiwa:
> @kayiwa , why do you think the bluemountain key(s) are failing? What do we need to do to resolve that?

This PR resolves that. It has SSL certs for bluemountain.

PR #976, opened 2019-11-05
Creating separate proxy config files for staging and prod

carolyncole:
Changing the staging rule to point at the new lib-mudd-staging server


roles/pulibrary.nginxplus/files/conf/http/templates/libwww-proxy-pass-prod.conf
```diff
@@ -0,0 +1,6 @@
+    rewrite ^/locator/(.*)$ https://locator-prod.princeton.edu/$1 redirect;
+    
+    rewrite ^/special_collections/mudd-dbs/(.*)$ /mudd-dbs/$1 redirect;
+    location /mudd-dbs {
+        proxy_pass http://libruby-prod.princeton.edu/mudd-dbs/;
```
kayiwa: will this be modified with a future PR?

kayiwa: I expected this to be `https://lib-mudd-prod.princeton.edu/`

carolyncole: @kayiwa yes, once we get the green light that staging is working as expected we will modify the production path to the new prod machine.

PR #991, opened 2019-11-07
Implementing a task for checking out the PUL finding aids using subversion in the pulibrary.plantain Role

jrgriffiniii:
This is necessary to advance https://github.com/pulibrary/plantain/issues/12


jrgriffiniii:
The Subversion task for this should only be run on the server - testing would require that the Molecule container have the proper access to the Vault SVN credentials. I'm setting `running_on_server` and removing the tests in order to remedy this.

PR #997, opened 2019-11-08
Issue 979 arks redirects

kayiwa:


christinach:
but I also see this build error and I'm not sure what to do in this case. Do I merge? 
`Workflows can't be executed on this repository. Please check your payment method or billing status.`

Issue #988, opened 2019-11-07
SSL certificate for static library sites


carolyncole:
We will need SSL certificates for the new static library sites
- republic-staging.princeton.edu
- firestone-renovations-staging.princeton.edu
- milberg-staging.princeton.edu

These will get added to nginx all the sites reside on libstatic-staging1

PR #998, opened 2019-11-08
Adding nginx configuration for static

carolyncole:
Adds firestone-republic, milberg, and republic staging to nginx

refs #988

@kayiwa this is my first attempt at a multi machine.  Please look at this and see if I have done it correctly.  I have just one cache for the entire set, but 3 zones, 3 servers, and 3 ssl servers.  Does that seem correct?


kayiwa:
I would have DRY'ed it myself @carolyncole I would have created one upstream since they are all call the same server. Look at veridian-sites.conf as an example. Otherwise the zones, caching etc., are used when you have more than one upstream server (I feel it is a good habit to be in even if we don't here)

PR #1006, opened 2019-11-12
Moving proxy pass for mudd-dbs to the new server

carolyncole:


carolyncole:
Change this to be a proxy pass instead of a redirect

PR #1009, opened 2019-11-13
add letsencrypt on adc

kayiwa:
this PR allows us to update the cicognara LE certificates on the adc and allow this to run automatically.


tpendragon:
@kayiwa I can merge this, but I don't really understand. Are the letsencrypt keys here already?

kayiwa:
> @kayiwa I can merge this, but I don't really understand. Are the letsencrypt keys here already?

ran those manually this AM. May need to make that a mounted share though.

PR #1013, opened 2019-11-13
Confluence install

kayiwa:


pmgreen:
Two failing checks relate to billing structure.

PR #1040, opened 2019-11-21
add an upstream block and not just a redirect

kayiwa:
this closes issue #1039


cwulfman:
@kayiwa there's a build fail.

PR #1046, opened 2019-11-23
Arks redirect princeton nginx

kayiwa:
pending upstream confirmation on if arks-redirect need to be a hostname or an alias'ed one.


kayiwa:
thanks for merging this @escowles

PR #1072, opened 2019-12-02
1070 cleaner 502

kayiwa:
this adds the error page to all our staging servers. If this works today a subsequent PR will go in.


carolyncole:
@kayiwa I wonder if these changes should be an include like the proxy pass rules.  Seems like we are changing all the files together.

PR #1082, opened 2019-12-09
source our lpass environment variables

kayiwa:


kayiwa:
this PR adds a shell script that will one can `source` to set your environment inside your virtualized shell. I prefer it because I'd rather not have this globally for other ansible projects I do that do not have a need for vault (which produce errors) and or lastpass.

PR #1117, opened 2019-12-12
fix pythonic linting errors

kayiwa:
remove trailing space and add space after a variable


kayiwa:
the locator role had linting problems that this fixes.

PR #1137, opened 2019-12-18
rather than use a template to deploy postgres

kayiwa:
we will use a stock installation until it fails us


carolyncole:
@kayiwa this PR seems very unhappy...

kayiwa:
it is curious that it is looking for patroni when patroni is deleted. Looking into this.

PR #1168, opened 2020-01-06
add two new networks

kayiwa:
we have two new VPN networks. This adds them to the allow list

This closes issue #1167


hackmastera:
Also makes changes to the error handling on plantain staging

PR #1176, opened 2020-01-08
change URI to no solr collection path

kayiwa:


carolyncole:
@kayiwa apologies for being very late to the game here, but do we want to do this now?

PR #1177, opened 2020-01-09
add the lib-svn to the hosts

kayiwa:
we add a playbook and group_vars for lib-svn.
also remove mariadb2 and mariadb3


kayiwa:
this adds lib-svn-prod1 (it has a `lib-svn` alias)

PR #1199, opened 2020-01-16
add apache role as a dependency

kayiwa:
this also adds configuration for dav and dav_svn
the apache role now has an apache reload handler
the www-data user should be able to read and write the subversion
directories


sdellis:
@kayiwa and @christinach did we get approval to not password protect the web interface? That's the only additional work that needs to be done here.

PR #1219, opened 2020-01-21
Configure DPUL1 to be redis leader.

tpendragon:


group_vars/dpul_production.yml
```diff
@@ -62,6 +62,10 @@ rails_app_vars:
     value: '{{ dpul_honeybadger_key }}'
   - name: POMEGRANATE_COLLECTIONS_URL
     value: 'https://figgy.princeton.edu/iiif/collections'
+  - name: POMEGRANATE_REDIS_URL
+    value: 'dpul1.princeton.edu'
+  - name: POMEGRANATE_REDIS_DB
+    value: '2'
```
hackmastera: where do you get this value?

tpendragon: It's just arbitrary. Figgy is set to 1, so if we ever do a shared Redis server we don't have to change this.

PR #1237, opened 2020-01-24
parametrize ruby versions

kayiwa:


roles/pulibrary.ruby/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
 # defaults file for roles/pulibrary.ruby
+ruby_version: "{{ ruby2_6 | default(ruby2_4) }}"
```
hackmastera: I think this will be confusing when we want to upgrade to 2.7. Then we would set `ruby2_6: "ruby2.7"`? Maybe we could use something like `ruby_version: "{{ ruby_version_override | default(ruby_version_default) }}"?

PR #1246, opened 2020-01-27
Moving approvals to ruby 2.6

carolyncole:
I had to remove the entry in all to get the override to process.
Additionally I had to upgrade bundler to get it to point at the new ruby.

I also had to ssh onto the box and restart nginx.


group_vars/all/vars.yml
```diff
@@ -1,5 +1,4 @@
 ---
-ruby_version: "{{ ruby_version_default }}"
```
kayiwa: I wondered about this myself but wasn't brave enough to remove the hard linked 2.3.1

PR #1248, opened 2020-01-27
clamav does multiple downloads

kayiwa:
and makes idempotency fail


roles/pulibrary.clamav/tasks/main.yml
```diff
@@ -14,3 +14,5 @@
   command: /usr/bin/freshclam --quiet
   args:
     creates: /var/lib/clamav/main.cvd
+  changed_when: false
```
jrgriffiniii: I see, this normally provides a false positive.

PR #1251, opened 2020-01-29
add lib-solr8-staging configuration

kayiwa:


kayiwa:
> It looks good. I see more cache settings on lib-solr-staging, but I'm not familiar with those cache decisions. If you think they're not needed then 👍

thanks for bringing it up

PR #1284, opened 2020-02-11
new path to LE ssl file

kayiwa:
it took two tries to check the path of the LE files location
the archiving is caused by two LB running the process

this closes issue #1282 which will be an ADR


architecture-decisions/0003-cicognara-letsencrypt-path.md
```diff
@@ -0,0 +1,24 @@
...
+
+There are instances where we will be unable to use our InCommon Federation SSL certificate framework for certs. In instances where a dot org, etc., In those instances Let's Encrypt is an adequate solution. 
+
+The Lets Encrypt expects the SSL certs to be under `/etc/letsencrypt/` which differs from where we place the rest of our SSL certificated
+
+
+## Decision
+
+make sure certificates are under `/etc/letsencrypt/`
+create a directory under `/var/www/letsencrypt/<servicename>`
```
hackmastera: What happens with this directory? Is it empty?

kayiwa: yes it is written to temporarily when LE is verifying.

hackmastera: 👍 maybe just add that as the end of the sentence?

PR #1291, opened 2020-02-13
Orangelight: add env var for RESERVES_SOLR_URL

hackmastera:
advances pulibrary/orangelight#1933


hackmastera:
No longer needed since we can't move this to a different server due to cross-core join.

PR #154, opened 2018-01-23
Solr Cloud

kayiwa:
this role installs solr cloud and zookeeper on *at least* 3 VMs 

Mostly needs documentation (zookeeper is new to me)

It takes a while to build

PR #1299, opened 2020-02-24
Remove Jetty Configuration causing issues with Solr 8.x

tpendragon:
These jetty config files were added in during https://github.com/pulibrary/princeton_ansible/pull/314/files#diff-bc10ecbc9a1130f0fc92d984791827b5R43. Unfortunately those config files differ between Solr 7 and Solr 8, resulting in a bug making solr boxes being unable to connect to one another during query. We don't seem to change any of the configuration in our local group_vars, but we may have changed the numbers when they first went in.

@jrgriffiniii do you remember when you first made this role if these were variables brought over from somewhere else, or if we changed any of the numbers initially?

### DIFF
--- a/group_vars/solr8cloud/staging.yml
+++ b/group_vars/solr8cloud/staging.yml
@@ -1,6 +1,7 @@
 ---
 solr_heap_setting: '20g'
-solr_cloud_download_version: 8.2.0
+solr_cloud_download_version: 8.4.1
+solr_cloud_url: "https://archive.apache.org/dist/lucene/solr/{{ solr_cloud_version }}/{{ solr_cloud_package}}"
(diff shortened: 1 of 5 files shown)


jrgriffiniii:
I apologize, but I'm pretty certain that this was structured by @kayiwa : https://github.com/pulibrary/princeton_ansible/pull/154#issuecomment-419210069

https://github.com/pulibrary/princeton_ansible/pull/314/files was a rebase which integrated much of that work, but none of it was my own.

jrgriffiniii:
https://github.com/idealista/solrcloud-role/blob/master/defaults/main.yml#L63 looks to me as if this might be the original Role.

tpendragon:
Great! Seems like it's safe to delete then.

PR #1308, opened 2020-02-27
add new solr prod hosts

kayiwa:


kayiwa:
so much for simple eh? I've corrected this.

PR #1349, opened 2020-03-05
Move dss staging to use solr8 cloud

hackmastera:
This application uses the catalog configset, which is moving to solr8.


hackmastera:
@kayiwa this is fixed, thanks!

PR #1372, opened 2020-03-17
Updating the pulibrary.plantain to interface with Solr remotely

jrgriffiniii:
Updates the Solr Core URL for the pulibrary.plantain Role, and ensures that Solr is not installed locally for the pulibrary.plantain Role


jrgriffiniii:
This should be closed, as `plantain` has been renamed to `pulfalight`.

PR #1375, opened 2020-03-18
[Cantaloupe] Revert to KakaduDemoProcessor

tpendragon:
We immediately got complaints and the production server crashed.


kayiwa:
thanks @tpendragon

PR #1399, opened 2020-03-26
Rename plantain to pulfalight

hackmastera:
Notable:
* Rename passenger app root
* Rename rails app (env vars file, logrotate files)
* Rename sidekiq workers
* Rename database
* Rename database user
* Change rails app var names (coordinated with change in codebase)


jrgriffiniii:
The test script appears to be failing with:

```
Run ROLE=pulibrary.plantain python run_molecule.py
sh: 1: cd: can't cd to roles/pulibrary.plantain
```

Perhaps `plantain` is being passed to https://github.com/pulibrary/princeton_ansible/blob/master/run_molecule.py#L59?

hackmastera:
@jrgriffiniii thanks; this is now fixed. `ag` had skipped the hidden files (usually so helpful).

PR #1421, opened 2020-04-04
replace lib-app-status renewed certs

kayiwa:


roles/pulibrary.nginxplus/files/conf/http/researchdata_prod.conf
```diff
@@ -1,39 +1,39 @@
 $ANSIBLE_VAULT;1.1;AES256
```
jrgriffiniii, line 1: I am assuming that the redirection here is temporary, until we have a dedicated app. for managing research data.

PR #1437, opened 2020-04-15
Making sure that postges is restarted after access is changed

carolyncole:
Update the readme for required variables

Co-authored-by: Francis Kayiwa <kayiwa@users.noreply.github.com>

### DIFF
--- a/roles/pulibrary.bibdata/defaults/main.yml
+++ b/roles/pulibrary.bibdata/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 running_on_server: false
+db_host: "{{ postgres_host | default(omit) }}"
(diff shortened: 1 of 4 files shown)


roles/pulibrary.postgresql/tasks/main.yml
```diff
@@ -75,7 +75,18 @@
...
     - not postgresql_is_local
-  notify: 'reload remote postgres'
+    - running_on_server
+
+- name: reload remote postgres now for the access to take place
+  service:
+    name: postgresql
+    state: reloaded
+  become: yes
+  delegate_to: '{{ db_host }}'
```
carolyncole, line 85: @kayiwa this should be `delegate_to: '{{ postgres_host}}' ` like line 73

roles/pulibrary.bibdata/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 running_on_server: false
+db_host: "{{ postgres_host | default(omit) }}"
```
carolyncole, line 3: This would not be needed if we delegate to postgres_host

PR #1455, opened 2020-04-22
initialize new role

kayiwa:


roles/pulibrary.timezone/molecule/default/playbook.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Converge
+  hosts: all
+  vars:
+    timezone: America/New_York
+  roles:
+    - role: pulibrary.timezone
```
christinach, line 7: Can we add a var for the timezone in this PR? Or does this needs to be done in a separate PR?

kayiwa: there's one in [defaults/main.yml](defaults/main.yml). The benefit here is if one adds the variable `timezone: America/New_York` in their `group_vars/yourrole/vars.yml` it will add that.

PR #1495, opened 2020-05-06
Add figgy role dependency

hackmastera:


carolyncole:
@hackmastera it seems odd to have your dependencies split
You have a load of them in the playbook: https://github.com/pulibrary/princeton_ansible/blob/master/playbooks/figgy_staging.yml#L9-L25

carolyncole:
@hackmastera and I had a conversation outside of git.  The reason for this one being it the meta is it is required for the role to run properly.

PR #1501, opened 2020-05-07
this bumps our elastic version

kayiwa:


roles/pulibrary.elastic/tasks/main.yml
```diff
@@ -1,33 +1,48 @@
...
+
+- name: Elastic | disable elastic
+  service:
+    enabled: false
+    masked: true
+    name: elasticsearch
+  when: elastic_conf_dir is not defined
+
+- include_tasks: configure.yml
+- include_tasks: post_configuration.yml
```
eliotjordan, line 45: I don't see a `post_configuration.yml` file in this PR? Is it missing or is this not needed?

kayiwa: thanks for the catch. It is missing. It is not needed to go "green" because all the steps on run on the server.

PR #1509, opened 2020-05-11
1508 percona non cluster

kayiwa:
[percona] this adds a role which install percona in an unclustered setup. Currently points to `maria-staging3.princeton.edu` as the host


hosts
```diff
@@ -119,7 +119,7 @@ lib-approvals-staging1.princeton.edu
 [approvals_prod]
 lib-approvals-prod1.princeton.edu
 [percona_staging_cluster]
-maria-staging1.princeton.edu
+#maria-staging1.princeton.edu
```
carolyncole, line 122: Instead of commenting out the items that are still in the cluster machines we should create a new set of hosts percona_staging & percona_prod and move the host that will not be clustered there.

hackmastera:
Ha! I see what you did there.

PR #1523, opened 2020-05-15
add global protect and modify listening port

kayiwa:
[lib-monitor] currently is unable to render if one is using the new VPN


carolyncole:
@kayiwa why did the server line change to `lib-monitor1.princeton.edu:3000` vs `lib-monitor1.princeton.edu`?  Seems like a strange change to allow the Global Protect VPN IPs.

kayiwa:
we are testing [sensu](https://sensu.io) instead of netdata

Issue #1542, opened 2020-05-28
Add PILOT_USERS list Environment to the catalog-production and catalog-staging playbook


kevinreiss:
The app_configs environment file should set a PILOT_USERS variable that contains a list of net id's eligible for pickup requests.

kevinreiss:
Closing as we decided to do this via Capistrano.

PR #1546, opened 2020-05-29
[Orangelight] Adding PILOT_USERS environment variable

carolyncole:
Also moving production group vars into the orangelight folder

@christinach & @kevinreiss after some thought I realized when ever the playbook runs it would delete the pilot users, so it would be better to also add them here.

I did create a cap command so we can easily change the values without continuously changing and running ansible.  Once they stabilize we should then update here.

https://github.com/pulibrary/orangelight/commit/da11b2385874cec3c39790022fe041314232c0df

refs #1542


carolyncole:
Realized we could put the env into it's own file.  We do not need this...

PR #1589, opened 2020-06-18
Updated SVN username and password

kelynch:
This updates username and password for the "figgysvn" user created in "lib-svn."

Do the SVN URL or directory Figgy variables need to be changed as well?


kelynch:
Updated URL and directory for the load balancer.

kelynch:
SVN URL and directory variables needed to be updated to point to the new SVN server and remove a duplicate directory reference.  This should now be addressed.

kelynch:
Remote URL is incorrect.  Testing for correct one now; I'll update the PR once we have it.

PR #1612, opened 2020-07-02
hide the xframe options

kayiwa:


hackmastera:
We've decided to do this in the application.

PR #1632, opened 2020-07-10
Revert "Replace whitelist with allowlist"

carolyncole:
Reverts pulibrary/princeton_ansible#1628

We were just not quite ready for this yet.  Let's revert it until we are.


hackmastera:
We'll try again once we're ready / able to deploy the application code that goes along with this. Thanks!

PR #1634, opened 2020-07-14
increase ulimit for redis

### LINKED ISSUE #1560: [BIBDATA] Redis can't set maximum open files to 10032 because of OS error: Operation not permitted.

kayiwa:
the [bibdata] roles need a higher ulimit for the redis application

this closes #1560

### DIFF
--- a/roles/pulibrary.redis/templates/redis_ulimit.conf.j2
+++ b/roles/pulibrary.redis/templates/redis_ulimit.conf.j2
@@ -1,0 +1,3 @@
+# {{ ansible_managed | comment }}
+redis soft nofile 65535
+redis hard nofile 65535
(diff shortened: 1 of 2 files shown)


roles/pulibrary.redis/templates/redis_ulimit.conf.j2
```diff
@@ -0,0 +1,3 @@
+# {{ ansible_managed | comment }}
+redis soft nofile 65535
+redis hard nofile 65535
```
christinach, line 3: should the soft nofile be a different number?

kayiwa: If this was your desktop machine you'd want "the warning" that soft limit would buy us. If redis is consuming all of those resources we want to know with a hard crash.

PR #1651, opened 2020-07-23
a new apache tomcat role

kayiwa:
this role install tomcat from upstream without relying on the Ubuntu
package manager

this is needed for [dataspace] and [oar] and is based of my draft PR to make review easier


roles/pulibrary.apache_tomcat/templates/tomcat.service.j2
```diff
@@ -0,0 +1,23 @@
...
+After=syslog.target network.target
+
+[Service]
+Type=forking
+
+Environment=JAVA_HOME={{ tomcat_java_home }}
+Environment=CATALINA_PID={{ tomcat_install_path }}/{{ tomcat_service_name }}/temp/tomcat.pid
+Environment=CATALINA_HOME={{ tomcat_install_path }}/{{ tomcat_service_name }}
+Environment=CATALINA_BASE={{ tomcat_install_path }}/{{ tomcat_service_name }}
+Environment='CATALINA_OPTS=-Xms{{ tomcat_jvm_memory_percentage_xms*ansible_memtotal_mb//100 }}M -Xmx{{ tomcat_jvm_memory_percentage_xmx*ansible_memtotal_mb//100 }}M -server -XX:+UseParallelGC {% if tomcat_debug_mode == True %}-Xdebug -Xrunjdwp:transport=dt_socket,address={{ tomcat_port_debug }},server=y,suspend=n{% endif %}'
```
jrgriffiniii, line 13: `-Xms` and `-Xmx` look to be set to the same expression `{{ tomcat_jvm_memory_percentage_xms*ansible_memtotal_mb//100 }}M` - I don't oppose setting the initial and maximum memory pool for the Tomcat process, but I suspect that `-Xmx` might need to be increased in another process should performance issues be encountered.

kayiwa: given that in this particular case our Tomcat will be optimized using an Atmire config I erred on the side of generic. In all other cases we use Ubuntu's package manager and the final config decisions are not made in Tomcat but by the application (a different role) that is going to need Tomcat.

PR #1650, opened 2020-07-23
1621 apache ant and maven role

kayiwa:
- add a manual apache tomcat roleA
- add apache_maven role

this is a pulled from my draft 1621 PR to make these more manageable for review.

PR #1653, opened 2020-07-23
increase the proxy buffers for pulmap

kayiwa:
this changes will increase the buffer allocation for pulmap requests.

This will close #1650


jrgriffiniii:
The following were updated for the NGINX Plus load balancing for the pulmap servers:

```
proxy_buffer_size
proxy_buffers
proxy_busy_buffers_size
```

PR #1660, opened 2020-08-04
point to the postgresql staging cluster

kayiwa:


christinach:
@kayiwa thank you for fixing that

PR #1662, opened 2020-08-04
Move poppler install to vips role

hackmastera:
refs pulibrary/figgy#4067


roles/pulibrary.vips/tasks/main.yml
```diff
@@ -1,13 +1,21 @@
 ---
+- name: detect vips commands
+  command: vips -l
+  register: vips_list
```
kayiwa, line 4: Nice!

PR #1667, opened 2020-08-06
Creating a production playbook for the research data machines

carolyncole:
Also updated postgres to get around unknown attribute login_port


roles/pulibrary.postgresql/tasks/create_db.yml
```diff
@@ -3,7 +3,7 @@
   postgresql_db:
     name: '{{ application_db_name }}'
     login_host: '{{ postgres_host | default(omit) }}'
-    login_port: '{{ postgres_port | default(omit) }}'
+    port: '{{ postgres_port | default(omit) }}'
```
carolyncole, line 6: @kayiwa this change was really strange, but I got a no attribute error when running the playbook.

roles/pulibrary.postgresql/tasks/create_users.yml
```diff
@@ -3,7 +3,7 @@
   postgresql_user:
     name: '{{ application_dbuser_name }}'
     login_host: '{{ postgres_host | default(omit) }}'
-    login_port: '{{ postgres_port | default(omit) }}'
+    port: '{{ postgres_port | default(omit) }}'
```
carolyncole, line 6: @kayiwa here too.  Not entirely sure why...

roles/pulibrary.postgresql/tasks/create_db.yml
kayiwa: surprised it worked given this. https://docs.ansible.com/ansible/2.3/postgresql_db_module.html (there's no `login_port`) Good catch.

roles/pulibrary.postgresql/tasks/create_users.yml
kayiwa: same here. https://docs.ansible.com/ansible/latest/modules/postgresql_user_module.html

PR #1675, opened 2020-08-12
add environment variables for aws

kayiwa:


roles/pulibrary.dataspace/tasks/dataspace_config.yml
```diff
@@ -1,19 +1,23 @@
 ---
-- name: oar | Install site configuration
```
jrgriffiniii, line 2: I am very sorry for missing this.

group_vars/dataspace/vault.yml
```diff
@@ -1,38 +1,42 @@
 $ANSIBLE_VAULT;1.1;AES256
```
jrgriffiniii, line 1: This has a placeholder value for the AWS S3 access credentials 👍

PR #1705, opened 2020-08-23
change the hostname

kayiwa:


jrgriffiniii:
(I merged this due to the urgency introduced an emergency authentication issue reported on 08/23/20)

PR #1710, opened 2020-09-03
Fix special collections redirects for exhibit sites

axamei:


kevinreiss (approved):
Correcting the typo in "special_collections" to "special-collections".

PR #1713, opened 2020-09-04
install vector data router to all our VMs

kayiwa:


carolyncole (approved):
I ❤️ that it is tested!

PR #1725, opened 2020-09-14
Redirect libimages1 to iiif.princeton.edu.

tpendragon:
This handles a bunch of complicated rewrites for the fact that Loris
used to be able to handle slashes in the ID, but Cantaloupe follows the
spec appropriately. This can handle going at least 8 directories deep.


kayiwa (approved):
OK

PR #1730, opened 2020-09-21
Stop redirecting libimages2 errors to libimages1.

tpendragon:
There's no reason to redirect anymore - all the files are copied over. This is one step in getting rid of libimages1.


kayiwa (approved):
OK

PR #1731, opened 2020-09-21
Add Memcached

tpendragon:
We'll be able to use this for Figgy web caching.


kayiwa (commented):
pretty sure `roles/pulibrary.memcached/molecule/default/molecule.yml-e` got in by mistake. Tests would be nice if there are any, otherwise OK.

tpendragon:
@kayiwa Added a test and got rid of that file. Looks like maybe that molecule generation task makes it?

kayiwa (approved):
OK

PR #1756, opened 2020-10-01
Refactorama Figgy Molecularization

tpendragon:


roles/pulibrary.figgy/README.md
```diff
@@ -0,0 +1,48 @@
+Role Name
+=========
+
+A brief description of the role goes here.
```
carolyncole, line 4: We should add a description of figgy here

roles/pulibrary.figgy/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Requirements
+------------
+
+Any pre-requisites that may not be covered by Ansible itself or the role should
```
carolyncole, line 9: Did we want to describe the additional items added in the playbook?

roles/pulibrary.figgy/molecule/default/tests/test_default.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+import testinfra.utils.ansible_runner
+
+testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
+    os.environ['MOLECULE_INVENTORY_FILE']
+).get_hosts('all')
+
+
+def test_hosts_file(host):
```
carolyncole, line 10: We should probably add a couple tests, but that would not need to be done here...

roles/pulibrary.figgy/README.md
carolyncole, line 4: We should add a description of figgy here

roles/pulibrary.figgy/README.md
carolyncole, line 9: Did we want to describe the additional items added in the playbooks for workers ect.?

carolyncole (changes requested):
Looking good.  I think we should at the very least update the Readme before a merge.

roles/pulibrary.figgy/molecule/default/tests/test_default.py
kayiwa: added some tests.

carolyncole (approved):

PR #1786, opened 2020-10-21
[libwww] Adding hr-share role to libwww

### LINKED ISSUE #1780: [libwww] Add hr-share to library servers
Allow access to the HR share on the library servers so that we can start using the library staff report to generate the staff directory.

carolyncole:
Gives access to the staff report for building the staff directory

fixes #1780


kayiwa (approved):
OK

carolyncole:
We are moving the logic that will use the share to the lib-jobs server.  I'm going to close this since we no longer need it here.

PR #1789, opened 2020-10-24
add department names attributes for dataspace and oar

kayiwa:


carolyncole (approved):
Looks ok to me, but you likely want someone with more knowledge to review

PR #1809, opened 2020-11-10
Updating to the new git repo names

carolyncole:
We removed drupal and wp from repository names, since these sites were no longer drupal or wordpres sites


kevinreiss (approved):
👍

PR #1821, opened 2020-11-18
allow cleaner lumping of grouped applications

kayiwa:
This is in preparation for monitoring (look towards bottom of file)
  where will need to define specific groups and children to run ansible
  against

this will be needed by #1784 but no reason not to take advantage of this now

### DIFF
--- a/hosts
+++ b/hosts
@@ -1,222 +1,352 @@
-[bibdata_staging]
-bibdata-staging1.princeton.edu
-bibdata-staging2.princeton.edu
-[bibdata]
-bibdata1.princeton.edu
-bibdata2.princeton.edu
-[cicognara_staging]
-cicognara-staging1.princeton.edu
-[cicognara_production]
-cicognara1.princeton.edu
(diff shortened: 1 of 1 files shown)


eliotjordan (approved):

hosts
```diff
@@ -1,222 +1,353 @@
...
-[gcp_dataspace_staging]
-gcp_dataspace_staging1 ansible_host=10.132.0.5
-[gcp_dataspace_production]
-gcp_dataspace_prod1 ansible_host=10.244.0.2
+---
+all:
+  children:
+	bibdata_staging:
+	  hosts:
+		bibdata-staging1.princeton.edu:
```
christinach, line 6: @kayiwa I just opened an issue for that. We use only bibdata-staging2. https://github.com/pulibrary/princeton_ansible/issues/1822

christinach: I don't know if there is another reason that we should keep bibdata-staging1.

PR #1826, opened 2020-11-20
Adding lib-jobs-prod

carolyncole:
Added to configuration
Added to nginx


kayiwa (approved):
OK. Probably want someone with more lib-jobs experience to look it over

PR #1831, opened 2020-11-24
update environment for molecule

kayiwa:
update python, molecule and ansible


.github/workflows/molecule_tests.yml
```diff
@@ -74,6 +62,14 @@ jobs:
           - pulibrary.apache_maven
           - pulibrary.pulmap
           - pulibrary.memcached
+  test-molecule3:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        role:
+          - figgy
```
carolyncole, line 70: I would expect redis, ruby, subversion, vips, nodejs, bind9, freetds, passenger, &  rails-app all to be here...  Was there a reason for no moving them?

Pipfile
```diff
@@ -7,15 +7,14 @@ verify_ssl = true
...
-ansible = "*"
-docker = "*"
 invoke = "*"
 neovim = "*"
 gitpython = "*"
 yamllint = "*"
 flake8 = "*"
 ansible-lint = "*"
+ansible = "*"
+molecule = {extras = ["docker"], version = "*"}
```
carolyncole, line 17: Cool!  This is how you got past the error I was seeing...  Nice!

roles/common/.ansible-lint
```diff
@@ -1,2 +1,3 @@
 skip_list:
   - '303'
+  - '106'
```
carolyncole, line 3: I know I added these, but I wonder if we should remove them?

carolyncole (changes requested):
Just a couple questions on the git action config and if we should keep all the 106 skips.  We can merge this PR and create tickets to remove those also.

.github/workflows/molecule_tests.yml
kayiwa: done.

PR #1836, opened 2020-11-30
[Cicognara] Use Solr8 for Production.

tpendragon:
Progresses https://github.com/pulibrary/cicognara-rails/issues/474

### DIFF
--- a/group_vars/cicognara_production.yml
+++ b/group_vars/cicognara_production.yml
@@ -47,7 +47,7 @@ rails_app_vars:
   - name: GOOGLE_CLIENT_ID
     value: '{{ vault_cicognara_google_client_id }}'
   - name: CICOGNARA_SOLR_URL
-    value: 'http://lib-solr.princeton.edu:8983/solr/cicognara'
+    value: 'http://lib-solr-prod4.princeton.edu:8983/solr/cicognara'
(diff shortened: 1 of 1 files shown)


eliotjordan (approved):

group_vars/cicognara_production.yml
```diff
@@ -47,7 +47,7 @@ rails_app_vars:
   - name: GOOGLE_CLIENT_ID
     value: '{{ vault_cicognara_google_client_id }}'
   - name: CICOGNARA_SOLR_URL
-    value: 'http://lib-solr.princeton.edu:8983/solr/cicognara'
+    value: 'http://lib-solr-prod4.princeton.edu:8983/solr/cicognara'
 sidekiq_worker_name: cicognara-workers
```
christinach, line 51: Why not going through the load balancer `lib-solr8-prod.princeton.edu` ?

tpendragon: Because we've had issues with that being slow in the past haven't we?

christinach: We had with orangelight when we were pulling data from solr for the call_numbers.

christinach: The solr_url for orangelight in the production boxes is pointing `lib-solr8-prod`

PR #1841, opened 2020-12-01
[Figgy] Forward client IP through load balancer.

tpendragon:
We restrict material based on IP, and without this all requests to Figgy
were coming in as the load balancer.


hackartisan (approved):
oh my

PR #1842, opened 2020-12-02
Setup ci molecule3

carolyncole:
Move common and deploy-user to molecule 3 while leaving all of the other roles as molecule 2


kayiwa (approved):
OK

PR #1887, opened 2020-12-03
add bind9 tests and role to molecule3

### LINKED ISSUE #1883: setup molecule3 bind9

kayiwa:
this closes #1883


roles/bind9/molecule/default/verify.yml
```diff
@@ -0,0 +1,24 @@
...
+  - name: test for bind9 packages
+    assert:
+      that:
+        - not pkg_status.changed
+
+  - name: check status of bind9
+    service:
+      name: bind9
+      state: started
+    check_mode: true
```
carolyncole, line 24: Do you need an assert here?

carolyncole (changes requested):

roles/bind9/molecule/default/verify.yml
kayiwa: I was blindly bringing in the tests from the >= 2.22

carolyncole (approved):

PR #1906, opened 2020-12-04
1900 figgy pub sub molecule3

### LINKED ISSUE #1900: setup figgy-pubsub-worker molecule3

kayiwa:
closes #1900


carolyncole (approved):
❤️ Real tests!

roles/solrcloud/tasks/install.yml
```diff
@@ -43,7 +43,7 @@
         solr_config_xml not in solrcloud_check.content)
   retries: 3
   delay: 10
-  until: not solr_downloaded.failed
+  until: solr_downloaded.result == 'success'
```
carolyncole, line 46: @kayiwa this change does not look like it works: `fatal: [instance]: FAILED! => {"msg": "The conditional check 'solr_downloaded.result == 'success'' failed. The error was: error while evaluating conditional (solr_downloaded.result == 'success'): 'dict object' has no attribute 'result'"}`

carolyncole (changes requested):

roles/solrcloud/tasks/install.yml
kayiwa: hey @carolyncole so I looked at the size of this file. It's kinda large and I'm wondering if that is practical to do or even to be a good network citizen responsible to do. I'm thinking this should be :running_on_server'ed:

kayiwa: okay... maybe not 
```
ls -alh ../../local_files/solr/solr-7.7.2.tgz
-rw-r--r-- 1 kayiwa staff 164M Dec  3 13:24 ../../local_files/solr/solr-7.7.2.tgz
```

carolyncole: I think the issue is `solr_downloaded.result` does not exists.  When I looked at it locally there was not result in `solr_downloaded`

carolyncole (approved):

PR #1912, opened 2020-12-04
add rails_app tests and role to molecule3

### LINKED ISSUE #1910: setup rails-app molecule3

kayiwa:
closes #1910


roles/rails_app/molecule/default/playbook.yml
```diff
@@ -6,4 +6,4 @@
     - passenger_listen_port: '8080'
     - node_identifier: 1
   roles:
-    - role: rails-app
+    - role: rails_app
```
carolyncole, line 9: Should this be deleted?

carolyncole (changes requested):
I think you also may need to delete the old tests too.

carolyncole (approved):

PR #1924, opened 2020-12-07
1911 workers molecule3

### LINKED ISSUE #1911: add workers molecule3

kayiwa:
- remove rails-app
- add sidekiq_workers tests and role to molecule3
- add sneakers_workers tests and role to molecule3
closes #1911


roles/sidekiq_worker/molecule/default/verify.yml
```diff
@@ -3,18 +3,16 @@
...
-    assert:
-      that:
-        - not pkg_status.changed
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sidekiq-workers status"
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sidekiq-workers restart"
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sidekiq-workers start"
+    check_mode: true
+    register: present
+    failed_when:
+      - present is changed
```
carolyncole, line 18: The old test used to check for `f = host.file('/etc/systemd/system/sidekiq-workers.service')` Is this still needed?

roles/sneakers_worker/molecule/default/verify.yml
```diff
@@ -0,0 +1,18 @@
...
+      line: "{{ item }}"
+      state: present
+    loop:
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sneakers status"
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sneakers restart"
+      - "deploy ALL=(ALL) NOPASSWD: /usr/sbin/service sneakers start"
+    check_mode: true
+    register: present
+    failed_when:
+      - present is changed
```
carolyncole, line 18: The old tests used to check for `f = host.file('/etc/systemd/system/sneakers.service')` Is this still needed?  I see the tests for sudoers which look great, but the service itself might not exist.

carolyncole (changes requested):

roles/sneakers_worker/molecule/default/verify.yml
kayiwa: these are systemd tests that never ran before molecule3. I elected not to introduce them and issued them with #1926

kayiwa: OK?

carolyncole (approved):

PR #1967, opened 2020-12-09
Converting research data to molecule 3

### LINKED ISSUE #1966: convert researchdata to molecule 3

carolyncole:
fixes #1966


roles/researchdata/molecule/default/verify.yml
```diff
@@ -0,0 +1,31 @@
...
+    check_mode: true
+    register: present
+    failed_when:
+      - present is changed
+  - name: test for postgres database and login
+    command: psql -U researchdata -w researchdata -c '\d'
+    register: postgres_results
+    ignore_errors: true
+    changed_when: false
+  - name: assert cron jobs exists
```
kayiwa, line 28: better label maybe? we are asserting a DB exists.

kayiwa (commented):
a better labeled assert maybe?

roles/researchdata/molecule/default/verify.yml
carolyncole: Updated...

kayiwa (approved):
OK

PR #2016, opened 2020-12-11
update python, ansible and molecule

### LINKED ISSUE #2015: update ansible and molecule 3

kayiwa:
closes #2015


carolyncole (approved):
If we are updating requirements.txt should we remove requirements3.txt and update the CI workflow?  That can be separate...

PR #2021, opened 2020-12-14
Updating the production playbook for capistrano

### LINKED ISSUE #1974: Move PAS to Capistrano deploy
Move pas role to molecule 3
Remove pas_code role
Install any items needed for deploy

carolyncole:
Deleting the pas_code role since it is no longer needed

fixes #1974


kayiwa (approved):
LOOK AT ALL THOSE LINES OF YAML we no longer have to worry about. 😆

PR #2025, opened 2020-12-14
Orangelight playbook updates

christinach:
Changed role dependencies for orangelight to the meta,
changed the order of the postgresql role to make easier to read,
Added orangelight postgres variables in the production


kayiwa (approved):
OK

PR #2036, opened 2020-12-21
update default PHP version to 7.4

### LINKED ISSUE #1877: php 7.2 EOL'ed

kayiwa:
closes #1877

### DIFF
--- a/playbooks/pas_production.yml
+++ b/playbooks/pas_production.yml
@@ -8,7 +8,7 @@
   vars:
     - pas_upload_path: 'pas/pas-production'
   roles:
-    #    - role: roles/datadog
+    - role: roles/datadog
     - role: roles/pas
   post_tasks:
     - name: tell everyone on slack you ran an ansible playbook
--- a/playbooks/pas_production.yml
+++ b/playbooks/pas_production.yml
@@ -7,11 +7,8 @@
     - ../group_vars/pas/vault.yml
   vars:
     - pas_upload_path: 'pas/pas-production'
-  pre_tasks:
-    - set_fact:
-        deploy_id_rsa_private_key: "{{  lookup('file', '../roles/pas_code/files/id_rsa')  }}\n"
   roles:
-    - role: roles/datadog
+    #    - role: roles/datadog
     - role: roles/pas
   post_tasks:
     - name: tell everyone on slack you ran an ansible playbook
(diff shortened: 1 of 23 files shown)


playbooks/pas_production.yml
```diff
@@ -7,11 +7,8 @@
     - ../group_vars/pas/vault.yml
   vars:
     - pas_upload_path: 'pas/pas-production'
-  pre_tasks:
-    - set_fact:
-        deploy_id_rsa_private_key: "{{  lookup('file', '../roles/pas_code/files/id_rsa')  }}\n"
   roles:
-    - role: roles/datadog
+    #    - role: roles/datadog
```
carolyncole, line 11: This needs to be put back in

roles/libwww/tasks/main.yml
```diff
@@ -73,36 +73,36 @@
...
-- name: ensure max file size is set in apache2 php.ini
+- name: libwww | ensure max file size is set in apache2 php.ini
   lineinfile:
-    dest: "/etc/php/7.2/apache2/php.ini"
+    dest: "/etc/php/7.4/apache2/php.ini"
     regexp: "upload_max_filesize ="
     line: "upload_max_filesize = 8M"
 
-- name: ensure memory limit is set in apache2 php.ini
+- name: libwww | ensure memory limit is set in apache4 php.ini
```
carolyncole, line 90: Is this really apache4?

roles/pas/vars/main.yml
```diff
@@ -5,6 +5,7 @@ capistrano_shared_links: []
 capistrano_base_dir: '/opt'
 capistrano_web_owner: 'www-data'
 capistrano_web_service: 'apache2'
+php_webserver: 'apache2'
```
carolyncole, line 8: Did we decide this was not needed

carolyncole (changes requested):
A couple small changes...  Otherwise looks great.

Also who uses the matomo role.  I do not think we tested that.

roles/libwww/tasks/main.yml
kayiwa: stupid perl scripts.

kayiwa:
> A couple small changes... Otherwise looks great.
> 
> Also who uses the matomo role. I do not think we tested that.

I am pretty sure we are shutting that down this month. (I've said that for the last three months.)

carolyncole (approved):

PR #2049, opened 2021-01-05
[Bibdata] Add APPLICATION_URL to group_vars

tpendragon:
Allows for the correct URLs to be generated for reindexing tasks.


kayiwa (approved):
OK

PR #2064, opened 2021-01-12
Changing machine name for fpul-staging1

carolyncole:
set to lib-fpul-staging1 which does not exist

```
diff --git a/roles/nginxplus/files/conf/http/fpul-staging.conf b/roles/nginxplus/files/conf/http/fpul-staging.conf
index 410b6b64..02e775d3 100644
--- a/roles/nginxplus/files/conf/http/fpul-staging.conf
+++ b/roles/nginxplus/files/conf/http/fpul-staging.conf
@@ -3,7 +3,7 @@ proxy_cache_path /data/nginx/fpul-staging/NGINX_cache/ keys_zone=fpul-stagingcac\
... 
-    server lib-fpul-staging1.princeton.edu resolve;
+    server fpul-staging1.princeton.edu resolve;
...
```


kayiwa (approved):
OK

PR #2074, opened 2021-01-19
Adding dom dependency that is no longer there by default in 7.4

carolyncole:
fixes https://github.com/PrincetonUniversityLibrary/pas-craft3/issues/15


kayiwa (approved):
OK

PR #2076, opened 2021-01-19
Allow creation of role with exported variable

carolyncole:
Change the instructions to export a variable and then use it in the following commands so there is not as much thought needed to create a new role


kayiwa (approved):
OK

PR #2108, opened 2021-02-03
Move dataspace google cloud

kayiwa:


group_vars/dataspace/staging.yml
```diff
@@ -11,23 +11,25 @@ java_type: jdk
...
-postgres_admin_user: "{{ db_clusteradmin_user }}"
-postgres_admin_password: "{{ db_clusteradmin_password }}"
+postgres_admin_user: "{{ dspace_db_user }}"
+postgres_admin_password: "{{ dspace_db_password }}"
 postgresql_is_local: false
-postgres_version: 12
-postgres_host: "gcp-postgres-staging1"
+postgres_version: 13
+postgresql_is_cloud: true
+postgres_host: "10.134.16.3"
```
jrgriffiniii, line 32: Should this IP address be migrated into a vault-encrypted file?

kayiwa: I've never understood why we vault them at all beyond (we always did) this is also a non-routable IP that no one can get to who isn't allowed to.

jrgriffiniii (approved):

PR #2118, opened 2021-02-10
ansible tower db

kayiwa:


playbooks/ansible_tower.yml
```diff
@@ -0,0 +1,96 @@
+---
+- hosts: ansible_tower
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/ansible_tower/main.yml
+    - ../group_vars/researchdata/vault.yml
+  tasks:
+    - name: add postgres:10 appstream
```
carolyncole, line 9: I wonder if you could just depend on the postgresql role, instead of having all these commands in here...

carolyncole (changes requested):

playbooks/ansible_tower.yml
kayiwa: Ubuntu and Centos have varying ways of implementing postgres.  Will look into conditional logic.

playbooks/ansible_tower.yml
```diff
@@ -0,0 +1,96 @@
+---
+- hosts: ansible_tower
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/ansible_tower/main.yml
+    - ../group_vars/researchdata/vault.yml
```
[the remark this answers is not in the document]
kayiwa: 🤦🏿

carolyncole (approved):
Nice!  I ❤️  leveraging roles!

PR #2214, opened 2021-04-19
ojs production

kayiwa:


roles/ojs/templates/config.inc.php.j2
```diff
@@ -27,7 +27,7 @@
 
 ; Set this to On once the system has been installed
 ; (This is generally done automatically by the installer)
-installed = On
+installed = Off
```
bess, line 30: @kayiwa Don't we want this set to On?

bess (approved):

PR #2218, opened 2021-04-20
add unvaulted nginx configs

kayiwa:


roles/nginxplus/files/conf/http/api.conf
```diff
@@ -0,0 +1,20 @@
+#
```
carolyncole, line 1: @kayiwa if this file needed?

carolyncole (approved):
One file I am unsure of why it is a part of the PR.  I'll believe you if you say it is...

roles/nginxplus/files/conf/http/api.conf
kayiwa: every time we tunnel to the LB and need to add or remove services using the web UI we need this file.

PR #2236, opened 2021-04-29
Adding a role for special-collections

### LINKED ISSUE #2231: Special Collections Role
We need to pull special collections out or libwww and into it's own role. ref #2230 for the separation reasons

carolyncole:
Also adding a molecule_mariadb role so we can test mariadb roles with molecule

fixes #2231


kayiwa (approved):
nice... the `molecule_mariadb`

PR #2244, opened 2021-05-04
Updating gitattributes to reflect unvaulted files

### LINKED ISSUE #2240: Need to be removed now that they are no longer vaulted?
https://github.com/pulibrary/princeton_ansible/blob/main/.gitattributes#L5-L6

These are causing issues like 
```
ERROR! input is not vault encrypted data/var/folders/y7/39n8wgws7lndd8b36kntt5yh0000gq/T//OGUgi9_lib-sc-staging.conf is not a vault encrypted file for /var/folders/y7/39n8wgws7lndd8b36kntt5yh0000gq/T//OGUgi9_lib-sc-staging.conf
fatal: unable to read files to diff
```

carolyncole:
nginx files are now no longer vaulted.  This causes errors when trying to amend commits or run a git diff.
fixes #2240


kayiwa (approved):
Dang! Sorry y'all! OK

PR #2264, opened 2021-05-14
add file mode option

kayiwa:


christinach (approved):

group_vars/bibdata/production.yml
```diff
@@ -162,7 +162,7 @@ samba_shares:
   - name: marc_liberation_files/campus_access_files
     group: sambashare
     write_list: +sambashare
-    writable: yes
+    writable: true
```
carolyncole, line 165: I wonder if we need to add `owner` here.  I do not understand looking at the code how deploy became the owner and the group of the directory.

carolyncole (changes requested):
I'm unsure how the directory got to be deploy/deploy instead of pulsys/sambshare.  Maybe adding owner for the share to be pulsys would also be good.   It would not explain the deploy/deploy though

group_vars/bibdata/production.yml
kayiwa: this will be done on Maintenance Monday with a different still issued PR.

PR #2265, opened 2021-05-14
Removing campus access from list

carolyncole:
We do not want it owned by deploy as we can not write to it if it is


kayiwa (approved):
OK!

PR #2278, opened 2021-05-24
Update timezome role in orangelight workers

hackartisan:
The role was renamed but hadn't been updated in this playbook


kayiwa (approved):
OK!

PR #2279, opened 2021-05-25
Adjust IP access to CDH geniza sites

rlskoeser:
- add cPanel ip to test site to give access from old PGP site
- restrict production site to campus/vpn+cPanel while in development

Please confirm I did this right! I just copied the other examples

### DIFF
--- a/roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
+++ b/roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
@@ -48,10 +48,7 @@ server {
         health_check interval=10 fails=3 passes=2;
 
         # allow princeton network
-        allow 128.112.0.0/16;
-        allow 140.180.0.0/16;
-        allow 172.20.95.0/24;
-        allow 172.20.192.0/19;
+        include /etc/nginx/conf.d/templates/restrict.conf;
         # allow access from old PGP site on cPanel
         allow 198.199.71.236;
         # block all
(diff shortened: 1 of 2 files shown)


roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
```diff
@@ -46,6 +46,16 @@ server {
         proxy_set_header X-Real-IP $remote_addr;
         proxy_cache geniza_prodcache;
         health_check interval=10 fails=3 passes=2;
+
+        # allow princeton network
+        allow 128.112.0.0/16;
```
kayiwa, line 51: can you replace these 4 lines with `include /etc/nginx/conf.d/templates/restrict.conf;`

roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
```diff
@@ -46,6 +46,16 @@ server {
...
+        allow 140.180.0.0/16;
+        allow 172.20.95.0/24;
+        allow 172.20.192.0/19;
+        # allow access from old PGP site on cPanel
+        allow 198.199.71.236;
```
kayiwa, line 56: you want this vaulted?

roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
rlskoeser: yes, sure!

roles/nginxplus/files/conf/http/cdh_geniza_prod.conf
rlskoeser: should it be? doesn't seem like a big risk to me (even though I'm sure a lot of sites run on reclaim)

it's only being restricted because the production site is in development and we don't want people to look at it or use it yet

kayiwa (approved):
OK!

PR #2283, opened 2021-05-25
Make test-geniza.cdh and geniza.cdh the canonical hostnames

rlskoeser:
configures cdh-test-geniza and cdh-test-geniza as redirect locations so we don't have two different location blocks and two different urls


kayiwa (approved):
OK!

PR #2291, opened 2021-06-02
i11 abid certs

kayiwa:
- add tls certs for abid staging
- add abid staging config

this will close https://github.com/pulibrary/abid/issues/11


roles/nginxplus/files/conf/http/abid-staging.conf
```diff
@@ -0,0 +1,46 @@
+# Ansible managed
+proxy_cache_path /data/nginx/abid-staging/NGINX_cache/ keys_zone=abid-stagingcache:10m;
+
+upstream abid-staging {
+    zone abid-staging 64k;
+    server abid-staging1.princeton.edu resolve;
+    server abid-staging2.princeton.edu resolve;
```
tpendragon, line 7: Is there an abid-staging2? Do we need it?

tpendragon (changes requested):

roles/nginxplus/files/conf/http/abid-staging.conf
kayiwa: nice catch! updated.

tpendragon (approved):

Issue #2290, opened 2021-06-02
Sort out bibdata samba creation conflicts


hackartisan:
follow-up to https://github.com/pulibrary/princeton_ansible/pull/2289

the `bibdata` role has its own samba server / client task files.
It also runs the `hr_share` role which has the `samba` role as a dependency.

The way these roles are running or configured is in conflict such that the directories in `/data/marc_liberation_files` are created with mode 511, which means deploy can't write the files it needs to write.

The above PR hacked in a fix but it would be better to simplify and detangle what's going on here.

Here's an ansible run showing the order of tasks: https://gist.github.com/hackartisan/1f40874c9cd72e5d8d8a9b5eb1acd1d2

PR #2311, opened 2021-06-14
WIP: Work on sorting out bibdata samba clients from servers

hackartisan:
We should not provide yml config when we want the box to be a samba
server. That config gets used by the samba role (a prerequisite
for the hr_shares role) to set up the client connection.

refs #2290


hackartisan:
@kayiwa and I tried a different direction on #2313 for which which we have a sligtht preference.

hackartisan:
And the description on this PR is incorrect. After reviewing the samba role more carefully, those configs are for setting up a share on the server-side, as @carolyncole had initially said.

Since we don't want to do both, and hr_share is working, #2313 removes the bibdata samba_server tasks in favor of the samba role configs, and resolves the directory permissions issues we were previously seeing.

PR #2313, opened 2021-06-15
Align samba share creation in bibdata-related roles / group vars

### LINKED ISSUE #2290: Sort out bibdata samba creation conflicts

hackartisan:
Use a consistent way of creating samba shares (via the samba role) in bibdata-related playbooks.

This has been run with the bibdata-alma-staging and bibdata-alma-production playbooks.

Should probably be squash-merged.

closes #2290
Alternative to #2311


tpendragon (approved):

kayiwa (commented):
OK!

christinach (approved):

PR #2338, opened 2021-06-29
[Bibdata] fix write errors

hackartisan:
fixes pulibrary/bibdata#1457

The smb.conf file was somehow coming up with 511 file and directory permissions even though we had configuration and a default value asking it to do otherwise. We ended up just hard-coding these values because there isn't a use case for any other value on any of our boxes.

Note this has been run on bibdata-alma production boxes.

### DIFF
--- a/group_vars/bibdata/alma_staging.yml
+++ b/group_vars/bibdata/alma_staging.yml
@@ -139,22 +139,16 @@ samba_shares:
     group: sambashare
     write_list: "deploy, +sambashare"
     writable: true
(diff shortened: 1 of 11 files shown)


tpendragon (changes requested):
This looks fine, CI is just failing for bibdata.

hackartisan:
@tpendragon, @kayiwa fixed it; give another look?

kayiwa (approved):
OK!

tpendragon (approved):

PR #2362, opened 2021-07-16
Take alma-dependent services off the load balancer

hackartisan:
During final migration phase
closes pulibrary/bibdata#1492


hackartisan:
this change has been manually put in place on both load balancer machines.

tpendragon (approved):

PR #2364, opened 2021-07-16
Use Isilon mount for files.

tpendragon:


christinach (approved):

hackartisan (commented):
So do staging and prod write to different subdirectories?

tpendragon:
> So do staging and prod write to different subdirectories?

Yes.

hackartisan:
👍🏻 this seems great but I'm not sure why the test is failing

PR #2389, opened 2021-07-26
[AppDeploy] Heaven configuration

tpendragon:


roles/resque_worker/templates/worker.service
```diff
@@ -0,0 +1,22 @@
...
+WorkingDirectory=/opt/{{rails_app_directory}}/current
+ExecStart=/bin/bash --login -c "/usr/local/bin/bundle exec rake resque:work QUEUE=* RAILS_ENV={{rails_app_env}}"
+RestartSec=1
+Restart=on-failure
+StandardOutput=syslog
+StandardError=syslog
+SyslogIdentifier=resque
+
+[Install]
+WantedBy=some-worker.service
```
jrgriffiniii, line 22: Should this be `some-worker`? I apologize, I am not familiar with the server.

jrgriffiniii (approved):

roles/resque_worker/templates/worker.service
tpendragon: Good catch, fixed.

PR #2393, opened 2021-07-27
Write traject logs to datadog

hackartisan:
closes pulibrary/bibdata#1507


eliotjordan (approved):
This looks good to me. Do we just want the traject errors logged, or is that actually a log of all traject events?

hackartisan:
It's the only traject log as far as I can tell.

hectorcorrea:
The [logging level](https://github.com/traject/traject/blob/master/doc/settings.md#logging-and-progress) (e.g. info, error) probably depends on how the setting is defined on the server:

PR #2408, opened 2021-08-02
[Postgres] Stop configuring pg hba

### LINKED ISSUE #2405: Remove pg_hba configuration from postgres roles
Right now if we run the postgres roles on the servers it'll wipe pg_hba which is configured by other playbooks, resulting in everything being unable to access postgres.

tpendragon:
Closes #2405


carolyncole (changes requested):
@tpendragon I'm fairly certain the postgres role sets up the pg_hba file correctly for a number of roles.  I know I do not touch it, but running the playbook fixed pg_hba errors for at least approvals and ReCAP.

Maybe a switch to turn the default off?

tpendragon:
@carolyncole I don't really see how that's possible - maybe we can look at it together? I don't see the variable `postgresql_hba_entries` configured with any IPs anywhere.

acozine:
This PR also got swamped by the changes in #2782. I'm going to close the PR, keep the issue open, and make sure we address it as part of the postgresql refactor and the postgresql migration epic.

PR #2413, opened 2021-08-04
Uses bibdata-staging (instead of bib data production)

hectorcorrea:
Makes sure Orangelight QA points to bibdata **staging** instead of bibdata production.


hackartisan (approved):
I think this is what we want for right now.

Once we get the sandbox on Aug 11th and provision new orangelight staging infrastructure, then we'll want catalog-staging to point to bibdata-staging and catalog-qa to point to bibdata.

PR #2424, opened 2021-08-09
[Bibdata] Move RAILS_ENV to start of poller execution

### LINKED ISSUE #2427: Run 'systemctl daemon-reload' to reload units
I updated the bibdata-sqs-poller service and ran the playbook to get the changes in place. The restart handler showed that it ran successfully. However, I didn't see the changes in effect when I ran `ps | aux` on the box. So I ran the restart on the box and it said

```
$ sudo service bibdata-sqs-poller restart
Warning: The unit file, source configuration file or drop-ins of bibdata-sqs-poller.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

This should be accounted for in the handler. There may be other affected handlers.

hackartisan:
closes pulibrary/bibdata#1586
closes #2427

### DIFF
--- a/hosts
+++ b/hosts
@@ -42,8 +42,6 @@ gcp_oar_prod1 ansible_host=10.244.0.3
 [gcp_oar_staging]
 gcp_oar_staging1 ansible_host=10.132.0.4
 [bibdata_staging]
-bibdata-staging2.princeton.edu
-[bibdata_staging]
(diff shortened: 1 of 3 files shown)


christinach (approved):

kayiwa (approved):
OK

PR #2455, opened 2021-09-01
confluence staging

kayiwa:
- add postgresql variables for staging
- use local tasks to install postgresql
- add tests
- add postgresql variables for staging
- use local tasks to install postgresql
- add tests


roles/confluence/defaults/main.yml
```diff
@@ -1,3 +1,7 @@
 ---
 # defaults file for roles/confluence
-confluence_version: "7.0.4"
+confluence_version: "7.2.0"
```
carolyncole, line 3: I'm curious if this will upgrade prod when it runs there, or cause any issue.

carolyncole (approved):
Looks ok to me.

roles/confluence/defaults/main.yml
kayiwa: When I started this PR we didn't have a need for an emergency patch. I thought it was have been a slow one. Updated it.

PR #2469, opened 2021-09-07
Updating openjdk to allow for use of newer versions

carolyncole:


carolyncole:
This may be related to DSS.

roles/openjdk/vars/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 # vars file for roles/openjdk
 _java_default_version:
-  Ubuntu-19: 13
+  Ubuntu-19: 17
```
acozine, line 4: This seems merge-able if we don't update the default version:
```suggestion
  Ubuntu-19: 13
```

PR #2473, opened 2021-09-09
[MUDD-DBS] Fix links when on the local machine

carolyncole:
Also allow staging to be the rails staging environment


kayiwa (approved):
OK

PR #2477, opened 2021-09-14
Pathe is a drupal site not a static site

carolyncole:
We had to revert this change, because the search did not work in the static site

https://github.com/pulibrary/pathebaby/issues/3


kayiwa (approved):
OK

PR #2499, opened 2021-10-01
change variables to add hostname

kayiwa:
update TLS to use a SAN key for both ojs-dev and ojs-dev1


roles/common/tasks/main.yml
```diff
@@ -1,4 +1,23 @@
 ---
+- name: Remove specified repository from sources list
+  ansible.builtin.apt_repository:
+    repo: 'deb https://oss-binaries.phusionpassenger.com/apt/passenger {{ ansible_distribution_release }} main'
```
carolyncole, line 4: @kayiwa Looks like you need a sane default for ansible_distribution_release in common.  This is failing CI

carolyncole (changes requested):

PR #2506, opened 2021-10-06
remove passenger repo so it can be reinstalled

kayiwa:


hackartisan:
@kayiwa is there an issue for this work? I'm curious what's going on.

PR #2553, opened 2021-11-02
really restart nginx

kayiwa:


carolyncole (approved):

roles/nginxplus/.ansible-lint
```diff
@@ -0,0 +1,8 @@
+skip_list:
+  - 'ANSIBLE0002'
+  - 'ANSIBLE0006'
+  - 'ANSIBLE0010'
+  - ANSIBLE0016
```
acozine, line 5: Should this also have single quotes?

playbooks/nginxplus_production.yml
```diff
@@ -15,3 +15,18 @@
...
         channel: #server-alerts
+
+- hosts: nginxplus
+  remote_user: pulsys
+  strategy: linear
+  become: true
+
+  tasks:
+    - name: nginxplus | restart nginx for realsies
+      command: systemctl restart nginx
```
acozine, line 26: Would the service module work here? instead of `command`?

playbooks/nginxplus_production.yml
```diff
@@ -15,3 +15,18 @@
...
+  post_tasks:
+    - name: tell everyone on slack you ran an ansible playbook
```
hackartisan, line 28: Do we need this twice? Maybe delete the earlier one?

hackartisan (approved):

PR #2575, opened 2021-11-19
Configure stall time to be very high.

tpendragon:
Figgy was timing out on large updates.

This has been deployed, as we're unsure if it'll fix our issue.

Refs https://github.com/pulibrary/figgy/issues/4802


acozine:
@tpendragon did this fix the issue? I'm happy to merge it if so.

tpendragon:
It did!

Issue #2585, opened 2021-11-30
Install shibboleth via ansible


bess:
The shibboleth role should:
- [x] Install all needed software and configuration files
- [x] Provide instructions for how to request access from OIT
- [x] Provide instructions for how to test the connection
- [x] Document and provide to OIT which shibboleth attributes are required for Vireo
- [x] Get our installation authorized by OIT 

Dec 8 update:
Our shibboleth installation is working and we've asked OIT to authorize it. Awaiting a response from them right now.

Dec 10 update: 
Jay at OIT says this last batch of attributes was the list he was looking for. He is working on this today.

bess:
My response to OIT (copied here for reference):

Hi, Jay.

1) These are the shibboleth attributes we need:
"SHIB_netid"
"SHIB_mail"
"SHIB_uin"
"SHIB_givenname"
"SHIB_initials"
"SHIB_sn"
"SHIB_dateOfBirth"
"SHIB_eduPersonAffilation"
"SHIB_phone"
"SHIB_postal"
"SHIB_mail"
"SHIB_permanentPhone"
"SHIB_permanentPostal"
"SHIB_permanentMail"
"SHIB_degree"
"SHIB_department"
"SHIB_college"
"SHIB_major"
"SHIB_gradYear"
"SHIB_gradMonth"
"SHIB_orcid"

2) We are already using SSL for connection, so I don't think we need to further encrypt the traffic, but I am happy to follow OIT's guidance here.

3) The Login can be tested at https://vireo-staging.princeton.edu/Shibboleth.sso/Login

Thank you!
Bess Sadler

bess:
As of 14 Dec, we are able to do a *partial* authentication to shibboleth. However, it is returning the user to the wrong url. After troubleshooting with @kayiwa and Jay at OIT, here is the diagnosis:

```
using this URL https://vireo-staging.princeton.edu/Shibboleth.sso/Login
I was able to see the application hit our IDP

Within the SAML request: the ACS URL is: https://vireo-staging/Shibboleth.sso/SAML2/POST

To fix this, you need to update your application to use the correct ACS URL of https://vireo-staging.princeton.edu/Shibboleth.sso/SAML2/POST which is within your metadata file.

Thanks, Jay
```

[...]

PR #2592, opened 2021-12-03
When we use port 443 we also need https

bess:
Connected to https://github.com/pulibrary/princeton_ansible/issues/2585


kayiwa (approved):
OK

PR #2597, opened 2021-12-07
Update apache

acozine:
Creates a new group for apache servers (all Ubuntu-based).

Adds a playbook to upgrade the version of apache on those hosts.


kayiwa (approved):
OK

acozine:
Note for Future Us: 
The original Apache bug was fixed in the upstream version 2.4.49, which was released for Bionic as version 2.4.29. See https://ubuntu.com/security/CVE-2021-42013 for details.

PR #2598, opened 2021-12-07
use apt module

kayiwa:
the package module does not have update cache

Co-authored-by: Alicia Cozine<acozine@users.noreply.github.com>


acozine:
This was an indentation issue - update cache is a parameter of the apt module instead of a task-level attribute.

acozine:
Closing because something went wrong with the rebase here, and a 4-line change ballooned to 44 commits changing 138 files. See #2600 for a clean PR with this same content.

PR #2562, opened 2021-11-08
remove digitalocean repo

kayiwa:

PR #2602, opened 2021-12-08
remove digitalocean repo

acozine:
Supersedes #2562, which got a bunch of merge commits in it by mistake.

Updates the mirrors for mariadb. The old mirror disappeared, which was causing apt commands to fail.

### DIFF
--- a/roles/drupal/tasks/main.yml
+++ b/roles/drupal/tasks/main.yml
@@ -168,7 +168,7 @@
 - name: install npm for git actions
   apt:
     name: ["nodejs-dev", "node-gyp", "npm", "libssl1.0-dev"]
-    state: present
+    state: build-dep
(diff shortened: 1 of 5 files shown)


acozine:
Second commit should get the test failure passing. Don't merge this, though, until we understand why `no_log` was set to `false` here: https://github.com/pulibrary/princeton_ansible/commit/61abf4ceba3f605d6771131aa27f0bd311b3c3db#diff-4edf8ed0965ef589b744dd57d77f0a9f0867d51e27e4617f5983535c6ca48618R33

acozine:
Per @carolyncole the `no_log` change was likely left over from testing. It was not an intentional change and can/should be reverted.

kayiwa (approved):
OK

PR #2607, opened 2021-12-10
Update the timeout for maps to something very large

carolyncole:
Should fix the 'upstream timed out (110: Connection timed out) while reading response header from upstream'


kayiwa (approved):
OK

PR #2610, opened 2021-12-10
Update the timeout for libwww to something very large

carolyncole:
Should fix the 'upstream timed out (110: Connection timed out) while reading response header from upstream'


carolyncole:
Drafted for Monday

christinach (approved):

PR #2615, opened 2021-12-10
add pudl config and redirects

kayiwa:


roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
```diff
@@ -0,0 +1,307 @@
+# arks redirects
+    rewrite ^/objects/(.*)$ /objects/$1 redirect;
+    location /objects/ {
+        proxy_pass https://arks.princeton.edu/ark/88435/;
```
escowles, line 4: Is this a proxy or a redirect? I was expecting to redirect them, but I don't know nginx config, so I'm not sure what this is doing.

roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
```diff
@@ -0,0 +1,307 @@
...
+    location /paris {
+        proxy_pass https://dpul.princeton.edu/marquand-library;
+    }
+
+    rewrite ^/collections/scheide(.*)$ /scheide$1 redirect;
+    location /scheide {
+        proxy_pass https://dpul.princeton.edu/scheide;
+    }
+
+    rewrite ^/collections/versailles(.*)$ /versailles$1 redirect;
```
escowles, line 54: Will this handle both `/collections/versailles2` and `/collections/versailles3`?

kayiwa: it should. Only way to test is after the fact. Unfortunately. :-(

acozine:
Rebasing should make the tests pass, since those failures have been fixed.

roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
kayiwa: if this configuration lived on pudl dot princeton dot edu then a redirect would do. It lives on lib-adc dot princeton dot edu and proxy pass will work more reliably.

roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
```diff
@@ -1,307 +1,130 @@
...
+
+    rewrite ^/collections/garden(.*)$ https://dpul.princeton.edu/marquand-library$1 redirect;
+
+    rewrite ^/collections/papyri(.*)$ https://dpul.princeton.edu/papyri$1 redirect;
+
+    rewrite ^/collections/paris(.*)$ https://dpul.princeton.edu/marquand-library$1 redirect;
+
+    rewrite ^/collections/scheide(.*)$ https://dpul.princeton.edu/scheide$1 redirect;
+
+    rewrite ^/collections/versailles2$ https://dpul.princeton.edu/ga_treasures$1 redirect;
```
escowles, line 24: Thanks for splitting out the two versailles patterns — now I don't see a `(.*)` pattern in them, so is the `$1` in the destination still needed?

kayiwa: probably not... incidentally the proxy_pass was not necessary (sorry about that). I've made the name switch so you can test these and comment on broken items here.

PR #2624, opened 2021-12-16
[NGINXPLUS] Changing the lockers pages on staging

carolyncole:
on staging only (for the moment) change to the new system.
The old proxy to the windows server will still be utilized in prod


kayiwa (approved):
OK

kevinreiss (approved):

PR #2629, opened 2021-12-17
Fix shibboleth redirect issue

bess:
If ServerName and UseCanonicalName are not set, then shibboleth
redirection will not work.


kayiwa (approved):
OK

PR #2633, opened 2021-12-21
add the lib-postgres-prod2

kayiwa:
add the hosts name and the variables needed for replicant and the
standby server


group_vars/postgresql/production.yml
```diff
@@ -2,20 +2,26 @@
...
-  - name: test_db
+  - name: replicant
 postgresql_users:
   - name: postgres
     password: "{{ vault_postgres_admin_password }}"
     db: "postgres"
     priv: "ALL"
     role_attr_flags: "SUPERUSER"
+  - name: replicant
+    password: "{{ vault_postgres_admin_password }}"
```
acozine, line 21: Should this be `"{{ vault_replicant_password }}"` or `"{{ replication_user_password }}"`? When we looked at the staging server, the correct database pw was in use, so this may take some more investigation.

PR #2636, opened 2021-12-21
adds note about apache config settings and link to vireo example

### LINKED ISSUE #2634: shibboleth role apache
Add information in the shibboleth role describing the need to complete the apache2 dependency which will require use of UseCanonicalName == On

see vireo --> templates --> vireo.conf.j2 for example.

acozine:
Closes #2634.

Adds a note for future users who need to install Shibboleth about the apache config requirements, with a link to the example in the Vireo role.


kayiwa (approved):
OK

PR #2641, opened 2021-12-29
add pemm.cdh.princeton tls certs

kayiwa:
also correct the whitespace in dpul


acozine:
Playbook succeeded, both sites look good.

PR #2578, opened 2021-11-22
add lib-elixir TLS certs

kayiwa:
Co-authored-by: Christina Chortaria <actspatial@gmail.com>

check for cert match with the following command:

```bash
echo “--Certificate:” && openssl x509 -noout -modulus -in lib-elixir_princeton_chained.pem && echo “--Key:” && openssl rsa -noout -modulus -in lib-elixir_princeton_edu_priv.key
“--Certificate:”
Modulus=FAFE90B150413F1E56CB562FB83AB3C66C9F1C9E0B63779762A71E3C4CEB37748DCB3FC21983C24E60EE91124F7689CCAD54E467EAC3AC3963A29833899C9D8AF8DD08E07FF366AE238405D5D582801BA8D95CA8D6A9FA1122B5862A4CB2FD34D814B1CE56E24717B47F3C922EEB3D135C090C7AA8AFC56FD2A13F57ACFEA264686AA086D72B0F13E2A40C00B6FA73D94FEFC49DCA9151B18CB663DAB7725560DE8AEBDCEAE5B8D0B3D64389BE09FBD1DF3FFE474B7A307BAD9CAA2BF7F0E83509E072551180D4196CF4E0940EE2329479D58644FD3F8588327E294AECA58277F0ADB2F07EF84EF8E0C73660A7FB51690113AF9666EF0CCF75E488BEB4E727E7
“--Key:”
Modulus=FAFE90B150413F1E56CB562FB83AB3C66C9F1C9E0B63779762A71E3C4CEB37748DCB3FC21983C24E60EE91124F7689CCAD54E467EAC3AC3963A29833899C9D8AF8DD08E07FF366AE238405D5D582801BA8D95CA8D6A9FA1122B5862A4CB2FD34D814B1CE56E24717B47F3C922EEB3D135C090C7AA8AFC56FD2A13F57ACFEA264686AA086D72B0F13E2A40C00B6FA73D94FEFC49DCA9151B18CB663DAB7725560DE8AEBDCEAE5B8D0B3D64389BE09FBD1DF3FFE474B7A307BAD9CAA2BF7F0E83509E072551180D4196CF4E0940EE2329479D58644FD3F8588327E294AECA58277F0ADB2F07EF84EF8E0C73660A7FB51690113AF9666EF0CCF75E488BEB4E727E7
```

then check for validity of CN and expiration with the following command:

```bash
openssl x509 -in lib-elixir_princeton_chained.pem -text
Certificate:
    Data:
        Version: 3 (0x2)
[...]

PR #2642, opened 2021-12-29
Remove lib elixir

acozine:
Related to #2578.

Since we are not using the lib-elixir site, we did not renew the SSL certificate for it. However, the site is still configured on the load-balancer. This PR would decommission the site entirely:

- [ ] removes the site config from the load balancer
- [ ] removes the VM from the hosts file

This PR should remain a draft until the full DLS team can review.


hackartisan (changes requested):
I think we could also remove the playbook and the associated group_vars files: https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/elixir_test.yml

Also the elixir role.

Block pending confirmation of all this with DLS team.

acozine:
Confirmed at standup today: this project, including the VM, the role, and the website, is now decommissioned and all trace of it can be removed.

hackartisan (approved):
🎉

acozine:
Ran the playbook today and the lib-elixir.princeton.edu site is gone from the load balancer, though it lives on in the cache of one of my browsers.

acozine:
Also needed to decommission the IPs, remove alias (via OIT), and remove associated firewall rules. Taking care of those today.

PR #2647, opened 2022-01-06
Revise nginxplus playbook and role

### LINKED ISSUE #2639: Add a tag to nginxplus role for certificates
The nginxplus role is long and complicated. It would be great to add a tag for `update_certs` or something similar, so that when we need to update an SSL certificate, we can do that quickly and easily, without running the entire nginxplus role.

See #2366 for an example.

acozine:
Closes #2639 

The current nginxplus playbook installs datadog and adds the deploy user (plus a bunch of common setup tasks), then installs nginxplus from scratch. This is overkill for the kinds of updates we do regularly, and it's slow to run.

To address these challenges, this PR:
- [ ] renames the current nginxplus playbook, with the datadog and common roleas `nginxplus_production_rebuild` for disaster recovery/migration use cases
- [ ] modifies the playbook named `nginxplus_production.yml` to focus on updates only
- [ ] adds tags to make uploading new SSL keys and certs quick and easy (`SSL` for specific tasks; `always` for restarts at the end)
Once this is merged, we should be able to run `ansible-playbook playbooks/nginxplus_production.yml -t SSL` to upload new/replaced SSL certs and keys.

If possible, let's wait to merge this until we can test it in the staging environment we're creating.

### DIFF
--- a/playbooks/nginxplus_production.yml
+++ b/playbooks/nginxplus_production.yml
@@ -2,6 +2,7 @@
 - name: update nginx
   hosts: nginxplus
   remote_user: pulsys
+  strategy: linear
   become: true
 
 # updates existing load balancer
--- a/playbooks/nginxplus_production.yml
+++ b/playbooks/nginxplus_production.yml
@@ -11,6 +11,7 @@
 - name: restart nginx and announce success
   hosts: nginxplus
   remote_user: pulsys
+  strategy: linear
   become: true
 
   tasks:
--- a/playbooks/nginxplus_production_rebuild.yml
+++ b/playbooks/nginxplus_production_rebuild.yml
@@ -22,6 +22,7 @@
       service: 
         name: nginx
         state: restarted
+      tag: always
 
   post_tasks:
     - name: tell everyone on slack you ran an ansible playbook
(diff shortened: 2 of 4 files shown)


playbooks/nginxplus_production.yml
```diff
@@ -1,17 +1,16 @@
...
   roles:
-    - role: ../roles/datadog
-    - role: ../roles/deploy_user
     - role: ../roles/nginxplus
 
-- hosts: nginxplus
+- name: restart nginx and announce success
+  hosts: nginxplus
   remote_user: pulsys
-  strategy: linear
```
hackartisan, line 14: I think we still want the linear strategy on the restart so that both load balancer boxes don't go down for restart at the same time, which results in a downtime blip for all our sites.

hackartisan (changes requested):
this looks really great!! I had one concern, inline.

playbooks/nginxplus_production.yml
acozine: Linear is the [default strategy](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html). We can reinstate the line for clarity, though.

hackartisan: Oh, interesting I didn't know that.

well if we're running it all as linear than I guess it's not needed. Feel free to remove it from both if you like and I'll go ahead and approve so I'm not a blocker either way.

hackartisan (approved):

PR #2658, opened 2022-01-18
Updating orangelight ruby to 2.75

### LINKED ISSUE #2676: Remove mailcatcher from alma production
mailcatcher is referenced twice once with false and then true. We dont want mailcather in prod.

carolyncole:
Adding molecule testing in for the orangelight role

fixes #2676. @tpendragon any insight into why you did this would be appreciated


christinach (approved):
will merge this tomorrow.

PR #2669, opened 2022-01-20
Allow ws proxying for vireo

bess:
This fixes an error where the vireo application can't access javascript
via the ws protocol

Work toward https://github.com/pulibrary/rdss-catchall/issues/32


kayiwa (approved):
OK

PR #2670, opened 2022-01-20
Adds a role for studio processing

kayiwa:
we are adding the studio processing machine to be managed via ansible


roles/lib_proc4/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
+---
+# tasks file for lib_proc4
+- name: Create mount directories
+  file:
+    path: "/mnt/{{ item }}"
+    state: "directory"
+    mode: 0755
+  with_items:
```
acozine, line 8: Use `loop` here instead of `with_*`

roles/lib_proc4/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    - studio.new.smb.credentials
+
+- name: Create mount to diglibdata shares
+  mount:
+    path: "/mnt/diglibdata/{{ item.path }}"
+    src: "//diglibdata1.princeton.edu/{{ item.src }}"
+    fstype: cifs
+    opts: "credentials=/etc/{{ item.opts }}.smb.credentials"
+    state: mounted
+  loop:
```
acozine, line 38: yay! loop!

roles/lib_proc4/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    - {path: "archives", src: "archives", opts: "archives"}
+    - {path: "archives_bd", src: "archives_bd", opts: "archives_bd"}
+    - {path: "ddd", src: "ddd", opts: "ddd"}
+    - {path: "historicalperiodicals", src: "historicalperiodicals", opts: "historicalperiodicals"}
+    - {path: "princetonperiodicals", src: "princetonperiodicals", opts: "princetonperiodicals"}
+    - {path: "pudl", src: "pudl", opts: "pudl"}
+    - {path: "pul_store", src: "pul_store", opts: "pul_store"}
+    - {path: "ingest_scratch", src: "hydra/binaries/ingest_scratch", opts: "ingest_scratch"}
+  become: true
+  ignore_errors: true
```
acozine, line 49: remove before merging - we want to know if this fails!

roles/lib_proc4/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: Create mount to libimages bluemountain shares
```
acozine, line 51: update the name here? I don't think this is related to bluemountain

acozine (commented):
Overall looks good, I had a few questions, including several for my own education.

acozine:
Also, I might change the title of this PR, and/or move the commit adding the CIFS utilities into a separate PR, since that is only a very small part of this overall change-set.

roles/lib_proc4/.travis.yml
```diff
@@ -0,0 +1,29 @@
+---
```
acozine, line 1: We're not using Travis. Can we remove this file?

acozine (commented):
This looks like a very focused role.

I would suggest renaming the role. I'd call it "StudioProcessing" or "PhotoProcessing" or something like that. Naming the role after the work being done (not after the server where we currently do that work) makes sense because:

1. We are moving toward a new naming standard for VMs, so the server name may change
2. We may need multiple servers for this work in the future
3. The playbook is called `studio_proc.yml`, so calling the role that (or calling them both a variant on that name) would match our current naming convention in this repo

hosts
```diff
@@ -360,3 +360,4 @@ lib-sc-staging1.princeton.edu
 lib-sc-prod1.princeton.edu
 [vireo_production]
 vireo-prod1.princeton.edu
+lib-proc4.princeton.edu
```
acozine, line 363: Right now this is adding `lib-proc4` to the `vireo_production` group. I don't think that's what you intended. It won't matter for the studio proc role, as currently set up, since you're calling the individual host by name. But when someone runs the vireo role we may see weird behavior on the studio processing server.

I'm not normally a fan of single-server groups, but in this case, the name of the server doesn't immediately tell me what it's for (library processes could mean a lot of things), so I'd create a group for it.

playbooks/studio_proc.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: lib-proc4.princeton.edu
```
acozine, line 2: If you do create a group, update this line to match.

roles/studio_proc/molecule/default/verify.yml
```diff
@@ -0,0 +1,15 @@
...
+  - name: test for mount directories
+    stat:
+      path: "/mnt/{{ item }}"
+    register: smb_dir
+    loop:
+      - libimages
+      - diglibdata
+      - hydra_sources
+    failed_when:
+      - smb_dir.stat.isdir is not defined
```
acozine, line 15: Will this test fail if only two of the three Samba directories get mounted correctly?

acozine (commented):
Overall looks good, just a couple of questions/comments.

acozine (commented):
Hosts change looks great, thanks! 

I still don't know for sure if the verification test would pass or fail if we had only 2 of the 3 directories. It probably doesn't matter hugely for this test, but it would be a good learning experience for me to work through this. Maybe we can pair early next week and take a look. 

Otherwise the PR looks great.

acozine:
Rebased and force-pushed today.

PR #2679, opened 2022-01-25
Remove figgy 2 and 4 from the load balancer

hackartisan:
this is temporary until their firewall rules are fixed
refs #4916


hackartisan:
Firewall rules were updated.

PR #2686, opened 2022-01-26
Remove archivesspace role and more

### LINKED ISSUE #2622: Remove archivesspace role
ArchivesSpace is now hosted, and things seem to be working well.

Can we remove this role now?

acozine:
Closes #2622.

Removes the archivesspace role, group vars, testing references, playbooks, and host listing. These are separate commits, in case we need/want to keep any of those "extra" items when we delete the role.


kayiwa (approved):
OK

PR #2687, opened 2022-01-26
Fix failing tests

### LINKED ISSUE #2683: appdeploy is failing the  main build
https://github.com/pulibrary/princeton_ansible/runs/4951939817?check_suite_focus=true

Seems like forever is no longer being installed.

acozine:
Closes #2683.
Fixes two recent test failures:

Error: linting failed in the figgy role because we were upgrading a package without any checks and balances
Fix: add a version parameter to the task

Error: verification failed in the appdeploy role because the verification test required a specific version of `forever`, but the install task installed the latest version, so as soon as a new version was released, the test started to fail
Fix: change the test to be version-agnostic

### DIFF
--- a/roles/appdeploy/molecule/default/verify.yml
+++ b/roles/appdeploy/molecule/default/verify.yml
@@ -4,7 +4,7 @@
   gather_facts: false
   tasks:
   - name: Forever installed
-    shell: 'forever --version | grep v4.0.1'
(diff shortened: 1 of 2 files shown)


hackartisan (approved):
🎉  thank you!

kayiwa (approved):
OK

Issue #2685, opened 2022-01-26
Deduplicate role inclusion for molecule testing?


acozine:
We're running dependencies like the `common` role multiple times per test run for some projects. For example, when the tests run for the `lib_jobs` role, molecule runs:
- include lib_jobs
- common
- ruby
- common (second time)
- deploy_user
- common (third time)
- deploy user (second time)
- mariadb
- redis
- common (fourth time)
- common again (fifth time)
- bind9
- common (sixth time)
- deploy_user (third time)
- common (seventh time)
- ruby (second time)
- passenger
- postgresql
- nodejs
- rails_app
- sidekiq_worker
- common (eighth time)
- common again (ninth time)
- bind9 (second time)
- common (tenth time)
- deploy_user (fourth time)
- common (eleventh time)
- ruby (third time)
- passenger (second time)
and so on until I got bored and stopped scrolling, see https://github.com/pulibrary/princeton_ansible/runs/4957465231?check_suite_focus=true for the full run.

In production, of course, the duplication may be needed where different roles run on different servers, so the dependencies must run on each server, but it seems inefficient for testing. Can anything be done to streamline the test runs?

hackartisan:
it's very curious. shouldn't ansible deduplicate the meta.yml entries and only run each once?

https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#running-role-dependencies-multiple-times-in-one-play

acozine:
I don't think Ansible deduplicates down the nested tree, and we have a lot of nested dependencies. For example, the dependency tree for the `lib_jobs` role looks like this:

- ruby (which depends on common)
- deploy_user (which also depends on common)
- mariadb (which depends on deploy_user ((which in turn depends on common)) )
- redis (no dependencies!!!)
- sidekiq_worker (which depends on rails_app ((which in turn depends on common, bind9, deploy_user, passenger, postgresql, nodejs. . . most of which have dependencies of their own)) )
- rails_app (which depends on common, bind9, deploy_user, passenger, postgresql, nodejs . . . see above)
- hr_share (which depends on samba)

hackartisan:
This makes me think we should move away from using meta for dependencies and toward just putting everything in playbooks.

acozine:
See also [this SO post](https://stackoverflow.com/questions/58411630/molecule-tests-for-ansible-roles-how-to-structure-to-avoid-duplication).

hackartisan:
[...]

PR #2693, opened 2022-01-27
Dedup tests

acozine:
First attempt / first steps toward streamlining Molecule tests as discussed in #2685. 

This branch uses the `openjdk` role as a test case, because it's small, has only one dependency, and is used in several playbooks. This branch:
- removes the dependency on the `common` role from the `openjdk/meta/main.yml` file
- explicitly adds the `common` role to every playbook that calls the `openjdk` role
- explicitly adds the `common` role to the tests for the `openjdk` role


acozine:
Note to Future Me:
The `openjdk` role listed a dependency on `common`.
The `saxon` role listed a dependency on `openjdk`, with no other depedencies.
When I deleted the dependency in the `openjdk` role, the `saxon` role tests failed with a missing package failure.

I'm betting this means that the `saxon` tests need to run the `common` role, and did not. So I've added a dependency on `common` to the `saxon` role (and also to the `zookeeper` role, which had the same single dependency on `openjdk`). If both pass CI, I will understand more about how all this fits together.

jrgriffiniii (approved):

bess (approved):

hackartisan (approved):

roles/openjdk/meta/main.yml
```diff
@@ -14,5 +14,5 @@ galaxy_info:
         - 16.04
         - 18.04
 
-dependencies:
-  - {role: common}
+dependencies: []
+#   - {role: common}
```
eliotjordan, line 18: suggestion: Remove the commented-out line.

eliotjordan (changes requested):

roles/openjdk/meta/main.yml
acozine: good idea, thanks

acozine:
I updated the ADR to reflect the agreement we would be making if/when we merge this PR. The change is still under discussion, though!

roles/openjdk/meta/main.yml
acozine: addressed in c24073f756c948537a461764eb8b41b029186f49

acozine:
@carolyncole is concerned about the maintenance burden in playbooks, and specifically about the possibility of drift between staging and production playbooks. This led to a conversation about the way the repo currently handles groups in the hosts file, group_vars, role variables, and playbooks. Can we have a single playbook per project, with both the target hosts and the variables determined at runtime (for example, something like `ansible-playbook playbooks/pulmap.yml -e env=prod` and `ansible-playbook playbooks/pulmap.yml -e env=staging` instead of separate playbooks `pulmap_production.yml` and `pulmap_staging.yml`? This would reduce both maintenance and drift.

hackartisan:
providing the stage at playbook runtime seems fine to me.

acozine:
Recent changes to our CI have made improvements to this problem - see https://github.com/pulibrary/princeton_ansible/runs/6040136600?check_suite_focus=true for a run in which the `common` role was NOT run multiple times!!!

hackartisan:
> Recent changes to our CI have made improvements to this problem

What were the changes?

PR #2698, opened 2022-01-29
add mysql role

kayiwa:


roles/mysql/molecule/default/destroy.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Destroy
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  no_log: "{{ molecule_no_log }}"
+  tasks:
```
acozine, line 7: Does this play destroy anything? It looks like it dumps the config and leaves the docker container running. Is that on purpose? I'll take a look at some other `destroy.yml` files, but I'll leave this comment here as a reminder.

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+    install_recommends: false
+    update_cache: true
+
+- name: MySQL | Install python database client packages
+  pip:
+    name: mysqlclient
+  when:
+    - running_on_server
+
+- name: MySQL | install mysql
```
acozine, line 28: I'd add `server` to the task name and to the variable here for clarity ("install mysql server" and `mysql_server_packages`).

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+  ansible.builtin.apt:
```
acozine, line 29: In this line it's `ansible.builtin.apt`; in line 16 above it's `apt`. For consistency, let's either use FQCN  or the bare module names everywhere (at least in this role).

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+
+- name: MySQL | configure mysql client
+  community.general.ini_file:
+    dest: "{{ mysql_configuration_destination }}/{{ mysql_name }}-client.cnf"
+    section: client
+    option: socket
+    value: "{{ mysql_socket }}"
+    mode: "0644"
+  when:
+    - mysql_server
```
acozine, line 63: Do we want to restrict this task? We install the client packages even if `mysql_server` isn't set. Do we also want to configure the client in that case?

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+  loop: "{{ mysql_databases }}"
+  when:
+    - mysql_databases is defined
+    - running_on_server
+
+- name: MySQL | create users on server
+  community.mysql.mysql_user:
+    name: "{{ item.name }}"
+    state: present
+    password: "{{ item.password | default(omit) }}"
```
acozine, line 163: This would allow creating a user without a password. What is the use case for that?

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+    login_user: root
+    login_password: "{{ mysql_root_password }}"
+  loop: "{{ mysql_users }}"
+  when:
+    - mysql_users is defined
+    - mysql_server
+    - running_on_server
+  no_log: true
+
+- name: MySQL | create users
```
acozine, line 179: We have two scenarios - when we're installing a server, and when we're not. Should this task run in both scenarios? Either way, a better `name:` could clarify (`create the same users on the client` or `create client users` or something). If it should only run when we're not installing a server, then add a condition of `not mysql_server`.

roles/mysql/templates/my.cnf.j2
```diff
@@ -0,0 +1,5 @@
+{{ ansible_managed | comment }}
+
+[client]
+user='root'
+password='{{ mysql_root_password }}'
```
acozine, line 5: Ooh, I'm rusty on this . . . but does this need double-quote as well? `'"{{ mysql_root_password }}"'`? Or does it work as is?

roles/mysql/vars/main.yml
```diff
@@ -0,0 +1,34 @@
+---
+# vars file for mysql
+#
+#
+mysql_name: mariadb
```
acozine, line 5: Okay, so we are using mariadb.

acozine (commented):
This is a big PR and I'm guessing it will be a big improvement. But I haven't done a good technical review because I don't understand the goals of writing this new role yet. The PR doesn't describe why we're replacing the mariadb role or what innovations or simplifications the new mysql role offers. So far my questions are more about intentions than about implementation.

ansible.cfg
```diff
@@ -10,6 +10,7 @@ roles_path = external_roles
 inventory = hosts
 pipelining = true
 local_tmp  = ~/.ansible/tmp
+remote_tmp = /tmp/.ansible-${USER}/tmp
```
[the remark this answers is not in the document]
kayiwa: good catch this was my debugging why we suddenly couldn't connect to docker containers.

group_vars/gcp/vault.yml
```diff
@@ -0,0 +1,2 @@
+---
+vault_gcp_dev_bastion_host: pulsys@35.243.164.191
```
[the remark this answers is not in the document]
kayiwa: yeah it is for testing but will vault it regardless.

roles/mysql/molecule/default/verify.yml
```diff
@@ -0,0 +1,19 @@
...
+    package:
+      name: "{{ item }}"
+      state: present
+    check_mode: true
+    register: pkg_status
+    loop:
+      - mariadb-server
+      - mysql-common
+
+  - name: test for ruby packages
```
carolyncole, line 16: mariadb or mysql (not ruby)

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+# tasks file for mysql
+#
+- name: MySQL | configure apt repositories
+  ansible.builtin.apt_key:
+    keyserver: keyserver.ubuntu.com
+    id: "0xF1656F24C74CD1D8"
+
+- name: MySQL | add apt repositories
+  ansible.builtin.apt_repository:
+    repo: deb [arch=amd64] https://mirrors.gigenet.com/mariadb/repo/10.2/ubuntu {{ ansible_distribution_release }} main
```
carolyncole, line 11: This mirror keep changing.  I wonder if we should set this as a variable instead?

group_vars/pas/pas_gcp_dev.yml
```diff
@@ -0,0 +1,35 @@
...
+
+mysql_server: false
+
+mariadb__databases:
+    - name: "{{ pas_db_name }}"
+      encoding: utf8mb4
+      collation: utf8mb4_general_ci
+
+mysql_host: "10.0.10.6"
+mysql_root_password: "{{ vault_maria_mysql_root_password }}"
```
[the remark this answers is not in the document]
kayiwa: the name of the role is mysql... if we even move to Oracle's mysql, or percona we can still keep it. Underneath we have retained mariadb (just not in role name. - I take your point) Also -for now- I want the password to remain the same.

roles/mysql/templates/my.cnf.j2
kayiwa: works as is.

roles/mysql/tasks/main.yml
kayiwa: was used in debugging and forgot to remove.

roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,199 @@
...
+    encoding: "{{ item.encoding | default(omit) }}"
+    collation: "{{ item.collation | default(omit) }}"
+    login_unix_socket: "{{ mysql_socket }}"
+    login_host: "{{ mysql_host }}"
+    login_user: root
+    login_password: "{{ mysql_root_password }}"
+  delegate_to: "{{ mysql_host }}"
+  loop: "{{ mysql_databases }}"
+  when:
+    - mysql_databases is defined
```
carolyncole, line 156: Will this run twice if mysql_server is defined?  It seems like the command that as a mysql user my roles would utilize.

roles/mysql/tasks/main.yml
carolyncole: We have wanted that for testing some times, but I can not see a real reason.  I'm guessing this may have come out of the other roles.

roles/mysql/tasks/main.yml
kayiwa: yeah. This is how the db server will treat client connections to it.

group_vars/pas/pas_gcp_dev.yml
```diff
@@ -0,0 +1,35 @@
...
+mysql_databases:
+    - name: pas_prod
```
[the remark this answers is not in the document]
kayiwa: nice catch.

kayiwa:
We are simplifying the mariadb/mariadbserver roles. These roles we by design expecting a clustered mariadb-server using galera to make them highly available. We are making this simpler by installing mariadb (named mysql) as a standalone server. We are also combining the mariadb role and mariadbserver to install from one role.

carolyncole (commented):
If I am reading this PR correctly we have a new role, mysql, that either installs client packages OR creates a server based on the `mysql_server` variable.  This then also creates a new PAS machine that runs with the new mysql server.  If we do not want to provision a new machine we can just utilize pas staging.

roles/mysql/molecule/default/verify.yml
kayiwa: Heya @carolyncole I needed to make sure it does the right thing. So the pas is simply a proof of concept that it does. It is also installed on google cloud. So the role will set up a mariadb server (rarely) but most of the times it will be setting up connections to the mariadb server. The flag `mysql_server` is used when setting up a server.

roles/mysql/tasks/main.yml
kayiwa: the `mysql_server` will be defined as false except in the `group_vars/mysql` (this is to set up a server) All other connections will be defined in the projects with a `mysql_server: false` so as not to install mariadb

host_vars/gcp_mariadb_dev1.yml
```diff
@@ -0,0 +1,2 @@
+---
+gcp_mariadb_dev: 10.0.10.6
```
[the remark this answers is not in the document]
kayiwa: In order to make sure this PR doesn't break things. I am testing this on Google Cloud to lower the surprise factor. I am able to create a db and a db user. It could have been anywhere but I want to be sure that my new PR's will not break our current implementation.

playbooks/pas_gcp_dev.yml
```diff
@@ -0,0 +1,20 @@
+---
+- hosts: pas_gcp_dev
```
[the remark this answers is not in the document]
kayiwa: In order to add name here we will need a routable IP address. We can't get a routable IP address from our Google Cloud VMs. They all have [RFC 1918](https://datatracker.ietf.org/doc/html/rfc1918) so if I enter a name it will fail. The connection goes through a bastion host.

roles/mysql/molecule/default/destroy.yml
kayiwa: this comes from the `molecule init` command. I've created an issue https://github.com/pulibrary/princeton_ansible/issues/2715

playbooks/mariadb_gcp_dev.yml
```diff
@@ -0,0 +1,25 @@
+---
+- hosts: gcp_mariadb_dev
```
acozine, line 2: I'd still like to see `name:` lines on all plays. The `name:` line has no requirements, it's just a title for the play so when you read the playbook or watch it execute you can easily understand the intent. For example, this one might be:
`- name: build a test db on gcloud`. Once you add `- name:`, the `hosts:` entry becomes the second line of the play and loses the `-`. 

Adding `- name:` to a play is similar to adding `- name:` to a task, but instead of describing what the task does, it describes what the entire play does.

playbooks/pas_gcp_dev.yml
acozine, line 2: Here too - a good name for this one might be `- name: build a test instance of the Princeton and Slavery site on gcloud`.

roles/mysql/README.md
```diff
@@ -0,0 +1,60 @@
+MariaDB (MySQL) role
+=========
+
+This role installs and manages connections to mariadb. The role supports different modes of operation depending on the presence of MariaDB installed locally or not
```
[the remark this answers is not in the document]
acozine: Thanks for the explanation, @kayiwa! For those following along at home, see #2737 for the full story.

roles/mysql/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
...
+mysql_bind_address: "127.0.0.1"
+
+
+# The buffer pool size.
+mysql_innodb_buffer_pool_size: 1G
+
+# The io capacity.
+mysql_innodb_io_capacity: 4000
+
+# the hose
```
acozine, line 14: typo
```suggestion
# the host
```

roles/mysql/handlers/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+# handlers file for mysql
+#
+- name: systemctl daemon-reexec
```
acozine, line 4: How about this as a name?
```suggestion
- name: restart systemd without rebooting the entire system
```

roles/mysql/handlers/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    daemon_reexec: true
+  when:
+    - running_on_server
+
+- name: restart mysql server
+  ansible.builtin.service:
+    name: mysql
+    state: restarted
+  when:
+    - not ansible_check_mode | bool
```
acozine, line 15: I don't think we need to state the data type here. `check_mode` is a boolean magic var, set to no/false by default. It gets set to yes/true when you run a play with `- check`. See https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#skipping-tasks-or-ignoring-errors-in-check-mode.

roles/mysql/vars/main.yml
```diff
@@ -0,0 +1,36 @@
...
```
acozine, line 5: Values set in this file will override anything in `group_vars` - is that what we want? I suspect these should go in `defaults/main.yml`.

acozine (commented):
Thanks for the explanation about where this PR is taking us and what the steps are. Now I've been able to give this a more thorough technical review.

acozine (approved):
Looks good, thanks!

kayiwa:
> 
It does the following. Allows us to 

- install and configure mariadb server
- install and configure a mariadb client connection

we no longer need a mysql and mysql-server setup. Use of `when: [ mysql_server: true]` determines when the server will be installed.

carolyncole (approved):

PR #2710, opened 2022-02-01
explicitly adds 'production' to all host groups for production machines

### LINKED ISSUE #181: refactor host names
In several cases there are hosts with name patterns like `X` and `X_production`. It's confusing because if you see `X` first you might (as I did) assume it's the production configuration. I think we should rename them to be `X_staging` and `X_production`.

hosts with both entries, in which only one has the appended string:

- [x] bibdata
- [ ] pulmap
- [x] archivesspace
- [x] orangelight
- [ ] zookeeper
- [ ] solrcloud

Probably should document this policy either in an ADR or in the README, as well

acozine:
Closes #181.

Group names for all projects with both staging and production groups now reflect the environment name as well as the project name. No more `project` and `project_staging`.  However, some groups use `stage` instead of `staging` and `prod` instead of `production`. Should I update those group names as well? Or are we okay with a little variation as long as the group clearly indicates the environment it's targeting?

### DIFF
--- a/hosts
+++ b/hosts
@@ -56,7 +56,7 @@ cicognara1.princeton.edu
 cicognara2.princeton.edu
 [bibdata_workers]
 bibdata-worker2.princeton.edu
-[pulmap]
+[pulmap_production]
 maps1.princeton.edu
 maps2.princeton.edu
 [pulmap_staging]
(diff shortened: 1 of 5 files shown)


acozine:
Oh, I also re-ordered the groups for the `solr8*` entries, putting `production` ahead of `staging` (which is alphabetical, and also was the way the previous groups were ordered. Then I realized that a lot of groups were listed in the other order, and I stopped making this change. Again, I'm not sure if we want complete consistency, or just clarity and iterative improvement.

hackartisan (approved):
+1 to clarity and iteration

escowles (approved):

kayiwa (approved):
OK

Issue #2088, opened 2021-01-21
Add datadog role to production playbook for fpul


carolyncole:

acozine:
Looks like datadog is running on fpul: https://github.com/pulibrary/princeton_ansible/blob/0870016bde6c5f1db4e5004006aa14835442e187/group_vars/friends_of_pul/friends_of_pul_prod.yml#L28

acozine:
Oops, re-opening. Looks like we added the variables, but the Datadog role is not getting run.

acozine:
Closed by #2711.

PR #2711, opened 2022-02-01
add datadog role

kayiwa:


kayiwa:
add datadog role to fpul. This closes #2088

acozine (approved):
Looks correct. Now those variables will start doing us some good!

PR #2718, opened 2022-02-03
Removing solr references in libwww role

carolyncole:
This is setup via capistrano, not ansible.  These are just confusing and out of date.


kayiwa (approved):
OK!

PR #2726, opened 2022-02-04
Deduplicate playbooks

acozine:
We currently have multiple playbooks for each project, one per environment. In most pairs, the only differences are the `hosts` and the `vars_files`. To streamline the repository, simplify maintenance, and reduce the risk of drift, we'd like to reduce this duplication where possible.

This branch creates two "unified" playbooks as proofs of concept:
- lae.yml
- pdc_discovery.yml

Each playbook can run be against production or staging. Both run against staging by default. The `hosts` and the `vars_files` are defined based on the variable `runtime_env` (though may be `ansible_env` would be better?). The existing playbooks are still there for comparison purposes.

This PR should stay a draft until these two playbooks have been tested. If the unified playbooks work, next steps would include:
- remove related `_production.yml` and `_staging.yml` playbooks
- do the same for other project playbook pairs
- audit `group_vars.yml` for duplication and drift

### DIFF
--- a/playbooks/lae.yml
+++ b/playbooks/lae.yml
@@ -1,0 +1,22 @@
+---
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- name: build the Latin American Ephemera site
+  hosts: lae_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+
+  vars_files:
+    - ../group_vars/lae/{{ runtime_env | default('staging') }}.yml
+
+  roles:
+    - role: roles/lae
+
+  post_tasks:
+    - name: tell everyone on slack you ran an ansible playbook
+      slack:
+        token: "{{ vault_pul_slack_token }}"
+        msg: "{{ inventory_hostname }} completed"
+        channel: #server-alerts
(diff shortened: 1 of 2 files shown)


hackartisan:
What about `target_env`? or `host_env`?

playbooks/lae.yml
```diff
@@ -0,0 +1,22 @@
+---
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- name: build the Latin American Ephemera site
+  hosts: lae_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+
+  vars_files:
+    - ../group_vars/lae/{{ runtime_env | default('staging') }}.yml
```
carolyncole, line 10: I like that the default is staging, and that you include in the top how to run production

carolyncole (commented):
I like the way this is heading.  I could get on board with runtime_env, or host_env.  We use just `environment` in the capistrano world, so that could be an option.

jrgriffiniii (approved):

jrgriffiniii:
@acozine and I have just tested the Playbook for `pdc_discovery`, and this was quite successful!

PR #2738, opened 2022-02-09
Removing alma from orangelight names

carolyncole:
There is no longer another version and keeping the alma makes it harder to remember, becuase they do not follow the patterns established


kayiwa (approved):
OK

PR #2746, opened 2022-02-10
add ansible-lint comments

### LINKED ISSUE #2715: explain the ansible lint exceptions in example
We use ansible lint exception without explanations in the example role

kayiwa:
instead of copy and pasting blindly provide reasons in comments for why
we skip these linting tests

closes #2715


hackartisan (approved):
Thank you for this!

PR #2748, opened 2022-02-10
[DPUL] Fix uploads directory ownership and use new dpul-staging2

hackartisan:
Required updates to the dpul role to fix directory ownership of symlink to mounted shared/uploads.

Remove dpul-staging1 because we're going to rebuild it. We'll add it back
when it's ready

refs pulibrary/dpul#1296


kayiwa (approved):
OK

PR #2754, opened 2022-02-10
Adds guidance to nginxplus role README and playbooks

### LINKED ISSUE #2751: Add to readme about nginxplus options
It may be best practice to use `--limit` to only run on one adc at a time.

Other options are the new tags `update_conf` and the other one for certs.

acozine:
Closes #2751. Related to #2732.

The nginxplus role has two tags that shorten execution time for common tasks. Document those tags in the playbook and in the role README. 

Add comments to the playbook and the role README about running load-balancer updates with `-limit` in case of unexpected turbulence.


kayiwa (approved):
OK

PR #2764, opened 2022-02-16
i2720 migrate mariadb staging

kayiwa:
-  this PR adds the mysql-db-staging
- add a python check


roles/mysql/tasks/main.yml
```diff
@@ -1,17 +1,22 @@
...
+    url: https://mariadb.org/mariadb_release_signing_key.asc
+    state: present
+  register: install_app_key
 
-- name: MySQL | add maria db repository
+- name: MySQL | Install MariaDB repository
   ansible.builtin.apt_repository:
     repo: "deb [arch=amd64] https://mirrors.accretive-networks.net/mariadb/repo/10.2/ubuntu {{ ansible_distribution_release }} main"
+    state: present
+    validate_certs: false
```
acozine, line 14: Can we get cert validation working? Also, should this be a handler?

roles/mysql/tasks/main.yml
```diff
@@ -30,8 +35,28 @@
...
   when:
     - mysql_server
-  notify:
-    - set root password
+
+- name: MySQL | Restart MySQL
+  ansible.builtin.service:
+    name: mysql
+    state: restarted
+  when: install_app_key.changed
```
acozine, line 43: This one seems like it should be a handler. We might want to restart mysql for other reasons, besides "the key changed".

roles/mysql/tasks/main.yml
```diff
@@ -30,8 +35,28 @@
...
+
+- name: MySQL | mysql_root_password
+  community.mysql.mysql_user:
+    login_unix_socket: "{{ mysql_socket }}"
+    login_user: root
+    login_host: localhost
+    login_password: ""
```
acozine, line 50: What is the difference between the `login_password` and the `mysql_root_password` below? Null passwords make me uneasy.

roles/mysql/tasks/main.yml
```diff
@@ -30,8 +35,28 @@
...
+    name: 'root'
+    password: "{{ mysql_root_password }}"
+  when: install_app_key.changed
+
+- name: MySQL | remove unix_plugin
```
acozine, line 55: This isn't a comment on the playbook, so much as a question about our procedure here. What is the unix plugin and why are we installing and then removing it?

playbooks/mysql_staging.yml
```diff
@@ -0,0 +1,33 @@
...
+        upgrade: dist
+        force_apt_get: true
+
+    - name: install pip on new VM
+      raw: test -e /usr/bin/python
+      changed_when: false
+      failed_when: false
+      register: check_python
+
+    - name: Install Python  # this has to be raw in the event python is not installed
```
acozine, line 21: Again, if no python is installed on a box at all, I would expect the first pre-task to fail. If we need to handle the case where a new VM has no python installed at all, we probably need to do that first. If we're just looking to get a more recent python, or python in the right place, I think we can use `apt`.

acozine (commented):
Progress! I had a couple of questions, but overall this looks good.

roles/mysql/tasks/main.yml
kayiwa: it can't be a handler. We want to get this certificate. Not all repos do TLS. So if we switch to new mirror we don't want to get the... needs TLS cert.

roles/mysql/tasks/main.yml
kayiwa: when you install mysql on Ubuntu it has no password for root. So we are saying log into mysql with your blank password then set the password to `mysql_root_password`

playbooks/mysql_staging.yml
```diff
@@ -0,0 +1,33 @@
...
+  pre_tasks:
+    - name: Upgrade all apt packages
+      apt:
+        upgrade: dist
+        force_apt_get: true
+
+    - name: install pip on new VM
+      raw: test -e /usr/bin/python
+      changed_when: false
+      failed_when: false
```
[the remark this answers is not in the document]
kayiwa: the order needs to be rearranged for sure but we have python absent in VMs. Ubuntu focal for sure expects python3.

roles/mysql/tasks/main.yml
kayiwa: Mysql password's changed at 5.7 Otherwise it fails. https://mariadb.com/kb/en/authentication-plugin-unix-socket/

roles/mysql/tasks/main.yml
kayiwa: MySQL here fails to work. It needs this to restart so that is isn't listening on `localhost` and listens on all interfaces.

PR #2769, opened 2022-02-16
Updating Node to Fermium LTS for the lae Role

jrgriffiniii:
This is necessary in order to resolve https://github.com/pulibrary/lae-blacklight/issues/293


jrgriffiniii:
Please note that this has been successfully tested against both the `staging` and `production` environments.

PR #2775, opened 2022-02-18
Automate access to the new studio machine

acozine:
Roel reported seeing `permission denied (publickey)` when trying to do the rake ingest of LAE material on the new studio processing machine. This is a baby step toward automating management of the studio processing machines - I think it will solve the immediate problem and I hope it will give us a foundation to build on, so we can automate whatever else needs to be installed/managed on those machines.


kayiwa (approved):
OK

acozine:
Related to #2670.

tpendragon (approved):

PR #2776, opened 2022-02-21
Proxy pass template

acozine:
The template we've been using for proxy pass configuration is more of a file than a template.

This PR pulls the data for proxy passes into its own (alphabetized) vars file, makes the template a template, and adds a task to fill out the template. Separating the data out doesn't reduce maintenance, but it does make the entries more readable, so folks can easily add or edit entries for sites that need proxy passes.


acozine:
This PR now attempts to:
- consolidate our proxy directives in the nginxplus role
- delete unused proxy directives from the libwww role
- record our proxy needs in a vars file
- use a template to generate nginx config with proxy directives
Still a bunch of work to be done, but this will be much tidier!

acozine:
Not sure of the correct destination for the `/njmaps` address in production - probably it should point to `lib-static-prod` like [this](https://github.com/pulibrary/princeton_ansible/blob/217b0c5ef0a1032a4e9c160283a55e52a10f12fb/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-prod.conf#L52) and not to `libphp-prod` like [this](https://github.com/pulibrary/princeton_ansible/blob/cbe9cafcc1e0fad3798908026a440227d21c8d1d/roles/libwww/templates/proxy_pass_rules.j2#L8), but confirm.

roles/nginxplus/vars/proxies_by_box.yml
```diff
@@ -0,0 +1,141 @@
+# list of proxy directives
```
carolyncole, line 1: @acozine the only concept I am not seeing is that we have different proxies for the staging library site and the production library site.  For example: https://library.princeton.edu/njmaps/ and https://library-staging.princeton.edu/njmaps/. go to two different places  https://lib-static-prod.princeton.edu/njmaps/ for prod and https://firestone-renovations-staging.princeton.edu/njmaps/ for staging.  This is pattern is used a lot to proxy the staging version of the page to library-staging.princeton.edu and the production version to library.princeton.edu.  This is necessary to test how the page will behave behind the proxy before putting changes into production.

carolyncole: I know this is a WIP, so I am just making sure you have that in mind...

acozine: Ah, thanks @carolyncole. I will make two files here - one for prod and one for staging.

acozine:
I can't tell why this got closed. I'd guess an automated process that closes Draft PRs after 2 months, except we have older Draft PRs that are still open. When someone actively closes a PR, the message says `Person closed this N days ago` but this one just says `This pull request was closed`. This is still something I'd like to work on, so I'm reopening the PR. Chime in if there are reasons to close it and pursue this work on a different path.

acozine:
Closing in favor of #3741.

PR #2777, opened 2022-02-21
add jtsviewer proxy

kayiwa:


roles/nginxplus/files/conf/http/cdh_prod_geniza.conf
```diff
@@ -59,6 +59,9 @@ server {
     location /indexcards/ {
         proxy_pass https://genizaprojects.princeton.edu/indexcards;
     }
+    location /jtsviewer {
+        proxy_pass https://genizaprojects.princeton.edu/jtsviewer/;
+    }
```
rlskoeser, line 64: should the slashes here match the ones for `indexcards`?  (trailing on location, no trailing on redirect?

rlskoeser (approved):

PR #2778, opened 2022-02-22
i2763 add pgrtm extension

kayiwa:
- add google cloud bastion variables
- we will add ansible pg_hba module
- we need to use pgpass to connect from standby
- one connection will be for server connection the second task is reserved for leader/standby connections
- management of pg_hba will have two connections
- ensure that our database owner is owned by db user
- add demo for postgres connection
- lower the permissions of the db users
- use reload instead of restart
- variables for ojs demo
- add delegate_to allows the role not to skip steps
- allow individual tasks to run on different hosts
- we were not giving superuser access
- make sure the postgres user can run services
- correct standy and clean variables
- correct standy and clean variables
- cherry pick to unbreak CI
- remove omit references where possible
- add google cloud labels for vms
- cherry pick from main to unbreak CI
- add new postgres VMs
- add host variables for two databases
- we make the owner of the file postgres
- server user creation doesn't need host
- add the pg_trgm extension


acozine:
We've done this work in a different way.

PR #2781, opened 2022-02-22
[Bibdata] Use 'bibdata', not 'marc_liberation' in app configs

### LINKED ISSUE #2765: [Bibdata] app_configs should be named bibdata
They are still named marc_liberation.  They should be named bibdata

https://github.com/pulibrary/princeton_ansible/search?q=marc_liberation

acozine:
Closes #2765.

Replaces `marc_liberation` with `bibdata` in the bibdata vars and role.

@carolyncole take a look and let me know if all of these really should be changed - I wasn't sure about the `/data/marc_liberation` directories. I made several small commits here in case we need to revert some of the changes.


carolyncole:
@acozine I believe we will have to make changes on the application side to accommodate these changes.  @christinach am I correct with that assumption.

We may also need to change something on the server with the data directory.  I too am a little unclear with the internal samba share going on between all the bibdata machines.

christinach:
@carolyncole @acozine yes we need to make changes on the application side before we merge this PR.

carolyncole (dismissed):
We need to coordinate this with code changes to remove marc_liberation in the codebase https://github.com/pulibrary/bibdata/search?p=2&q=marc_liberation. Not all of these are needed

roles/bibdata/defaults/main.yml
```diff
@@ -2,4 +2,4 @@
 running_on_server: false
 bibdata_samba_source_host: 'bibdata-worker1.princeton.edu'
 bibdata_mount_env: 'staging'
-bibdata_data_dir: '/data/marc_liberation_files'
```
sandbergja, line 5: I think we'll want to keep the files in this directory on-hand and accessible.  Could we add something like this to the roles/bibdata/tasks/mounts.yml?

```
  - name: bibdata | Check if there are old files in /data/marc_liberation_files
    ansible.builtin.stat:
       path: /data/marc_liberation_files
       register: marc_liberation_files_dir
  - name: bibdata | Move over old marc_liberation_files
    ansible.builtin.copy:
      src: '/data/marc_liberation_files/*'
      dest: '{{ bibdata_data_dir }}'
      when: marc_liberation_files_dir.exists
  - name: bibdata | Remove old marc_liberation_files directory
    ansible.builtin.file:
       path: /data/marc_liberation_files
       state: absent
```

sandbergja (changes requested):

roles/bibdata/defaults/main.yml
christinach: @sandbergja thanks for catching this. Yes we definitely want to keep the files in this directory. I think It will take some time to move them. @kayiwa @acozine would it be better to create the new directory `bibdata_data_dir: '/data/bibdata_files` and copy the files there before we run this playbook ??

christinach (changes requested):
see comment https://github.com/pulibrary/princeton_ansible/pull/2781/files#r904947438

roles/bibdata/defaults/main.yml
acozine: I wonder if we could solve this by adding a line to the playbook that renames the directories before changing the config . . . that way we would not have to move the files. Alternatively, we could just keep the old directory names, but that feels like adding a layer of cruft.

acozine:
Notes from Ansible meeting 6/28:
- for implementing this, try decoupling the directory changes from the code/config changes, to reduce the number of moving parts
- turn off the workers while we update
- take one box at a time off the load-balancer to update
- see https://stackoverflow.com/a/66283091 for a potential approach to the directory renaming
Thanks @sandbergja @hackartisan @maxkadel @stephayers !

sandbergja:
I just looked a bit more closely at the servers -- we might not need to move/rename any directories at all!  It looks like /data/marc_liberation_files is just a symlink.  So I think we just need to change that. :-)

roles/bibdata/defaults/main.yml
acozine: Since this is just a symlink, we think it will be okay. Testing on the staging servers first!

christinach:
@acozine please rebase the pr to get the new scsb auth key.

acozine:
@christinach I will get this rebased this afternoon, so it will be ready for Monday.

acozine:
I edited the vault file, added the missing key entry on the branch, then rebased the branch and selected the changes in the branch to solve the merge conflict. I think this produced the right set of changes in the end, but I'd welcome a coherence check.

sandbergja (approved):
Thank you @acozine !!

christinach (approved):

christinach:
We have tested the pr in bibdata-staging and bibdata-qa. We will run it in bibdata production on 08/16/2022.

PR #2783, opened 2022-02-23
Moving to mariadb-prod1 for the database

carolyncole:
This is in production, so it should be merged...


acozine:
Looks good, thanks @carolyncole

Issue #2735, opened 2022-02-08
Postgresql migration epic


acozine:
Component parts:

- [ ] all projects should run on new postgresql servers (migrate from `lib-postgres3` and `lib-postgres-prod1` to the new replicated servers, `lib-postgres-prod3` and its replicated box `lib-postgres-prod2`)
- [ ] all projects should upgrade from psql 10 to psql 13 (does this involve code changes?)

Check off when the project has been migrated to the new servers and to psql 13 if necessary. Projects that run on postgres:

- [x] ansible-tower
- [x] approvals [lib-postgres3, pgsql 10]
- [x] approvals-staging [lib-postgres3, pgsql 10]
- [x] bibdata [lib-postgres-prod, pgsql 13]
- [x] cicognara [lib-postgres3, pgsql 10]
- [x] cicognara-staging [lib-postgres3, pgsql 10]
- [x] dpul [lib-postgres3, pgsql 10]
- [x] dpul-staging [lib-postgres3, pgsql 10]
- [x] dss [lib-postgres3, pgsql 10]
- [x] dss-staging [lib-postgres3, pgsql 10]
- [x] figgy [lib-postgres3, pgsql 10]
- [x] figgy-staging [lib-postgres3, pgsql 10]
- [x] geaccirc [lib-postgres3, pgsql 10]
- [x] gscatalog/geoserver (investigate)
- [x] lae [lib-postgres3, pgsql 10]
- [x] lae-staging [lib-postgres3, pgsql 10]
- [x] libjobs [lib-postgres3, pgsql 10]
- [x] lib-jobs-staging [lib-postgres-staging1, pgsql 13]
- [x] libstatistics [lib-postgres3, pgsql 10]
- [x] libstatistics-staging [lib-postgres-staging1, pgsql 13]
- [x] lockers and study spaces [lib-postgres3, pgsql 10]
- [x] lockers and study spaces staging [lib-postgres3, pgsql 10]
- [x] mudd [lib-postgres3, pgsql 10]
- [x] mudd-staging [lib-postgres3, pgsql 10]
- [x] ojs [lib-postgres-prod1, pgsql 13]
- [x] ojs-staging [lib-postgres-staging1, pgsql 13]
- [x] openmonograph
- [x] openmonograph-staging
- [x] ouranos-staging [lib-postgres3, pgsql 10]
- [x] pulfalight [lib-postgres3, pgsql 10]
[...]

PR #2785, opened 2022-02-24
i2735 postgres staging variables

kayiwa:
- remove superuser for database users
- add roles impacted by new postgresql role
- add postgresql role
- commit the local staging environment
- add gcp google environment
- add a group vars for cicognara
- add a group vars for dpul
- add pg hba variables for staging


acozine (changes requested):
This is an outdated artifact of the postgres migration - we decided to split the changes up into many smaller PRs. We can keep this open until the migration is finished (see #2735 for the current status); a blocking review makes it obvious that we do not plan to merge this.

kayiwa:
closed by individual projects PRs.

PR #2796, opened 2022-02-28
Upgrade nodejs for orangelight, enable controlled upgrades generally

acozine:
For easy maintenance and controlled upgrades, we want the `nodejs` role to install a specific version.

Jane's change updates the version of `nodejs` for orangelight.

Alicia's change requires that version to be installed.

### DIFF
--- a/roles/nodejs/defaults/main.yml
+++ b/roles/nodejs/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
-nodejs__upstream_release: 'node_8.x'
+nodejs__upstream_release: 'node_14.x'
 nodejs__upstream_key_id: '9FD3B784BC1C6FC31A8A0A1C1655A0AB68576280'
 
 # URL of the NodeSource APT repository in ``sources.list`` format.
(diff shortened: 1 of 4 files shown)


acozine:
We can see the base version number now, but the `apt` task still fails, saying no version matches. The playbook should probably:
- check the installed version
- if the installed version matches the desired version, stop there
- if the installed version does not match the desired version, download node from `nodejs.org/download/release/latest-v{{ desired_version }}.x/node-v{{ desired_version }}."whatever comes next"-linux-x64.tar.gz`, then unzip and install node from that . . . but I'm not sure how to do the matching in a sustainable way (meaning, a way that works now for 16.14.0 and will still work when 16.20.8 comes out); alternatively, we could update the node version to be more specific (for example, 16.14.0 instead of 16.x)

acozine:
Turns out they DO keep older releases on each line, see https://nodejs.org/download/release/v16.14.0/

acozine:
Superseded by #2809.

PR #2802, opened 2022-03-01
[Figgy] Update to Ruby 2.7

tpendragon:
Draft until tested in staging.


tpendragon:
I'm merging this without approval since this is deployed now and I don't want it to fall out of sync.

PR #2809, opened 2022-03-03
build nodejs

kayiwa:
This PR eliminates the need for `apt` versions of node. Instead we download specific binaries from the nodejs project to install node. 
- add nodejs variables
- add specific node version info

It will supercede #2796 and #2787

### DIFF
--- a/roles/appdeploy/defaults/main.yml
+++ b/roles/appdeploy/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 # defaults file for roles/appdeploy
+desired_version: v16.14.0
--- a/roles/appdeploy/defaults/main.yml
+++ b/roles/appdeploy/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # defaults file for roles/appdeploy
-desired_version: v16.14.0
+desired_nodejs_version: v16.14.0
--- a/roles/appdeploy/tasks/main.yml
+++ b/roles/appdeploy/tasks/main.yml
@@ -1,9 +1,28 @@
 ---
-- name: Install "forever" node.js package globally.
+- name: appdeploy | Install "forever" node.js package globally.
   community.general.npm:
     name: forever
-    global: yes
-- name: Install "coffee-script" node.js package globally.
+    global: true
+
+- name: appdeploy | register forever binary path
+  ansible.builtin.file:
+    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/forever/bin/forever"
+    dest: "/usr/local/bin/forever"
+    owner: root
+    group: root
+    state: link
+  changed_when: false
+
+- name: appdeploy | Install "coffee-script" node.js package globally.
   community.general.npm:
     name: coffee-script
-    global: yes
+    global: true
+
+- name: appdeploy | register coffee-script binary path
+  ansible.builtin.file:
+    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/coffee-script/bin/coffee"
+    dest: "/usr/local/bin/coffee"
+    owner: root
+    group: root
+    state: link
+  changed_when: false
--- a/roles/appdeploy/tasks/main.yml
+++ b/roles/appdeploy/tasks/main.yml
@@ -6,7 +6,7 @@
 
 - name: appdeploy | register forever binary path
   ansible.builtin.file:
-    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/forever/bin/forever"
+    src: "/usr/local/node-{{ desired_nodejs_version }}-linux-x64/lib/node_modules/forever/bin/forever"
     dest: "/usr/local/bin/forever"
     owner: root
     group: root
@@ -20,7 +20,7 @@
 
 - name: appdeploy | register coffee-script binary path
   ansible.builtin.file:
-    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/coffee-script/bin/coffee"
+    src: "/usr/local/node-{{ desired_nodejs_version }}-linux-x64/lib/node_modules/coffee-script/bin/coffee"
     dest: "/usr/local/bin/coffee"
     owner: root
     group: root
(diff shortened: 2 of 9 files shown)


roles/nodejs/tasks/main.yml
```diff
@@ -1,45 +1,12 @@
...
-  when: mit_servers is failed
-  ignore_errors: true
-
-- name: Add upstream Yarn APT key from PGP
+- name: nodejs | Add upstream Yarn APT key yarn
   apt_key:
     url: https://dl.yarnpkg.com/debian/pubkey.gpg
     state: present
   register: yarn_servers
   ignore_errors: true
```
acozine, line 7: I wonder if we could make a single task for "find a Yarn APT key that works," pass it a list of all these possible sources, then define "failure" for that task to be "none of these worked"? Not sure exactly what that looks like, but I'm pretty sure it's possible. See https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#defining-failure for info on defining failure.

roles/nodejs/tasks/main.yml
```diff
@@ -57,54 +24,53 @@
   register: yarnmit_servers
   ignore_errors: true
 
-- name: Add upstream Yarn APT key from PGP
+- name: nodejs | Add upstream Yarn APT key from PGP
   apt_key:
     id: "{{ nodejs__yarn_upstream_key_id }}"
     state: present
     keyserver: "keyserver.pgp.com"
   when: yarnmit_servers is failed
```
acozine, line 32: Hm. This will run when the MIT key is missing but we already have a key from sks. Another reason to put this in a loop if we can.

roles/nodejs/tasks/main.yml
```diff
@@ -57,54 +24,53 @@
...
-  debug:
-    msg: "{{ nodejs_wanted_version }}"
+- name: nodejs | create symbolic link for node
+  ansible.builtin.file:
+    src: "/usr/local/node-{{ desired_version }}-linux-x64/bin/node"
+    dest: "/usr/local/bin/node"
+    owner: root
+    group: root
+    state: link
+  changed_when: false
```
acozine, line 62: I assume `changed_when: false` is here to make idempotence pass . . .

roles/nodejs/tasks/main.yml
```diff
@@ -57,54 +24,53 @@
...
 
-- name: Check for the LTS release
-  set_fact:
-    nodejs_lts_installed: "{{ nodejs_versions | select('match', nodejs_wanted_version) | list | length > 0 }}"
-  when: nodejs_versions is defined
+- name: nodejs | create symbolic link for npm
```
acozine, line 64: This task and the one above it could be a single task with a loop, I think.

roles/nodejs/tasks/main.yml
kayiwa: yep

roles/nodejs/tasks/main.yml
kayiwa: thanks!

roles/nodejs/tasks/main.yml
kayiwa: thanks!

roles/nodejs/tasks/main.yml
```diff
@@ -1,45 +1,12 @@
...
   apt_key:
     url: https://dl.yarnpkg.com/debian/pubkey.gpg
     state: present
   register: yarn_servers
   ignore_errors: true
 
-- name: Add upstream Yarn APT key from PGP
+- name: nodejs | Add upstream Yarn APT key from sks-keyservsers
   apt_key:
     id: "{{ nodejs__yarn_upstream_key_id }}"
```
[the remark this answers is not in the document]
kayiwa: this is the GPG defined key by the yarn folks upstream. It is defined in defaults.

roles/nodejs/tasks/main.yml
kayiwa: thanks!

roles/appdeploy/tasks/main.yml
```diff
@@ -1,9 +1,28 @@
...
+- name: appdeploy | Install "forever" node.js package globally.
   community.general.npm:
     name: forever
-    global: yes
-- name: Install "coffee-script" node.js package globally.
+    global: true
+
+- name: appdeploy | register forever binary path
+  ansible.builtin.file:
+    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/forever/bin/forever"
```
jrgriffiniii, line 9: 👍

roles/appdeploy/tasks/main.yml
```diff
@@ -1,9 +1,28 @@
...
+
+- name: appdeploy | Install "coffee-script" node.js package globally.
   community.general.npm:
     name: coffee-script
-    global: yes
+    global: true
+
+- name: appdeploy | register coffee-script binary path
+  ansible.builtin.file:
+    src: "/usr/local/node-{{ desired_version }}-linux-x64/lib/node_modules/coffee-script/bin/coffee"
```
jrgriffiniii, line 23: 👍

jrgriffiniii (approved):

sandbergja (approved):
This works great!  Thank you, @kayiwa!

roles/appdeploy/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 # defaults file for roles/appdeploy
+desired_version: v16.14.0
```
acozine, line 3: I'd make this `desired_node_version` throughout, because `desired_version` could be anything. Also, add a note that if the project requires a certain version, the value for this should match it. For example, https://github.com/pulibrary/orangelight/blob/0648c94b853a4d1ed61fad48128fa54eaf68317c/.tool-versions#L3.

roles/nodejs/tasks/main.yml
```diff
@@ -1,110 +1,56 @@
...
+    keyserver: "{{ item }}"
+    state: present
+  when: yarn_pkg_registered is failed
+  register: yarn_key_registered
+  loop:
+    - "hkp://pool.sks-keyservers.net"
+    - "pgp.mit.edu"
+    - "keyserver.pgp.com"
+  failed_when:
+    - yarn_key_registered == 0
```
acozine, line 20: This looks great!

acozine (commented):
Nice work!

christinach:
@kayiwa if an application is running a different desired_nodejs_version where will I overwrite this?

PR #2812, opened 2022-03-03
switch pulfalight vars to use load-balanced Solr8

### LINKED ISSUE #2810: configure pulfalight app to use solr through loadbalancer
it currently points directly to lib-solr-prod4

acozine:
Closes #2810.

We discovered today that pulfalight was pointing to an individual Solr box instead of using the load balancer to access Solr. This updates the config to use the load balancer.


kayiwa (approved):
OK

PR #2814, opened 2022-03-04
[Bibdata] Adding a QA playbook

### LINKED ISSUE #2753: Build bibdata-qa playbook
https://github.com/pulibrary/princeton_ansible/issues/2703 to build machines utilizing the VM

carolyncole:
fixes #2753


roles/bibdata/tasks/mounts.yml
```diff
@@ -38,6 +38,23 @@
...
+    group: '{{ deploy_user }}'
+
+- name: Make sure the mount directory exists
+  file:
+    path: '/mnt/bibdata_share/{{ bibdata_mount_env }}'
+    state: directory
+    recurse: no
+    owner: '{{ deploy_user }}'
+    group: '{{ deploy_user }}'
+  run_once: true
```
acozine, line 56: With this line added, if you run the role on multiple machines (for example, if we spin up 3 VMs to run bibdata), this task will only run on the first machine and not on the others

Is that the intended outcome? Why wouldn't we want to ensure that a mount directory exists on each machine?

carolyncole:
@acozine Thank you for the review.  I updated the postgres version.  That was a copy paste error...

PR #2816, opened 2022-03-04
Deduplicate playbooks that start with 'a'

acozine:
Related to #2726.

Deduplicates playbooks for three projects:
- abid
- annotations
- approvals

All playbooks run against staging infrastructure by default, and can run against production infrastructure if you pass `-e runtime_env=production` or (in the case of the approvals site) `-e runtime_env=prod`. 

Adds a `name:` line to all plays, and makes some whitespace adjustments.


kayiwa (approved):
OK

PR #2818, opened 2022-03-04
Deduplicate playbooks at the end of the alphabet

acozine:
Related to #2726.

Part of a project to replace all `*_production.yml` and `*_staging.yml` playbook pairs with unified playbooks that run against staging by default and against production when you pass `-e runtime_env=production` at the command line. This change was complicated because the `solrcloud` and `zookeeper` group_vars files, used in the zookeeper playbook, didn't follow our standard pattern.

This PR:

- Deduplicates two playbooks:
  - vireo
  - zookeeper
- moves the `solrcloud` and `zookeeper` group_vars files into subdirectories to match our pattern
- adds zookeeper production group_vars
- updates other playbooks that reference the `solrcloud` group_vars to point to the new location
- adds `name:` lines for all plays


kayiwa (approved):
OK

Issue #2828, opened 2022-03-09
ensure node version is in group vars


kayiwa:
With the new nodejs role in main lets ensure that all roles using nodejs do not
break

PR #2829, opened 2022-03-09
update nodejs variables

kayiwa:
the new nodejs role downloads the binary version of the file. We add the
values matched on the vms

### DIFF
--- a/group_vars/abid/production.yml
+++ b/group_vars/abid/production.yml
@@ -12,8 +12,7 @@ passenger_app_env: "production"
 passenger_ruby: "/usr/bin/ruby2.6"
 ruby_version_override: "ruby2.6"
 bundler_version: "2.2.19"
-nodejs__upstream_release: 'node_12.x'
-nodejs__upstream_key_id: '68576280'
+desired_nodejs_version: "v12.22.6"
 rails_app_name: "abid"
 rails_app_directory: "abid"
 rails_app_symlinks: []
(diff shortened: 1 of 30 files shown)


kayiwa:
this will close #2828

group_vars/abid/production.yml
```diff
@@ -12,8 +12,7 @@ passenger_app_env: "production"
 passenger_ruby: "/usr/bin/ruby2.6"
 ruby_version_override: "ruby2.6"
 bundler_version: "2.2.19"
-nodejs__upstream_release: 'node_12.x'
-nodejs__upstream_key_id: '68576280'
+desired_nodejs_version: "v12.22.6"
```
jrgriffiniii, line 15: This is not a blocking criticism, however I am curious as to why the minor and patch versions are not tracking the latest release for https://nodejs.org/download/release/latest-erbium/?

group_vars/dpul_staging.yml
```diff
@@ -8,8 +8,7 @@ passenger_app_env: "staging"
 passenger_ruby: "/usr/bin/ruby2.6"
 ruby_version_override: "ruby2.6"
 bundler_version: "2.1.4"
-nodejs__upstream_release: 'node_12.x'
-nodejs__upstream_key_id: '68576280'
+desired_nodejs_version: "v12.22.10"
```
jrgriffiniii, line 11: 👍

jrgriffiniii (approved):
Inspecting https://github.com/pulibrary/princeton_ansible/blob/main/roles/nodejs/tasks/main.yml#L34, this seems perfect to me. Thank you very much as always @kayiwa!

christinach (dismissed):
@kayiwa can you also include approvals with node version v12.22.5 
It is what we have in the production box.

group_vars/abid/production.yml
kayiwa: The problem we encountered with latest is some folks preferred to pin versions and brought this up.

PR #2833, opened 2022-03-09
i2623 ezproxy log rotate

### LINKED ISSUE #2623: Add daily log rotate config to EZproxy Roles
Logs to rotate are in /var/local/ezproxy/log

kayiwa:
- add log rotation of log files
- make reference to need to use capistrano to complete
- add ezproxy 7 binary
- tasks will be replaced by capistrano

this closes #2623


roles/ezproxy/tasks/main.yml
```diff
@@ -133,26 +119,6 @@
     job: "/usr/bin/geoipupdate"
   when: running_on_server
 
-- name: create base config file
```
kevinreiss, line 136: Right now capistrano won't handle these two steps for config.txt.tmp or user.txt.tmp. I will have to check into to see if capistrano can copy files to a directory that resides outside of the capistrano directory tree.

kevinreiss, line 136: Right now capistrano won't handle these two steps for config.txt.tmp or user.txt.tmp. I will have to check into to see if capistrano can copy files to a directory that resides outside of the capistrano directory tree.

kevinreiss (approved):
I think we can move forward with this PR, but will probably need to create another issue or confirm if cap can handle the set-up of the config.txt and user.txt files in the EZproxy root directory.

PR #2847, opened 2022-03-14
add postgresql production variables for abid

kayiwa:
We are migrating abid to the new replicating postgresql environment

### DIFF
--- a/group_vars/postgresql/production.yml
+++ b/group_vars/postgresql/production.yml
@@ -36,7 +36,7 @@ postgres_admin_user: "{{ postgresadmin }}"
 datadog_api_key: "{{ vault_datadog_key }}"
 datadog_config:
   log_enabled: true
-datadog_checks:
+datadog_typed_checks:
   postgresql_log_check:
     init_config:
     logs:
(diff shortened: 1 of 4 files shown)


group_vars/postgresql/production.yml
```diff
@@ -34,17 +34,15 @@ postgres_port: 5432
 postgres_admin_user: "{{ postgresadmin }}"
 
 datadog_api_key: "{{ vault_datadog_key }}"
-datadog_service_name: postgresql
-datadog_environment: prod
 datadog_config:
   log_enabled: true
-  process_enabled: true
-datadog_typed_checks:
```
carolyncole, line 42: @kayiwa why did you remove the typed check?  This is the newer and more correct configuration for datadog

carolyncole (changes requested):

group_vars/postgresql/production.yml
kayiwa: Oh, I didn't realize this @carolyncole I am still debugging this. OK for this to go in? if I make it an issue.

PR #2855, opened 2022-03-16
add postgresql new server connnection for lib-statistics

kayiwa:
We are migrating the lib-statistics to the new postgresql servers


acozine:
@carolyncole we restored `ruby2.6` for the lib-statistics staging and production servers and ran the playbooks. I'll rebase this and then I think it can be merged.

PR #2862, opened 2022-03-21
Remove SVN variables

thanyabegum:
Remove SVN variables as part of resolving [#4955](https://github.com/pulibrary/figgy/issues/4955)


acozine (approved):
Looks good.

PR #2863, opened 2022-03-21
research data db migration

kayiwa:
- modify researchdata playbooks
- add postgresql new variables for research data


playbooks/researchdata.yml
```diff
@@ -1,12 +1,14 @@
...
-- hosts: researchdata_prod
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- name: build the research data site
+  hosts: researchdata_{{ runtime_env | default('staging') }}
   remote_user: pulsys
   become: true
   vars_files:
-    - ../group_vars/researchdata/researchdata_prod.yml
+    - ../group_vars/researchdata/staging.yml
```
acozine, line 9: I'd match the usual pattern, so vars get added in both environments:
```suggestion
    - ../group_vars/researchdata/{{ runtime_env | default('staging') }}.yml
```

playbooks/researchdata.yml
```diff
@@ -17,5 +19,5 @@
     - name: tell everyone on slack you ran an ansible playbook
       slack:
         token: "{{ vault_pul_slack_token }}"
-        msg: "{{ inventory_hostname }} completed"
+        msg: Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}
```
acozine, line 22: The quotes aren't necessary here - you only need to quote variables if the phrase/line would otherwise start with `{`.
```suggestion
        msg: Ansible ran {{ ansible_play_name }} on {{ inventory_hostname }}
```

group_vars/researchdata/production.yml
```diff
@@ -1,16 +1,28 @@
...
+pg_hba_method: "md5"
+pg_hba_source: "{{ ansible_host }}/32"
 postgres_admin_user: "{{ db_clusteradmin_user }}"
 postgres_admin_password: "{{ db_clusteradmin_password }}"
 postgresql_is_local: false
-postgres_version: 10
+postgres_version: 13
 drupal_db_host: '{{ postgres_host }}'
 drupal_db_name: 'researchdata_prod'
+drupal_db_user: "researchdata"
```
carolyncole, line 20: @kayiwa why was this change needed?  It is defaulted in the role: https://github.com/pulibrary/princeton_ansible/blob/main/roles/researchdata/defaults/main.yml#L3

group_vars/researchdata/production.yml
```diff
@@ -1,16 +1,28 @@
...
+db_host: "lib-postgres-prod3.princeton.edu"
 db_clusteradmin_password: "{{ vault_postgres_admin_password }}"
-db_clusteradmin_user: "{{ vault_postgres_admin_user }}"
+db_clusteradmin_user: "postgres"
+
+desired_nodejs_version: "v8.17.0"
 
 postgres_host: "{{ db_host }}"
 postgres_port: "5432"
+pg_hba_postgresql_user: "all"
```
carolyncole, line 10: It strange that this needs to be copied everywhere: https://github.com/pulibrary/princeton_ansible/search?q=pg_hba_postgresql_user   I would have expected the defaults to be part of the role.

carolyncole (changes requested):

group_vars/researchdata/production.yml
kayiwa: there will be instances where on wants to restrict more but I think I over stated its importance.

kayiwa: https://github.com/pulibrary/princeton_ansible/issues/2870

group_vars/researchdata/production.yml
kayiwa: I'm confused. the change on `group_vars/researchdata/production.yml` is there to ensure that we can do a dump and import of the file from `lib-postgres3` to `lib-postgres-prod3` the `playbooks/postgresql_db_migration.yml` file looks in there to determine what to dump. So it will not know of the existence of what is set up in `roles/researchdata/defaults/main.yml`

playbooks/researchdata.yml
```diff
@@ -17,5 +19,5 @@
     - name: tell everyone on slack you ran an ansible playbook
       slack:
         token: "{{ vault_pul_slack_token }}"
-        msg: "{{ inventory_hostname }} completed"
+        msg: Ansible ran {{ ansible_play_name }} on {{ inventory_hostname }}
```
acozine, line 22: Oh, I forgot! Those weren't quotation marks, they were backticks to make the `play name` stand out in Slack! The inventory_hostname stands out because slack interprets it as a link, but the `play name` needs the backticks. Sorry I took them out before - can we reinstate them?
```suggestion
        msg: Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}
```

group_vars/researchdata/production.yml
carolyncole: Ok, this makes me sad that we can not have defaults in roles... But you are utilizing the group vars for something other than the role/playbook they were created for.

carolyncole (approved):

Issue #2619, opened 2021-12-13
Redirect Requests for and Retire libimages


tpendragon:
- [x] Configure load balancer to redirect all libimages requests to iiif.princeton.edu
- [x] Find a way to handle the fact that Loris supports invalid slashes in the ID part of the URL.
- [x] Install SSL certificate for libimages onto lib-adc and configure it.
- [x] Point libimages server name to the load balancer directly.

See https://github.com/pulibrary/princeton_ansible/blob/deab5d10d94dc80aa16dcfb15e3de79c19f70820/roles/nginxplus/files/conf/http/libimages.conf#L49-L104 for an example of how to do this.

tpendragon:
Done!

PR #2865, opened 2022-03-21
Handle libimages at iiif.princeton.edu

tpendragon:
Refs #2619 and pulibrary/figgy#5022

I don't think we need a new certificate - I downloaded the cert info and it looks like @kayiwa was clever and added both `libimages` and `libimages1` to the same cert:

```
X509v3 Subject Alternative Name:
                DNS:libimages.princeton.edu, DNS:libimages1.princeton.edu
```


kayiwa (approved):
OK

PR #2867, opened 2022-03-21
Support loris2 links.

tpendragon:
libimages has loris2 paths for legacy resources.

Refs https://github.com/pulibrary/figgy/issues/5022


kayiwa (approved):
OK

PR #2872, opened 2022-03-22
collections-space staging is dead and it's not coming back

acozine:
Clearing the decks:

The load-balancer was reporting a failed host for collectionsspace-staging. The VM has been decommissioned and the project abandoned. This PR removes the config for the project.


kayiwa (approved):
OK!

PR #2881, opened 2022-03-23
remove clamav role

kayiwa:
dspace and (dataspace and oar) needed clamav installed and configured. We no
longer need this role


acozine (approved):
Looks good. Glad to get another unused role tidied away.

PR #2883, opened 2022-03-23
clean up ruby role

kayiwa:
- fix linting from molecule
- add fqdn modules for ruby role

We are cleaning up the ruby role to prepare for the upgrade


acozine:
The test failure is in the bibdata role, which is complaining that:
```
TASK [switch to Ruby "ruby2.6" release for the system default] *****************
  fatal: [instance]: FAILED! => {"changed": false, "msg": "no command given", "rc": 256}
```

The failing task currently reads:
```
- name: switch to Ruby "{{ ruby_version }}" release for the system default
  ansible.builtin.command: ruby-switch --set "{{ ruby_version }}"
  changed_when: false
```

The root of the problem may be that the `ruby-switch` package has disappeared or is hard to find: https://launchpad.net/ubuntu/+source/ruby-switch. Can we use a different approach to setting the system default ruby?

carolyncole (approved):

kayiwa:
superseded by #2934

Issue #2853, opened 2022-03-15
Ensure notified handlers run predictably


acozine:
Add `flush_handlers` at the end of each role, so if a playbook fails, handlers from successful roles run before the failure. This will make our playbooks more resilient. 

See https://docs.ansible.com/ansible/latest/user_guide/playbooks_handlers.html#controlling-when-handlers-run for details.

PR #2884, opened 2022-03-23
improve deploy user role

kayiwa:
- improve linting from molecule suggestions
- add fqdn modules for deploy_user role

In order to upgrade to a newer version of molecule we need to clean up our
linting


roles/deploy_user/tasks/main.yml
```diff
@@ -93,5 +93,5 @@
 # Note that this should not be necessary, but the `restart sshd` handler does not work
 - name: restart sshd for real
   become: true
-  command: systemctl restart sshd
+  ansible.builtin.command: systemctl restart sshd
   when: running_on_server
```
acozine, line 97: Using the `service` module here might fix this. But I think we can replace this with a `flush_handlers` task. Try replacing these four lines with this:
```
  - name: Run all handlers notified by the deploy_user role
    meta: flush_handlers
```
I think we might want this in all our roles, see #2853.

kayiwa: I used the flush handlers (because it is the ansible way) but that step is supposed to get skipped by the `running_on_server`

kayiwa (commented):
OK

PR #2885, opened 2022-03-23
Remove google cloud and monitor_agents groups

### LINKED ISSUE #2465: hosts still in [monitor_agents]
A bunch of hosts have been removed but are still in the [monitor_agents] group. We should remove them. In particular, old orangelight and bibdata hosts.

acozine:
Closes #2465.

Removes several `gcp` groups from the hosts file. I believe these were spun up for testing and are no longer extant - @kayiwa can you confirm?

Also removes the `monitor_agents` group, which we were using to add hosts to Sensu. We should be able to use parent/child groups to set up Sensu automation without duplicating host entries.

I would also like to alphabetize the groups in this file, to make things easier to maintain, but I'll do that in a separate PR.


kayiwa (approved):
OK

PR #2891, opened 2022-03-24
add tls certs for pdc-describe

kayiwa:


bess (approved):

acozine:
Failure came from the PHP role:
```
  TASK [common | install vector] *************************************************
  fatal: [instance]: FAILED! => {"msg": "The conditional check 'running_on_server' failed. The error was: error while evaluating conditional (running_on_server): 'running_on_server' is undefined\n\nThe error appears to be in '/home/runner/work/princeton_ansible/princeton_ansible/roles/common/tasks/main.yml': line 58, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: common | install vector\n  ^ here\n"}
```
Should be fixed by #2892.

acozine:
Rebased after merging #2892.

PR #2893, opened 2022-03-25
add pdc-describe configuration

kayiwa:
- add discovery path for production
- add pdc-describe configuration


bess (changes requested):
This should be for pdc-describe, but the config uses the name pdc-discovery.

bess (approved):

PR #2895, opened 2022-03-25
removes workaround for obsolete apt issue

acozine:
The task being removed here is causing idempotency failures on a bunch of roles.

The original problem was an apt repo that disappeared. The original workaround was added as part of #2499. I'm hoping that the broken repo has now been removed from our machines and we can delete the workaround task.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -1,13 +1,4 @@
 ---
-- name: Remove specified repository from sources list
-  ansible.builtin.apt_repository:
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
Will need to be rebased with main otherwise OK

acozine:
Note for Future Us: The second commit message is incomplete - it was supposed to read "prepare step also requires `running_on_server` to be set"; git must ignore anything inside backticks in a commit message.

PR #2898, opened 2022-03-28
Ruby role checks ruby version before installing

### LINKED ISSUE #2276: Ruby role succeeds even if ruby version used is wrong
lib-pulindexer 2 and 3 are both running ruby 2.4 although 2.6 is installed and present.

I'm seeing some claims that `ruby-switch` is gone as of ubuntu 14? Although it definitely still appears to be on the servers. https://github.com/mojolingo/brightbox-ruby-cookbook/commit/dbd0f6c22bc2c1e98ea76ef59379d0fb753cd308

acozine:
Closes #2276.

I tested this by:
- installing Ruby 2.7 using the ruby role
- running the role again with `ruby_version` set to `ruby2.6`
- SSH'ing to the box before, between, and after the two playbook runs

### DIFF
--- a/roles/ruby/tasks/main.yml
+++ b/roles/ruby/tasks/main.yml
@@ -9,22 +9,18 @@
   apt_repository:
     repo: "ppa:brightbox/ruby-ng"
     state: present
-  ignore_errors: true
+    update_cache: true
 
 - name: check ruby version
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK

PR #2899, opened 2022-03-28
use 'figgy_solr_name' and 'solr_port' to set production Solr URL

### LINKED ISSUE #182: Use figgy_solr_name variable in group_vars/figgy_production
for symmetry with the vars file for figgy staging, we should set `figgy_solr_name` to the correct value and use it below in the `SOLR_URL` env var definition.

Or else determine that it's not used and remove it, set `solr_cores` appropriately.

acozine:
Closes #182.

Updates the production group_vars for figgy to match the staging group_vars for figgy. Sets `SOLR_URL` from the variables `figgy_solr_name` and `solr_port` in production. 

This should not change behavior/outcomes. It just makes the two environments harmonize.


kayiwa (approved):
OK

PR #2901, opened 2022-03-29
PDC Describe server build

bess:
Work toward https://github.com/pulibrary/pdc_describe/issues/4


kayiwa (commented):
If you name the `group_vars/pdc_describe/common.yml` `group_vars/pdc_describe/main.yml` ansible will automatically read it, eliminating the need for https://github.com/pulibrary/princeton_ansible/blob/i4-pdc_describe_playbook/playbooks/pdc_describe.yml#L10

bess:
@kayiwa I re-named `common.yml` to `main.yml` but I found I still had to include it in the playbook. When I didn't include it it didn't read the variables defined in that file, which led to errors like it defaulting back to postgres 10 instead of 13: 

Screen Shot 2022-03-29 at 11 03 25 AM

acozine:
Automatic loading of `main.yml` only happens in roles.

PR #2871, opened 2022-03-22
Deduplicate bibdata playbooks

acozine:
Consolidates the bibdata production and staging playbooks. This will need additional changes once we have QA environments. I'll rebase as necessary. 

Related to #2618 and #2616.

PR #2902, opened 2022-03-29
Postgresql fixes

### LINKED ISSUE #2815: [Postgresql] multiple machines try to create the same database
When running on two machines, both machines try to create the database and we get an error.  The error is from running the new bibdata-qa playbook that has two machines.
```
TASK [postgresql : PostgreSQL | create postgresql database] **************************************************************************************************
ok: [bibdata-qa1.princeton.edu]
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: DETAIL:  Key (datname)=(bibdata_qa) already exists.
fatal: [bibdata-worker-qa1.princeton.edu]: FAILED! => {"changed": false, "msg": "Database query failed: duplicate key value violates unique constraint \"pg_database_datname_index\"\nDETAIL:  Key (datname)=(bibdata_qa) already exists.\n"}
```

acozine:
Closes #2815.

- Sets the task that creates databases to `run_once`.
- Corrects the name of the psql admin password var.
- Tried re-using `application_db_*` variables for the `pg_hba` configuration task, but ultimately gave database connections access to `all` in `pg_hba.conf` so they can create databases - this helped #2871 work but we may be able to change it in future - added a comment for Future Us.

### DIFF
--- a/roles/postgresql/defaults/main.yml
+++ b/roles/postgresql/defaults/main.yml
@@ -72,8 +72,10 @@ postgres_data_directory: "/var/lib/postgresql/{{ postgres_version }}/{{ postgres
 
 # HBA Connections
 postgres_hba_entries: []
-pg_hba_postgresql_user: "all" # "{{ application_dbuser_name }}"
-pg_hba_postgresql_database: "all" # "{{ application_db_name }}"
+pg_hba_postgresql_user: "all"
+pg_hba_postgresql_database: "all"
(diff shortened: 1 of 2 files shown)


acozine:
Related to #2871.

acozine:
See this [comment](decf558c913b39ffe981632b6ff9cc68111cfc53) for more about the `pg_hba` problem.

PR #2807, opened 2022-03-02
update python, ansible and molecule

kayiwa:
this is a security update updating our version of ansible to 4.2.0
we are also updating our python version to 3.10.x

PR #2907, opened 2022-03-30
Update requirements.txt for GitHub Actions

acozine:
#2807 updated the dependencies for our local testing environments.

This PR updates `requirements.txt`, which is [used in our CI environment](https://github.com/pulibrary/princeton_ansible/blob/cb91d0e1551251af95dbd451a5d751fafc8f8d55/.github/workflows/molecule_tests.yml#L123) to match the current [pipfile.lock](https://github.com/pulibrary/princeton_ansible/blob/cb91d0e1551251af95dbd451a5d751fafc8f8d55/Pipfile.lock), so CI testing will match local testing.

Are there other places we need to update for CI that aren't included here yet?


kayiwa (approved):
OK

PR #2920, opened 2022-03-31
Add default postgres port

sandbergja:


roles/postgresql/defaults/main.yml
```diff
@@ -14,6 +14,7 @@ postgres_apt_repository: 'deb http://apt.postgresql.org/pub/repos/apt/ {{ansible
 # Pin-Priority of PGDG repository
 postgres_apt_pin_priority: 500
 postgres_version: 13
+postgres_port: 5432
```
kayiwa, line 17: I would put this in `postgresql/vars` based on the precedence. https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence

sandbergja:
Force-pushed a commit to put this in vars/main.yml instead.  Thanks @kayiwa!

PR #2921, opened 2022-03-31
[ORANGELIGHT] Update nodejs on staging and qa

sandbergja:


christinach:
@kayiwa @acozine why the molecule tests dont run?

acozine:
I think we had a lot of PRs getting merged today, so the test system was backed up. 

To fix the failure that happened when the tests finally did run, we can try editing [.ansible-lint for this role](https://github.com/pulibrary/princeton_ansible/blob/ae5054049b071c717c37693c41ca9b75603f8fe8/roles/capistrano/.ansible-lint) and adding `- role-name` to it. But that might not fix the issue, see [this bug report. ](https://github.com/ansible/ansible-compat/issues/78).

sandbergja:
Just rebased against main and force-pushed.

acozine:
Looks good, thanks @sandbergja.

PR #2930, opened 2022-04-04
use ansible fully qualified names

kayiwa:


roles/common/tasks/main.yml
```diff
@@ -1,9 +1,4 @@
 ---
-- name: common | update cache
```
acozine, line 2: What's your thinking here? This isn't an FQCN change.

kayiwa: It is redundant to https://github.com/pulibrary/princeton_ansible/blob/i2861_common_fqrn/roles/common/tasks/main.yml#L20 
and otherwise not idempotent.

acozine (approved):

Issue #2852, opened 2022-03-15
Make Ansible alerts more specific


acozine:
Some ideas for making alerts from Ansible playbook runs even more useful:
- make the Slack alert a shared tasks file
- make the Slack channel a variable so playbooks can choose specific channels
- include the play name in the alert

Sample code:
```
    - name: tell everyone on slack you ran an ansible playbook
      slack:
        token: "{{ vault_pul_slack_token }}"
        msg: Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}
        channel: #server-alerts
```

acozine:
Test changing `channel: #server-alerts` to `channel: "{{ slack_channel }}"` and then passing a list, to see if we can alert a single playbook role in multiple channels.

acozine:
Can we add the branch that ran to the output?

acozine:
Most projects that still have duplicate, environment-specific playbooks still have the old message. We have almost 40 instances of `"{{ inventory_hostname }} completed"` to update.

acozine:
Remaining examples of `{{ inventory_hostname }} complete` as of today - all in the `playbooks` directory:
- [ ] cantaloupe_production.yml, cantaloupe_staging.yml (consolidate)
- [ ] confluence_staging.yml (can this be renamed?)
- [ ] figgy_production.yml, figgy_staging.yml (consolidate)
- [ ] lib_smb_serve_production.yml (what is this?)
- [ ] lib_statistics_staging.yml (what is this?)
- [ ] libruby_dev.yml, libruby_production.yml (consolidate)
- [ ] libstatic_production.yml, libstatic_staging.yml (consolidate)
- [ ] libwww_production.yml, libwww_staging.yml (consolidate)
- [ ] mariadb_gcp_dev.yml
- [ ] mariadb_production.yml (can this be deleted?)
- [ ] nginxplus_production_rebuild.yml
- [ ] oar_gcp_beta.yml, oar_gcp_dev.yml, oar_gcp_production.yml, oar_gcp_staging.yml (consolidate)
- [ ] oawaiver_production.yml, oawaiver_staging.yml (consolidate)
- [ ] openbooks_production.yml
- [ ] pdc_describe.yml
- [ ] pdc_discovery.yml
- [ ] postgresql_gcp_dev.yml, postgresql_gcp_production.yml, postgresql_gcp_staging.yml (consolidate, maybe delete?)
[...]

PR #2933, opened 2022-04-04
Includes ansible play name in slack alerts, uses FQCN for slack module

acozine:
Partial fix for #2852.

Starts updating playbooks to mention the play name when they send an alert to slack.
Uses `community.general.slack` in the slack task.
Updates playbooks starting with `a` and `b` and playbooks that have already been consolidated (see #2726).


kayiwa (approved):
OK

PR #2940, opened 2022-04-06
build ruby from source

kayiwa:


hackartisan:
does or could this ref an issue?

hackartisan:
oh i see the issue number in the branch name. i'll add it to the description.

hackartisan:
Actually I don't understand how this is related to that issue.

kayiwa:
hey @hackartisan it is wrongly referring to the issue. @bess mentioned wanting to run her apps with ruby 3 on the slack dev channel. This is the result of that.

roles/ruby_s/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for roles/ruby
+ruby_version: "{{ ruby_version_override | default(ruby_version_default) }}"
+ruby_version_default: "ruby-2.7.0"
+ubuntu_ruby_version: "{{ ubuntu_ruby_version_override | default(ubuntu_version_default) }}"
```
acozine, line 5: Typo here in the default setting:
```suggestion
ubuntu_ruby_version: "{{ ubuntu_ruby_version_override | default(ubuntu_ruby_version_default) }}"
```
Though I wonder if we can just set the default to ruby2.7 directly . . .

roles/ruby_s/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for roles/ruby
```
acozine, line 2:
```suggestion
# defaults file for roles/ruby_s
```

roles/ruby_s/handlers/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# handlers file for roles/ruby
```
acozine, line 2:
```suggestion
# handlers file for roles/ruby_s
```

roles/ruby_s/meta/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+galaxy_info:
+  role_name: ruby
```
acozine, line 3:
```suggestion
  role_name: ruby_s
```

roles/ruby_s/molecule/default/converge.yml
```diff
@@ -0,0 +1,15 @@
...
+  become: true
+  pre_tasks:
+    - name: update cache
+      apt:
+        update_cache: true
+        cache_valid_time: 600
+  tasks:
+    - name: "Include ruby"
+      include_role:
+        name: ruby
```
acozine, line 15: We want this to run the `ruby_s` role, not the `ruby` role.
```suggestion
        name: ruby_s
```

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+      name: ["{{ ubuntu_ruby_version }}", "{{ ubuntu_ruby_version }}-dev", "ruby-switch"]
+      state: absent
+      update_cache: true
+    when: switch is success
+
+  - name: rvm | remove the Brightbox repository
+    ansible.builtin.apt_repository:
+      repo: "ppa:brightbox/ruby-ng"
+      state: absent
+    when: switch is success
```
acozine, line 33: This whole block seems to be about removing pre-existing artifacts. We could consider putting it in a separate file, something like `tasks/cleanup.yml` and then including it when ruby is already installed.

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+
+  - name: get ruby url
+    ansible.builtin.set_fact:
+      ruby_gzip_url: "{{ ruby_values[1] }}"
+
+  - name: get ruby checksum
+    set_fact:
+      ruby_sha_256: "{{ ruby_values[3] }}"
+
+  - name: ruby | setup install directory
```
acozine, line 51: I might set this value in `defaults/main.yml` instead of using `set_fact` here. It's easier to override the role default.

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && ./configure --enable-shared creates={{ install_path }}/{{ ruby_version }}/Makefile
+
+  - name: ruby | make ruby
+    ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && make
+
+  - name: ruby | install ruby
+    ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && make install
+    become: true
+  # when applied to each of the tasks in the block above
+  when: installed_ruby.failed
```
acozine, line 84: Where are we setting/registering `installed_ruby`? Also, do we want something like `when: installed_ruby!=ruby_version`? Basically, we want to check what version of Ruby is on the box. Then if Ruby is not present, or if it is present but it's not the version we want, we download/build/install it.

roles/ruby_s/vars/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# vars file for roles/ruby
+ruby_switch_path: "/r/ruby-switch/"
```
acozine, line 3: We won't be using `ruby-switch` any more with this . . . will we?

acozine (commented):
This looks good. I added a few comments to make it better, mostly copy-pasta fixes.

roles/ruby_s/tasks/main.yml
kayiwa: we get this dynamically by `curl https://cache.ruby-lang.org/pub/ruby/index.txt` and getting the values. So I don't think it is easily possible. You?

roles/ruby_s/tasks/main.yml
kayiwa: thanks for the tips!

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
+---
+- name: ruby | check version
+  ansible.builtin.shell: ruby -v | awk '{print $2}'
+  register: installed_ruby
+  ignore_errors: true
```
acozine, line 5: Ignoring errors here seems risky. What kind of errors cropped up when you ran this?

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: ruby | separate index entry values
+  ansible.builtin.set_fact:
+    ruby_values: "{{ ruby_index_line.stdout | regex_findall('(\\S+)') }}"
+
+- name: get ruby url
+  ansible.builtin.set_fact:
+    ruby_gzip_url: "{{ ruby_values[1] }}"
+
+- name: get ruby checksum
+  set_fact:
```
acozine, line 27:
```suggestion
  ansible.builtin.set_fact:
```

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+
+- name: prints out the installed_ruby variables
+  ansible.builtin.debug:
+    msg: ruby version "{{ installed_ruby.stdout }}"
+
+- include_tasks: cleanup.yml
+  when: installed_ruby.stdout != "{{ desired_ruby_version }}"
+
+- name: ruby |lookup download path and checksum from ruby-lang release index
+  ansible.builtin.shell: curl https://cache.ruby-lang.org/pub/ruby/index.txt | grep {{ ruby_version }}.tar.gz
```
acozine, line 15: Could we use the `get_url` module instead? I'm not sure how to replicate the grep part, so maybe not, but it would be worth exploring.

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  ansible.builtin.set_fact:
+    ruby_gzip_url: "{{ ruby_values[1] }}"
+
+- name: get ruby checksum
+  set_fact:
+    ruby_sha_256: "{{ ruby_values[3] }}"
+
+- name: ruby | setup install directory
+  ansible.builtin.set_fact:
+    install_path: "~/install"
```
acozine, line 32: Why use `set_fact` for the install path? That's a value on the local machine. We can set a role default and let each project override it if they want. I might choose a default value that's not dependent on the user (something like `/opt/install`, maybe?).

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: ruby | configure ruby
+  ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && ./configure --enable-shared creates={{ install_path }}/{{ ruby_version }}/Makefile
+
+- name: ruby | make ruby
+  ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && make
+
+- name: ruby | install ruby
+  ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && make install
+  become: true
+
```
acozine, line 62: Do we need a task here to install bundler?

acozine (commented):
This is very exciting! I had a few suggestions and questions.

roles/ruby_s/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+- name: ruby | check version
+  ansible.builtin.shell: ruby -v | awk '{print $2}'
+  register: installed_ruby
+
+- name: prints out the installed_ruby variables
+  ansible.builtin.debug:
+    msg: ruby version "{{ installed_ruby.stdout }}"
+
+- include_tasks: cleanup.yml
+  when: installed_ruby.stdout != "{{ desired_ruby_version }}"
```
acozine, line 11: Per @carolyncole's comment today, we should update this condition to handle the case where no ruby is installed. Something like `when: installed_ruby.stdout != ""` in addition to the current condition.

acozine: Or possibly `when: installed_ruby.stdout | length > 0`

bess (approved):

PR #2942, opened 2022-04-08
Adding aspace production information

carolyncole:
Co-authored-by: regineheberlein <regineheberlein@gmail.com>

Run on staging.


acozine:
Per @regineheberlein these settings are correct.

PR #2944, opened 2022-04-08
remove dspace based playbooks we no longer use

kayiwa:
removed hosts of servers we no longer have


acozine:
More tidying! Looks good.

PR #2948, opened 2022-04-08
Set no_log on postgresql tasks with sensitive data

acozine:
One task in the postgresql role uses the `postgres` and `replicant` users, printing out their passwords in the output. Currently anyone who can run that role can also access those passwords, but as we move to Tower/the Platform, that will not necessarily be true.

This PR sets `no_log: true` on that task.


kayiwa (approved):
OK

PR #2954, opened 2022-04-11
improve ansible output on the slack server alert channel

kayiwa:


acozine:
Looks great, thanks @kayiwa!

Partial fix for #2852.

acozine (approved):

PR #2955, opened 2022-04-11
Revise indexcards redirect with request uri

rlskoeser:
small tweak to the indexcards redirect — I think based on how it's behaving that `$request_uri` includes the `/indexcards` portion of the url, since right now it's getting doubled  on the redirect

### DIFF
--- a/roles/nginxplus/files/conf/http/cdh_prod_geniza.conf
+++ b/roles/nginxplus/files/conf/http/cdh_prod_geniza.conf
@@ -65,7 +65,7 @@ server {
         health_check interval=10 fails=3 passes=2;
     }
     location /indexcards {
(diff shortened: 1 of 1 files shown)


roles/nginxplus/files/conf/http/cdh_prod_geniza.conf
```diff
@@ -65,7 +65,7 @@ server {
         health_check interval=10 fails=3 passes=2;
     }
     location /indexcards {
-        return 301 https://genizaprojects.princeton.edu/indexcards$request_uri;
+        return 301 https://genizaprojects.princeton.edu/$request_uri;
```
kayiwa, line 68: the pattern in the docs is https://genizaprojects.princeton.edu$request_uri; (no slash)

kayiwa (approved):
OK!

PR #2960, opened 2022-04-12
Use a 302 for redirect, 301 doesn't work

bess:


kayiwa (approved):
OK! sorry for suggesting 301. :-(

PR #2961, opened 2022-04-12
change second play name for nginx playbook

acozine:
Follow up on #2954.

Related to #2852.

With better Slack alerting, the name of the last play gets reported. This PR updates the name of the last play in the nginx playbook to make it more user-friendly and communicative.


kayiwa (approved):
OK

PR #2963, opened 2022-04-12
Upgrade pdc_describe to ruby 3.0.3

bess:
Fixes https://github.com/pulibrary/pdc_describe/issues/41

Screen Shot 2022-04-12 at 1 48 07 PM

@kayiwa and @acozine The new `ruby_s` role worked right away! 🎉

### DIFF
--- a/group_vars/pdc_describe/main.yml
+++ b/group_vars/pdc_describe/main.yml
@@ -14,8 +14,6 @@ pg_hba_source: "{{ ansible_host }}/32"
 ansible_python_interpreter: /usr/bin/python3
 
 passenger_app_root: "/opt/pdc_describe/current/public"
-passenger_ruby: "/usr/bin/ruby2.7"
-ruby_version_override: "ruby2.7"
 bundler_version: "2.2.19"
 desired_nodejs_version: "v12.22.2"
 rails_app_name: "pdc_describe"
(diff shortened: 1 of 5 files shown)


group_vars/pdc_describe/main.yml
```diff
@@ -30,3 +28,7 @@ application_host: '{{passenger_server_name}}'
 application_host_protocol: 'https'
 application_port: '443'
 project_db_host: '{{postgres_host}}'
+
+# Use Ruby 3.0.3 and install from source
+install_ruby_from_source: true
+ruby_version_override: "ruby-3.0.3"
```
acozine, line 34: Did the name of this variable come from the fact that the old way of installing ruby was meant to "override" the default Ubuntu installed ruby? At some point in the future, I would like to see this var use the same pattern as the nodeJS var (`desired_{package_name}_version`). This will require changes across multiple projects, so it's not an immediate thing . . .

roles/passenger/meta/main.yml
```diff
@@ -14,5 +14,8 @@ galaxy_info:
       versions:
         - 18.04
 
+# To install ruby from source, which you must do to use ruby 3+, define
+# install_ruby_from_source: true
 dependencies:
-  - role: ruby
+  - { role: 'ruby_s', when: install_ruby_from_source is defined and install_ruby_from_source|bool == true }
+  - { role: 'ruby', when: install_ruby_from_source is not defined or install_ruby_from_source|bool == false }
```
acozine, line 21: Nice!

acozine (approved):
Looks good, thanks for making it all work together.

kayiwa (approved):
OK

PR #2975, opened 2022-04-15
Renews SSL certificates for catalog-alma-qa, decommission slavery-prod

acozine:
catalog-alma-qa is due to expire on 4/29
~~slavery-prod is due to expire on 5/11~~

~~Renew both certificates for another year.~~

We are not using the `slavery-prod` alias any more. OIT has removed it from the load balancer; this PR removes all related certs.


kayiwa (approved):
OK

PR #2979, opened 2022-04-18
Adds a playbook for SSL cert renewals

acozine:
This playbook automates the first step (Create the Certificate Signing Request) of https://github.com/pulibrary/pul-the-hard-way/blob/main/services/create_ssl_certs.md.

### DIFF
--- a/playbooks/cert_renewal.yml
+++ b/playbooks/cert_renewal.yml
@@ -6,8 +6,6 @@
 # for example: '-e cert_path=/home/foo/ssl_certs/'
 - name: Generate a new cert and key
   hosts: localhost
-  vars:
-    cert_path: ~/.ssl/
   tasks:
   - name: Generate a new private key
     openssl_privatekey:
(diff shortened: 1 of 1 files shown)


playbooks/cert_renewal.yml
```diff
@@ -0,0 +1,31 @@
+---
+# you must pass two variables:
+# pass the first element of the FQDN as '-e host_name=value'
+# for example: '-e host_name=fun' would generate a cert for fun.princeton.edu
+# also pass the path of the directory where you want the key and cert to live
+# with no trailing slash
+# for example: '-e cert_path=/home/foo/ssl_certs'
```
carolyncole, line 7: @acozine it would be nice to have a full example and also a link for what to do with the CSR once you have it.

carolyncole (changes requested):
@acozine I love this idea!  Scripting what we have in PUL the hard way.

I asked for an additional update to the comments.

playbooks/cert_renewal.yml
acozine: @carolyncole thanks for the review - I've added a full example and some more documentation (plus the ability to add a SAN). Take a look and see what you think.

carolyncole (approved):

PR #2997, opened 2022-04-27
Rotate logs on bibdata-alma-workers boxes

acozine:
The logs for the bibdata workers are not rotating. There is a current entry in `/etc/logrotate.d/` for `marc_liberation` that comes [from the rails_app role](https://github.com/pulibrary/princeton_ansible/blob/5bdcd66a3c33601f99cd98e03b525792401babc2/roles/rails_app/templates/logrotate.d.rails.j2), and I have two guesses as to why it's not rotating things. First guess: the config points to `/opt/marc_liberation/current/log/`, which is a symlink for `/opt/marc_liberation/shared/log/`, so maybe logrotate is not following the symlink? Second guess: the logs never reached the minimum size of 500 MB (as of today, the two log files are under 10 MB each on both worker boxes, but they may have been manually rotated recently).

We probably need to do more testing/investigation and also gather more info on what the correct behavior would be. What size should the logs be allowed to reach? Or do we want to rotate them daily regardless of size?


acozine:
Related to #3060

acozine:
Related to #3060

acozine:
This change would rotate all the log files on the worker boxes daily, regardless of size. Test on bibdata-staging to confirm this behavior.

acozine:
We think this was fixed by #3084.

PR #3005, opened 2022-04-29
i2720 migrate locator

kayiwa:
- add mysql variables
- ask mysql dependency


playbooks/locator.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: build the locator site
+  hosts: locator_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/locator/{{ runtime_env | default('staging') }}.yml
+    - ../group_vars/locator/vault.yml
+  roles:
+    - role: locator
+    - role: datadog
```
acozine, line 13:
```suggestion
    - role: datadog
      when: runtime_env | default('staging') == "production"
```
Only install datadog in production. We can add datadog to staging machines temporarily but we should not "build it in" on staging machines.

acozine (commented):
I don't know much (anything, really) about the locator site, but this matches our patterns. One suggestion, otherwise looks good.

carolyncole (approved):
Deployed to production and is working

PR #3006, opened 2022-04-29
i2720 migrate fpul

kayiwa:
- add mysql variables
- cleanup playbooks


carolyncole (approved):
Deployed on production

PR #3017, opened 2022-05-06
Renews certificates for geoserver-staging and catalog

acozine:
Adds new certificates for:
geoserver-staging.princeton.edu
~~catalog.princeton.edu~~
lib-monitor.princeton.edu

Looking at the history, though, I see @kayiwa also renewed the certs for the catalog. Let's discuss on Monday. I can revert the change easily enough but let's be sure we're not duplicating work on certs.


kayiwa (approved):
OK! Thanks!

PR #3022, opened 2022-05-09
move cicognara away from LE certificates

kayiwa:
We have purchased TLS certs for cicognara that we will use going forward


kelynch (approved):

kayiwa:
The certificates are purchased from [gandi.net](https://shop.gandi.net/en/certificate/create) Right now @acozine and @gpmenos have access as owners. There's nothing stopping us from adding anyone to do this.

PR #3030, opened 2022-05-10
Adding instructions for running the playbook

carolyncole:

### DIFF
--- a/playbooks/deploy_user.yml
+++ b/playbooks/deploy_user.yml
@@ -1,4 +1,7 @@
 ---
+# This playbook should not be run on all hosts during business hours.  
+# If you are running it outside the maintenance window please run with a limit flag.
+# For example: `ansible-playbook playbooks/deploy_user.yml --limit slavery-prod1.princeton.edu`
 - name: Add deploy user
   hosts: all
   remote_user: pulsys


playbooks/deploy_user.yml
```diff
@@ -1,4 +1,7 @@
 ---
+# This playbook should not be run on all hosts during business hours.  
+# If you are running it outside the maintenance window please run with a limit flag.
```
hackartisan, line 3: Thank you for adding this warning! Can we say why it shouldn't be run during business hours?

hackartisan (changes requested):

hackartisan:
Also relevant: I created https://github.com/pulibrary/princeton_ansible/issues/3031 based on some discussion in #operations on April 28th.

playbooks/deploy_user.yml
carolyncole: Yes, this means that no one can deploy while it is running.  Out thought was to not run it during business hours because of this.  That can be added to the comment.

playbooks/deploy_user.yml
```diff
@@ -1,4 +1,7 @@
 ---
+# This playbook should not be run on all hosts during business hours.
```
acozine, line 2:
```suggestion
# Only run this playbook on all hosts in a maintenance window, because no one can deploy while the playbook is running. 
```

playbooks/deploy_user.yml
acozine, line 3:
```suggestion
# If you are running it outside the maintenance window please run it on one box at a time using a limit flag.
```

acozine (commented):
@hackartisan do these suggested changes address your concerns?

hackartisan (approved):

PR #3039, opened 2022-05-12
removes pulfalight surge boxes from hosts file

acozine:
Traffic to pulfalight no longer requires extra hardware. Removing the three surge VMs from the rotation. 

Related to #2911.


tpendragon (dismissed):
Should we remove this from the load balancer config at the same time?

acozine:
Heh, I didn't look at the link and assumed it was an issue, not a PR. I'll roll back my second commit and we can merge both PRs on Monday.

PR #3043, opened 2022-05-12
Adds role and playbook to install the sensu agent

acozine:
Automates installing and configuring the sensu-go-agent on any VM we want to monitor with Sensu. Once the agent is installed and configured, it connects to the Sensu backend.

Still lots we can do with this, but at least now we have a working first step.

Co-authored-by: Bess Sadler <bess@ibiblio.org>


kayiwa (approved):
Thank you! OK

PR #3044, opened 2022-05-12
temporary change- dont run sqs poller service in bibdata-qa

christinach:
I'd like not to load sqs-poller service in bibdata-qa.
This is a temporary change until we fix https://github.com/pulibrary/bibdata/issues/1867

### DIFF
--- a/playbooks/bibdata.yml
+++ b/playbooks/bibdata.yml
@@ -13,6 +13,8 @@
   roles:
     - role: roles/bibdata
     - role: roles/bibdata_sqs_poller
+      # https://github.com/pulibrary/princeton_ansible/issues/3041
+      when: runtime_env | default('staging') != "qa"
     - role: roles/sidekiq_worker
       when: "'worker' in inventory_hostname"
     - role: 'hr_share'


playbooks/bibdata.yml
```diff
@@ -13,6 +13,8 @@
   roles:
     - role: roles/bibdata
     - role: roles/bibdata_sqs_poller
+      # https://github.com/pulibrary/princeton_ansible/issues/3041
+      when: runtime_env | default('staging') !="qa"
```
christinach, line 17: It ran the role in bibdata-qa.
Tell it not to run it.

acozine, line 17: I think this is just a spacing issue. Try committing this and see if it works:
```suggestion
      when: runtime_env | default('staging') != "qa"
```

christinach: @acozine you are right! I'm sorry they are all skipped :) . Thank you!

christinach: I also ran it against staging to confirm that sqs-poller tasks are not skipped on staging.

PR #3046, opened 2022-05-12
add condition and comment to sqs_poller tasks

### LINKED ISSUE #2041: Need a shared place for the campus access file on bibdata1 and bibdata2
Because there are multiple production machines we need a shared location for the campus_access file.  Currently when you upload a file it will get updated on either 1 or 2. If it ends up on bibdata1 the exceptions never get into the main list being kept on bibdata2

### LINKED ISSUE #3041: [Bibdata-QA] Temporarily deactivate SQS poller service in bibdata qa

acozine:
Closes #3041.

With this change, the bibdata playbook should include tasks from `bibdata_sqs_poller` when `runtime_env` is set to `production` or to `staging`, and skip the tasks from `bibdata_sqs_poller` when `runtime_env` is set to `qa`.

Running the playbook won't remove anything from the QA boxes or decommission anything, but it will avoid putting things back onto those boxes.

### DIFF
--- a/playbooks/bibdata.yml
+++ b/playbooks/bibdata.yml
@@ -13,6 +13,8 @@
   roles:
     - role: roles/bibdata
     - role: roles/bibdata_sqs_poller
+      when: runtime_env | default('staging') != "qa"
+      # temporarily run the sqs_poller tasks in staging and prod but not in qa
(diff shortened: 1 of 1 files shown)


acozine:
Closing as duplicate of #3044.

PR #3054, opened 2022-05-18
globus playbook

kayiwa:
- add boto3 libraries
- add globus playbook


acozine:
When this PR goes in, we will need to update our environments to install the boto libraries. See #2807 for pointers.

PR #3083, opened 2022-06-08
remove the nginx role

kayiwa:


acozine:
Note for Future Us: to install nginx as an open-source standalone server, use the nginxplus role and pass a parameter.

PR #3088, opened 2022-06-09
rotate empty files

kayiwa:
allowing empty files to rotate will allow us to know that log rotation
is happening.


kayiwa:
this will close #3088

kevinreiss (approved):

carolyncole (approved):

PR #3091, opened 2022-06-10
i3090 migrate recap

kayiwa:
- add new recap playbook
- move mysql variables


group_vars/recap/staging.yml
```diff
@@ -0,0 +1,13 @@
...
+drupal_db_user: 'recap-staging'
+drupal_db_password: "{{ recap_db_password }}"
+drupal_db_name: 'recap-staging'
+drupal_db_port: '3306'
+drupal_docroot: '/var/www/recap_cap'
+apache_app_path: '{{ drupal_docroot }}/current'
+
+db_host: 'mysql-db-staging1.princeton.edu'
+
+drupal_major_version: 7
```
carolyncole, line 13: @kayiwa I think this should be drupal 9 here.  I realize that it is wrong already, but would you mind updating it.  I just do not want running the playbook to break something.

group_vars/recap/production.yml
```diff
@@ -54,4 +37,4 @@ datadog_checks:
         source: recap
         sourcecategory: recap
     tags: "recap, environment:production, role:recap"
-drupal_major_version: 8
\ No newline at end of file
+drupal_major_version: 8
```
carolyncole, line 40: @kayiwa This should be drupal 9 here. I realize that it is wrong already, but would you mind updating it. I just do not want running the playbook to break something.

carolyncole (changes requested):
Just the drupal version issue.  Sorry that was a issue we created and I just noticed it now.

PR #3102, opened 2022-06-16
replace expiring iiif-staging certs

kayiwa:


acozine (commented):
The cert looks weird - shouldn't all the text be either new (green) or removed (red)?

acozine:
Huh - it looks like all new chained certs have an unchanged section - see #3123. I withdraw my question.

PR #3107, opened 2022-06-20
we add conditional logging

kayiwa:
Do we really need to record health checks from the loadbalancers? This
PR adds conditional logging which eliminates all requests (usually
health checks) from the load balancer.
OK?

### DIFF
--- a/roles/passenger/templates/passenger.j2
+++ b/roles/passenger/templates/passenger.j2
@@ -11,7 +11,14 @@ server {
   set_real_ip_from 128.112.203.146;
   real_ip_header X-Real-IP;
 
-  access_log /var/log/nginx/access.log custom;
+  access_log /var/log/nginx/access.log custom if=$loadbalancer_enable;
 
   {{passenger_extra_config}}
+}
+
+map $remote_addr $loadbalancer_enable {
+     "128.112.203.144" 0;
+     "128.112.203.145" 0;
+     "128.112.203.146" 0;
+     default 1;
 }


roles/passenger/templates/passenger.j2
```diff
@@ -11,7 +11,14 @@ server {
   set_real_ip_from 128.112.203.146;
   real_ip_header X-Real-IP;
 
-  access_log /var/log/nginx/access.log custom;
+  access_log /var/log/nginx/access.log custom if=$loadbalancer_enable;
```
acozine, line 14: This variable name is a bit confusing - it sounds like we're ENabling something on the loadbalancers, when in fact, this would DISable log messages from the load balancers. How about a different var name?
```suggestion
  access_log /var/log/nginx/access.log custom if=$logging_enabled;
```

roles/passenger/templates/passenger.j2
```diff
@@ -11,7 +11,14 @@ server {
...
 
   {{passenger_extra_config}}
 }
+
+map $remote_addr $loadbalancer_enable {
+     "128.112.203.144" 0;
+     "128.112.203.145" 0;
+     "128.112.203.146" 0;
+     default 1;
+}
```
acozine, line 24:
```suggestion
# this filters out the load-balancer IPs, so we don't log health checks from them
map $remote_addr $logging_enabled {
     "128.112.203.144" 0;
     "128.112.203.145" 0;
     "128.112.203.146" 0;
     default 1;
}
```

acozine (commented):
This is a cool idea. We discussed it in the Ansible meeting today, and had some suggestions for making it even better.

Though we also wondered . . . do the LBs pass along the original IPs? Or would this block ALL request logging routed through the Load Balancers?

roles/passenger/templates/passenger.j2
kayiwa: the passenger would just stop writing the logs from the loadbalancers. It will record the upstream. I ran this branch on `catalog1.princeton.edu` for coherence check

hackartisan (approved):
nice!

PR #3114, opened 2022-06-23
Use ruby_s role rather than ruby role

### LINKED ISSUE #3112: Make sure lib_jobs can build
When creating an unrelated PR, we discovered that lib_jobs cannot currently build - Francis and Alicia said it was probably because we are currently using the `ruby` role and need to update to the `ruby_s` role.

maxkadel:
Closes #3112


kayiwa (approved):
whooo! Thank you! OK

PR #3121, opened 2022-06-24
For lib_jobs use production, not prod

maxkadel:


playbooks/lib_jobs.yml
```diff
@@ -12,7 +12,7 @@
   roles:
     - role: roles/lib_jobs
     - role: roles/datadog
-      when: runtime_env | default('staging') == "production"
+      when: runtime_env | default('staging') == "prod"
```
kayiwa, line 15: can we please retain production? change the `hosts` file to production and make sure the group vars also do?

kayiwa (approved):
Thanks! OK

PR #3123, opened 2022-06-30
refreshes cert and key for libdeploy.princeton.edu

acozine:
New SSL certificate and private key for the libdeploy.princeton.edu site.

The current certs expire in <30 days, and Datadog is alerting us.


kayiwa (approved):
thanks! OK

PR #3124, opened 2022-07-06
Remove Signal Sciences key and module

acozine:
Attempting to load the `sigsci` module for nginx often results in config errors:

```
TASK [../roles/nginxplus : nginxplus | make sure nginx configs are valid] ******************************************************
fatal: [lib-adc1.princeton.edu]: FAILED! => {"changed": false, "cmd": ["/usr/sbin/nginx", "-t"], "delta": "0:00:00.020288", "end": "2022-07-06 14:48:19.818542", "msg": "non-zero return code", "rc": 1, "start": "2022-07-06 14:48:19.798254", "stderr": "nginx: [emerg] module \"/etc/nginx/modules/ngx_http_sigsci_module.so\" version 1021005 instead of 1021006 in /etc/nginx/nginx.conf:9\nnginx: configuration file /etc/nginx/nginx.conf test failed", "stderr_lines": ["nginx: [emerg] module \"/etc/nginx/modules/ngx_http_sigsci_module.so\" version 1021005 instead of 1021006 in /etc/nginx/nginx.conf:9", "nginx: configuration file /etc/nginx/nginx.conf test failed"], "stdout": "", "stdout_lines": []}
```

For this reason, we are not renewing our Signal Sciences contract. This PR removes the `sigsci` module and key from our nginxplus role.


kayiwa (approved):
Thanks! We hardly knew ye! Signal Sciences.

PR #3131, opened 2022-07-12
[DSS] Make it so a box doesn't go down on passive health checks.

tpendragon:


christinach (approved):

acozine:
Looks good. See https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/ for documentation on the change.

PR #3141, opened 2022-07-21
enable nginx waf on staging

kayiwa:


christinach:
@kayiwa do you also want to add waf in catalog-qa? I see you added waf in bibdata-qa.

christinach (commented):
@kayiwa do you also want to add waf in catalog-qa? I see you added waf in bibdata-qa.

acozine:
Reviewed today in the Ansible Open House. Thumbs up all around.

PR #3148, opened 2022-07-25
Updating asdf instructions

carolyncole:
Include work around for ruby install on Mac Chip


README.md
```diff
@@ -10,13 +10,21 @@ Princeton Ansible Playbooks
...
+       asdf install
+       ```
+       **NOTE** You may encounter an error `include the header <string.h> or explicitly provide a declaration for 'memcmp'`.   Many thanks to folks here for solving this (https://github.com/openssl/openssl/issues/18720#issuecomment-1185940347). Instead run:
+       ```
+       optflags=-Wno-error=implicit-function-declaration ASDF_RUBY_BUILD_VERSION=v20220630 asdf install
+       ```
+    1. Otherwise:
+       1. `brew install python`
+       1. `brew install rbenv`
+ 1. `brew install pipenv`
```
hackartisan, line 26: I think when installing python via asdf I installed pipenv via pip. Does installing it via brew also install python via brew as a dependency? Or it successfully uses the asdf-installed python?

carolyncole: I am unsure.  I just know I needed it it in addition to the asdf install. Installing it via pip seems ok too, but that is undocumented.  @hackartisan did you want to offer how you installed it via pip?

carolyncole: The original instructions include the brew install https://github.com/pulibrary/princeton_ansible/pull/3148/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L17

hackartisan: It seems like the original instructions include `brew install pipenv` as part of the "otherwise" section.

it would be like
```
      asdf install
      pip install pipenv
       ```

And then indent the "brew install pipenv" to keep it in the "otherwise" section. does that make sense?

Sorry I can't really figure out how to do the code change suggestion.

File: README.md [hunk 2]
```diff
@@ -10,13 +10,21 @@ Princeton Ansible Playbooks
...
-  * `brew install python`
-  * `brew install pipenv`
-  * `brew install rbenv`
- * If you have errors when running pipenv sync in the Setup your Environment
+ 1. `brew install docker`
+ 1. Install ruby and python
+    1. If using asdf, install the plugins as listed in .tool-versions
+       ```
+       asdf install
+       ```
```
hackartisan, line 18:
```suggestion
       pip install pipenv
       ```
```

README.md
hackartisan, line 26:
```suggestion
       1. `brew install pipenv`
```

hackartisan: Okay I think I did it with code suggestions.

hackartisan (approved):
I think this is right. If not we'll get another shot when the next person tries it!

PR #3149, opened 2022-07-25
Updating common to pass molecule

carolyncole:
fixes 'fqcn-builtins: Use FQCN for builtin actions' lint errors

### DIFF
--- a/roles/common/molecule/default/verify.yml
+++ b/roles/common/molecule/default/verify.yml
@@ -4,7 +4,7 @@
   gather_facts: false
   tasks:
   - name: check common package status
-    package:
+    ansible.builtin.apt:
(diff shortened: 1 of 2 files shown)


roles/common/molecule/default/verify.yml
```diff
@@ -4,7 +4,7 @@
   gather_facts: false
   tasks:
   - name: check common package status
-    package:
+    ansible.builtin.package:
```
kayiwa, line 7: can we use `ansible.builtin.apt` instead.

carolyncole: Done @kayiwa!

carolyncole:
@kayiwa it is green ❤️

Issue #3162, opened 2022-07-26
Update local pipenv and test environment (GitHub Actions) with dependabot suggestions


acozine:
We want our test environment on GHA to match our local environments exactly. To make this happen:

1. from main, run `pipenv update`
2. make sure we've updated `cookiecutter` and `paramiko` (see #3144 and #3145) (looks like we can't do this one at a time?)
3. check in changes to the pipfile.lock
4. run entire test suite locally
5. once tests all pass, update the requirements.txt file with `pipenv lock --requirements`
6. push to a new PR, the dependabot PRs should autoclose when the changes are in
7. Add this set of steps to the README for future reference

PR #3167, opened 2022-07-27
fix dependabot recommandations

kayiwa:


kayiwa:
closes #3162

hackartisan (approved):
🎉 looks like they're in better sync than they were before.

hackartisan:
`pipenv sync` and `molecule test` (in the common module) and both succeeded for me locally 👍🏻

PR #3173, opened 2022-07-27
[Bibdata] Use correct file permissions, groups, and mount credentials

sandbergja:
- Currently, we are using the scratch credentials; we should be using the bibdata credentials.
- Fixes some group ownership for mounted directories
- Fixes a directory permission issue


christinach (commented):
I triggered the incremental job in the alma sandbox to make sure that the files will come with no permissions error and get indexed and they did. The difference that I see with alma prod after running this playbook is that the /data directory is now owned by deploy and www-data and not root.

PR #3187, opened 2022-08-03
add passenger-config to the postrotate

### LINKED ISSUE #3150: nginx logs write to wrong error.log file
On pulfalight prod machines the nginx errors are writing to `/var/log/nginx/error.log.1` instead of `/var/log/nginx/error.log`. Because of this the error logs never rotate (`notifempty` is configured) and `error.log.1` never compresses, simply growing until it takes up all the space on the machine.

Log rotate configuration seems fine, but we need to figure out why nginx is writing to the wrong file.

`/var/log/nginx/access.log` exhibits correct behavior.

When we delete `error.log.1` and restart nginx it writes to the correct location. However, as soon as logrotate runs it reverts to the bad state.

We tried removing the prerotate and postrotate sections of the logrotate configuration but that didn't resolve the bug.

The bug may have been introduced in June of 2022. The last error.log writes were from the 15th. and the last good rotated file on the pulfalight staging box is June 14th.

kayiwa:
this allows nginx to rotate the error logs
remove references to xenial which we no longer use

this closes #3150

### DIFF
--- a/roles/passenger/meta/main.yml
+++ b/roles/passenger/meta/main.yml
@@ -12,10 +12,10 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - bionic
 
 # To install ruby from source, which you must do to use ruby 3+, define
 # install_ruby_from_source: true
 dependencies:
-  - { role: 'ruby_s', when: install_ruby_from_source is defined and install_ruby_from_source|bool == true }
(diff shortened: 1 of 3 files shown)


hackartisan:
ran this on pulfa3-staging1 -- success!

```
-rw-r----- 1 www-data adm    58461 Aug  3 15:03 error.log
-rw-r----- 1 www-data adm   105474 Aug  3 15:02 error.log.1
```

hackartisan (approved):

PR #3201, opened 2022-08-04
Ensuring that the client_max_body_size is set to 0 for pdc-describe NGINX config. files

jrgriffiniii:
This ensures that the `413` errors are no longer raised for form `POST` file uploads


kayiwa (approved):
thanks James. OK

PR #3202, opened 2022-08-05
I2735 dpul postgres migration

kayiwa:


acozine:
Thanks for moving the dpul secrets from `group_vars/all/` into a separate file!

We will also merge this one after the migration is done.

hackartisan (approved):

PR #3218, opened 2022-08-10
add postgresql variables for lib-jobs-prod

kayiwa:


kayiwa:
this has been run on staging and production

PR #3224, opened 2022-08-10
Change secrets for orangelight

sandbergja:


christinach:
changed to draft to add the borrow direct key

christinach:
we've run this in production and updated the specific credentials. We also run the bibdata playbook to add the borrow direct key there as well.

kayiwa (approved):
thanks! OK

PR #3242, opened 2022-08-11
Update dpul rabbit credentials

hackartisan:

PR #3248, opened 2022-08-12
add application variables for lae

kayiwa:


kayiwa:
this has been run on staging and production

group_vars/lae/production.yml
```diff
@@ -23,48 +29,48 @@ rails_app_dependencies:
   - autoconf
 lae_db_name: 'lae_production'
 lae_db_user: 'lae'
-lae_db_password: '{{vault_lae_db_password}}'
-lae_rabbit_user: '{{vault_lae_rabbit_user}}'
+lae_db_password: '{{ vault_lae_db_password }}'
+lae_rabbit_user: '{{ vault_lae_rabbit_user }}'
```
hackartisan, line 33: rabbit user and password for prod and staging need to be updated to use the new user and password. see e.g. https://github.com/pulibrary/princeton_ansible/pull/3242/files

It's always been one password, but each app was storing it separately in its own group vars file. with the password rotation it was made more explicit that it's a single password and all the group vars should reference the central password.

hackartisan: I have a PR ready to do this.

hackartisan:
ran this latest commit, as well.

hackartisan (approved):

PR #3250, opened 2022-08-12
Update scsb auth key in bibdata and orangelight

christinach:
Adds scsb auth key on bibdata and orangelight.


kayiwa (approved):
Thanks! OK

PR #3254, opened 2022-08-12
run bin setup

kayiwa:
this adds protecting vault commits as part of a launching lastpass


sandbergja (approved):
Nice!  I like it.  Thanks, @kayiwa

PR #3256, opened 2022-08-12
Add datadog to dpul role

hackartisan:
not sure when this was lost, but dpul should run datadog.

### DIFF
--- a/playbooks/dpul.yml
+++ b/playbooks/dpul.yml
@@ -11,6 +11,7 @@
     - ../group_vars/dpul/vault.yml
   roles:
     - role: roles/dpul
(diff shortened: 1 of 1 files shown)


playbooks/dpul.yml
```diff
@@ -11,6 +11,7 @@
     - ../group_vars/dpul/vault.yml
   roles:
     - role: roles/dpul
+    - role: roles/datadog
```
kayiwa, line 14: can we perhaps add?
```
```

so datadog is not installed on none production?

kayiwa (approved):
Added a commit to make datadog installation to only go to production

PR #3258, opened 2022-08-12
Inventory idea

acozine:
This PR would revise our inventory, moving away from a flat file called `hosts` and using a directory instead. This approach allows teams more freedom to organize their project inventories as they like, and makes it easier to find, review, and maintain the VMs for a particular project. It also lets us run playbooks against subsets of inventory (for example, we could update packages, keys, and other shared resources on just the staging boxes).

In this branch I've created individual inventory files for the projects. I lumped some things we have roles and group_vars for - things like the zookeeper boxes and postgresql boxes - into a `shared` file for now. If we agree that this is a good approach generally, we can decide how granular we want those files to get.

Please note that I have NOT yet removed the existing `hosts` file on this branch. That will be the last step, after we all confirm that existing VMs/machines are included in the correct files in the `inventory` directory.

I also included an architecture decision file to be a basis for debate.


inventory/by_environment/production
```diff
@@ -0,0 +1,85 @@
+[production]
```
sandbergja, line 1: Could we use [groups of groups](https://docs.ansible.com/ansible/2.7/user_guide/intro_inventory.html#groups-of-groups-and-group-variables) here, rather than listing all the servers in multiple files?

The goal would be to reduce the burden of maintaining these, so you could add a new VM to an existing group, say libwww_prod, and it would automatically be included in the production and apache_servers groups.

For example:

```
[production:children]
abid_production
annotations_production
approvals_production
bibdata_production
...
```

acozine: That's a great idea. We'd have to be sure every machine is in some group somewhere, but I don't think that's too far of a reach.

I did enjoy alphabetizing and adding comments, though.

acozine:
Looked at in the Ansible Open House - agreed to break out the inventory to match group_vars (so, for example, put the zookeeper boxes in a separate file).

acozine:
Add to main README: 
- steps when creating a new project
In ADR:
- generally how inventory is structured
- why the directory is called `all_projects`

inventory/by_environment/production
acozine: I've noted this in the ADR, but I'll put it here on the PR as well:

Ansible loads inventory in alphabetical order (well, linux/ASCII order). You cannot create a parent group until after all the child groups are loaded. For this reason, the per-project files need to be listed before the per-environment files. I've created a sub-directory for the per-project inventory files called `all_projects`. 

If you change it to `projects` or `by_project`, the child groups will be loaded later, the `[production]` and `[qa]` and `[staging]` groups will not work, and you will see the error message `Unable to parse /path/to/inventory/by_environment/ as an inventory source`.

hackartisan (changes requested):
looks good to me; there's one comment that needs to be updated

inventory/by_service/apache
```diff
@@ -0,0 +1,16 @@
+[apache_servers]
```
sandbergja, line 1: Could this file use the group of groups pattern as well?

inventory/all_projects/figgy
```diff
@@ -0,0 +1,25 @@
+[figgy_staging]
+figgy-staging2.princeton.edu
+# no staging worker VMs?
```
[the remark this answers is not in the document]
acozine: Addressed in 8d9c4fcb37806a796bcabcfa895ff8bd752bf8bd. Thanks!

inventory/by_service/apache
acozine: Addressed in 8d9c4fcb37806a796bcabcfa895ff8bd752bf8bd. Thanks!

PR #3268, opened 2022-08-16
increased file upload size for pdc_describe

leefaisonr:
Fixes https://github.com/pulibrary/pdc_describe/issues/290


acozine:
The  http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size docs suggest that this is not the source of the issue, since set to `0` the load-balancer should not check the size of uploaded files.

Two possible fixes to investigate:
- look at the block size in the location stanza
- check the browser tab for the error code we're seeing here

leefaisonr:
Closing because this is not the answer :)

PR #3278, opened 2022-08-18
Mudd-dbs: file permissions to allow deploy

sandbergja:
With `0644` permissions on this directory and file, deploying to staging fails.  Switching to `0744` allows the deploy to go through.

I also added fully qualified module names.


kayiwa (approved):
Thanks! OK

PR #3280, opened 2022-08-23
i2735 pulmap db migration

kayiwa:
- modify files to match postgresql migration playbook
- add new postgresql db variables


kayiwa:
superseded by #3441

PR #3281, opened 2022-08-23
i2735 pulflalight db migration

kayiwa:
- modify playbook
- add new infra db variables


kayiwa:
successfully run and deployed to staging and production.

PR #3294, opened 2022-08-30
Refresh certs, improve cert renewal playbook

acozine:
Refreshes certs for oawaiver.princeton.edu and for ansible-tower.princeton.edu.

Updates the cert_renewal playbook so it provides the `.princeton.edu` for the main domain name and for any SAN names.


acozine:
Rebased.

carolyncole (approved):

PR #3297, opened 2022-08-30
Use scsb-prod key for catalog-staging

### LINKED ISSUE #3295: Fix orangelight-staging scsb key
Orangelight-staging SCSB key should be for scsb-production

maxkadel:
Closes #3295


acozine:
The `uat` system is for scsb to test their own changes, more of a dev env for them. It's not well suited for us to test changes in our software. So we will use production keys in our staging env.

acozine:
Retaining the `uat` key in the vault, in case we need to use it to test future releases of scsb.

PR #3303, opened 2022-08-31
we no longer need this TLS certs/keys

### LINKED ISSUE #3298: defunct TLS certificates on nginxplus
we do not need these TLS certs

kayiwa:
these services have either been moved and/or deprecated

Closes #3298


acozine:
Looks good, thanks @kayiwa!

PR #3309, opened 2022-09-01
Add Amp Up the Archives training to static site

### LINKED ISSUE #3136: Add Amp Up the Archives Training to the lib-static Playbook
Add new static site repository for amp up the archives training https://github.com/pulibrary/amp-up-training to the list of static sites deployed with this playbook. The content itself will be embedded in libguides where appropriate.

maxkadel:
- Adds "Amp Up the Archives" training static site
- Ensures correct permissions for cloned repositories

Closes #3136

### DIFF
--- a/roles/libstatic/tasks/main.yml
+++ b/roles/libstatic/tasks/main.yml
@@ -13,8 +13,31 @@
   tags:
     - install_keys
 
+- name: Grant permissions for deploy user
+  ansible.builtin.file:
+    path: "{{ item.doc_root }}"
+    state: directory
+    owner: "{{ deploy_user }}"
+    group: "{{ deploy_user }}"
+    follow: false
+    recurse: true
+    mode: 0755
+  with_items: '{{ named_sites + sites }}'
+  changed_when: false
+  tags:
+  - install_sites
+
+- name: Make repos safe
+  ansible.builtin.command: git config --global --add safe.directory '{{ item.doc_root }}'
+  become: true
+  with_items: '{{ named_sites + sites }}'
+  changed_when: false
+  tags:
+    - install_sites
+
+
 - name: Install all git repos
-  git:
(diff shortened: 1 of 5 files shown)


roles/libstatic/tasks/main.yml
```diff
@@ -25,6 +48,21 @@
   become: true
   notify: restart apache
   with_items: '{{ named_sites + sites }}'
+  changed_when: false
+  tags:
+    - install_sites
+
+- name: Grant permissions on deploy user
```
sandbergja, line 55: This task looks really similar to the one that starts on line 16.  Does it really need to be run twice?

roles/libstatic/tasks/main.yml
```diff
@@ -13,8 +13,31 @@
...
+  become: true
+  with_items: '{{ named_sites + sites }}'
+  changed_when: false
+  tags:
+    - install_sites
+
+
 - name: Install all git repos
-  git:
+  ansible.builtin.git:
```
sandbergja, line 40: 🎉

sandbergja (changes requested):
Thanks @maxkadel -- just a question about the file permission tasks

kevinreiss:
The archivist who built the site, Valencia Johnson, confirmed that the staging URL works fine when embedded in a libguide so we can move this into production when it's otherwise ready.

roles/libstatic/tasks/main.yml
maxkadel: It does, because when the files are cloned they are owned by the root user, and we have to re-set all the permissions in order to make the child files and folders have the correct permissions.

This PR follows the pattern in ezproxy and drupal - (`playbooks/roles/ezproxy/tasks/main.yml`, `playbooks/roles/drupal/tasks/main.yml`)

PR #3313, opened 2022-09-01
I3308 app configs is only for rails

### LINKED ISSUE #3308: pul_library_drupal - resolve distracting error
When deploying to pul_library_drupal (aka libwww), the line `/home/deploy/.bashrc: line 1: /home/deploy/app_configs/*: No such file or directory` appears 92 times (see, for example, https://gist.github.com/pulbot/2f95c3847377eca8942f1d28603d0822). 

This makes the logs more difficult to read and is distracting. The environment file appears to live at `"/home/{{ deploy_user }}/.env.local"` currently. 

Adding `for f in ~/app_configs/*; do source $f; done` to `/home/deploy/.bashrc` appears to be part of the deploy_user role / playbook.

sandbergja:
closes #3308 

The .bashrc line that causes this error was previously added as part of the deploy_user role, but it is really only needed in the rails_app role, so I moved it to the rails_app role.

This PR also includes a separate one-time playbook to remove all those .bashrc lines from the apache servers.  I wasn't sure if it should be part of this repo or not, since it probably won't be very useful after it is run...

[Example deployment logs before](https://gist.github.com/pulbot/5d972e59c4b436856ff9bbbc4379d89f) running the remove_app_configs playbook vs. [after](https://gist.github.com/pulbot/d6fe94e21d1cc10767d3262b867d20d2)

### DIFF
--- a/playbooks/remove_app_configs.yml
+++ b/playbooks/remove_app_configs.yml
@@ -1,0 +1,19 @@
+---
+- name: Remove app_configs checking from apache servers
+  hosts: apache_servers
+  remote_user: pulsys
+  become: true
+  become_user: deploy
+  tasks:
+    - name: don't check for app_configs in .bashrc
+      ansible.builtin.lineinfile:
+        dest: "/home/deploy/.bashrc"
+        state: absent
+        regexp: '^for f in ~/app_configs/\*; do source \$f; done$'
+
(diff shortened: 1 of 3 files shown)


kayiwa (approved):
Thank you! OK

PR #3325, opened 2022-09-09
keep lib-postgres3 only for prod

acozine:
Today we saw outages in production when we ran the new "apt update" playbook against all staging machines. Traced the problem to an inventory issue.

Because `lib-postgres3` contains both production and staging databases, originally we put the machine in both the `postgresql_production` and `postgresql_staging` groups. Keeping it in only the production group for now.

The eventual fix for this problem is to separate prod and staging fully, see #2735.


kayiwa (approved):
thank you! OK

PR #3330, opened 2022-09-13
updates all bare 'include' statements to 'include_tasks'

### LINKED ISSUE #3305: fix deprecrated include
In Ansible 2.16 this will need to be replaced with include_task or include_role

acozine:
Closes #3305.

The use of `include` without any indication of what to include (role? tasks? vars?) is deprecated and the functionality will disappear in Ansible 2.16. Using the specific directives is clearer, anyway.

This PR updates the few instances we still had of generic `include` directives.

### DIFF
--- a/group_vars/ojs/main.yml
+++ b/group_vars/ojs/main.yml
@@ -19,6 +19,7 @@ pg_hba_method: "md5"
 pg_hba_source: "{{ ansible_host }}/32"
 postgres_version: "13"
 
+ojs_site_adminname: "{{ vault_ojs_site_adminname }}"
 ojs_site_adminpassword: "{{ vault_ojs_site_adminpassword }}"
 ojs_db_host: "{{ project_db_host }}"
 ojs_site_repoid: "pul_ojs"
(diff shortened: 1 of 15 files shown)


acozine:
This is what I get for not running the tests locally 😞.

acozine:
Current failure is `"AnsibleUndefinedVariable: {{ vault_ojs_site_adminname }}: 'vault_ojs_site_adminname' is undefined"`. But I've looked in group_vars and that vault variable does exist. Do we need to load the group_vars explicitly?

acozine:
Note for Future Us: the problem turned out to be that the default variables in the role were pointing to the vaulted variables. We do not decrypt vaulted variables in CI, so the tests were failing because some variables were not defined. The solution here was to provide fake values in the role's `defaults/main.yml` file for use in CI. The playbook pulls in `group_vars` and those files point to the vaulted variables. See, for example, [this line](https://github.com/pulibrary/princeton_ansible/blob/d5bb4fc6bb44047502062cb65e154b443136cfa8/group_vars/ojs/main.yml#L22).

acozine:
Error in the tests for the `composer` role is actually coming from the `common` role: `"Failure downloading https://github.com/bootandy/dust/releases/download//v0.8.3/du-dust_0.8.3_amd64.deb, HTTP Error 502: Bad Gateway"`.

bess (approved):
LGTM!

bess (approved):

kayiwa:
when we encounter loads of `changed_when: false` might want to use the pattern in this PR

PR #3343, opened 2022-09-20
remove nginxplus production

kayiwa:
we are removing the loadbalancers because having both restart
immediately after the other would be disastrous

Co-authored-by: Alicia Cozine<acozine@users.noreply.github.com>
Co-authored-by: Bess Sadler <bess@users.noreply.github.com>
Co-authored-by: Christina Chortaria <christinach@users.noreply.github.com>
Co-authored-by: Jane Sandberg <sandbergja@users.noreply.github.com>
Co-authored-by: Robert-Anthony Lee-Faison <leefaisonr@users.noreply.github.com>

### DIFF
--- a/inventory/by_environment/production
+++ b/inventory/by_environment/production
@@ -19,7 +19,6 @@ locator_production
 lockers_and_study_spaces_production
 mudd_production
 mysql_production
-nginxplus_production
 oawaiver_production
 obsd_httpd_production
 ojs_production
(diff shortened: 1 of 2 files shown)


acozine:
Could you also update the [comment in the os_updates playbook](https://github.com/pulibrary/princeton_ansible/blob/f980f649430b801e4686ef87aad13a4f9d513167/playbooks/os_updates.yml#L5-L8) so it's clear the playbook will not run on the load balancers when we pass `production` as the group?

PR #3345, opened 2022-09-20
Upgrade Pulfalight Ruby to 3.1

### LINKED ISSUE #3344: Update pulfalight to ruby 3.1
Ruby 2.x is EOL soon. Use Ruby 3.1

eliotjordan:
Closes #3344


group_vars/pulfalight/production.yml
```diff
@@ -17,8 +17,8 @@ solr_mirror: "https://archive.apache.org/dist"
 passenger_server_name: "findingaids.princeton.edu"
 passenger_app_root: "/opt/pulfalight/current/public"
 passenger_app_env: "production"
-passenger_ruby: "/usr/bin/ruby2.7"
-ruby_version_override: "ruby2.7"
+passenger_ruby: "/usr/bin/ruby3.1"
```
kayiwa, line 20: you know about the `ruby_s` role?

eliotjordan: I'm having some difficulty with this and set it down at the end of the day. I did try the `ruby_s` role, but then the playbook also ran the `ruby` role afterwards and failed.

kayiwa: always happy to pair. If @acozine and I ran the world we'd hoped moving to ruby 3 would sunset the ruby role altogether.

eliotjordan: Thanks. I think I've figured it out. The passenger role already depends on the `ruby` or `ruby_s` role so we don't need to explicitly call it in the pulfalight playbook. Updated PR coming shortly...

hackartisan (approved):

PR #3346, opened 2022-09-21
add all the cdh vms to inventory

kayiwa:
right now the cdh vms are not getting the weekly updates.
this addresses that

### DIFF
--- a/inventory/all_projects/pemm
+++ b/inventory/all_projects/pemm
@@ -1,4 +1,0 @@
-[pemm_staging]
-cdh-test-pemm1.princeton.edu
-[pemm_production]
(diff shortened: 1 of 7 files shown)


inventory/all_projects/pemm
```diff
@@ -0,0 +1,4 @@
+[pemm_staging]
+cdh-test-pemm1.princeton.edu
+[pemm_production]
+cdh-pemm1.princeton.edu
```
rlskoeser, line 4: we're just about to decommission the pemm servers, remember? (I think you were going to wait a week before shutting them down; you already turned off the public view)

rlskoeser (approved):

PR #3361, opened 2022-09-23
[Researchdata] Upgrade PHP to 8.1

sandbergja:


group_vars/researchdata/common.yml
```diff
@@ -0,0 +1,2 @@
+---
+php_version: '8.1'
```
kayiwa, line 2: why not go to 8.2 to prevent maintenance?

kayiwa (approved):
Thank you! OK!

PR #3362, opened 2022-09-26
fix molecule error in nfsserver role

acozine:
This role failed over and over again in #3024, so I decided to fix it separately.


kayiwa (approved):
Thank you! OK

PR #3364, opened 2022-09-26
i3349 fpul php81

kayiwa:
- fix linting
- update to php 8.1


sandbergja (changes requested):
Thanks, @kayiwa .  Unfortunately, it looks this site relies on the [drupal feeds module](https://www.drupal.org/project/feeds), which is [not yet PHP 8.1 compatible](https://www.drupal.org/project/feeds/issues/3254188).  So I think we'll have to wait for that to be addressed before upgrading the box.

kayiwa:
this work was already done by another PR.

Issue #3013, opened 2022-05-03
Make apt cache_update a separate playbook or create a maintenance role


acozine:
To avoid the problem where failure to update the apt cache makes a playbook fail (see https://github.com/pulibrary/princeton_ansible/issues/3009#issuecomment-1115109833), let's rethink the way we update the cache for apt with Ansible.

- [ ] audit roles for `update_cache: true` - probably use it once per role, mark that task `ignore_failures` so playbook runs can continue (and updates do happen when they can)
- [ ] create a maintenance playbook or role where a failure in the apt-cache update WILL make the playbook fail, add tasks to repair the keys there
- [ ] run maintenance playbook across our infrastructure, possibly quarterly?

acozine:
We can set this playbook to run on a schedule through Ansible Tower.

acozine:
Related to #3009.

carolyncole:
Remove #3283 changes when this is complete.

acozine:
Related to #3269.

acozine:
Related to #3206

acozine:
Closed by #3024.

PR #3365, opened 2022-09-26
Ensure all roles update the cache for CI

acozine:
Partial fix for #3013.

Without a pre-task to update the cache, the common role usually fails with `fatal: [instance]: FAILED! => {"changed": false, "msg": "No package matching 'htop' is available"}`. So we need to update the cache explicitly for testing before we can remove the `update_cache: true` settings from tasks within various roles.


kayiwa (approved):
THANKS! SO MUCH! OK

PR #3367, opened 2022-09-26
Update drupal9 and PHP roles to pass CI

### LINKED ISSUE #3013: Make apt cache_update a separate playbook or create a maintenance role

acozine:
A first attempt to fix #3013 ran into an idempotence failure on the `drupal9` role. This turned out be caused by a conflict in apt. You cannot install both `nodejs-dev` and `php7.4-dev` with apt - when you install the second one, apt uninstalls the first one.

We agreed to install NodeJS from source rather than as an apt package.

Fixing this problem uncovered a failure in the PHP role. Although the role accepted either apache2 or nginx as its webserver, some tasks hardcoded apache2 as the server to restart.

This PR fixes both of these issues.

### DIFF
--- a/roles/drupal9/molecule/default/converge.yml
+++ b/roles/drupal9/molecule/default/converge.yml
@@ -9,7 +9,6 @@
     - systems_user: 'deploy'
     - force_settings: false
     - running_on_server: false
-    - php_webserver: 'apache2'
   pre_tasks:
     - name: update cache
       apt:
--- a/roles/drupal9/molecule/default/converge.yml
+++ b/roles/drupal9/molecule/default/converge.yml
@@ -9,6 +9,7 @@
     - systems_user: 'deploy'
     - force_settings: false
     - running_on_server: false
+    - php_webserver: 'apache2'
   pre_tasks:
     - name: update cache
       apt:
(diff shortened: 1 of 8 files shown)


roles/drupal9/molecule/default/converge.yml
```diff
@@ -9,6 +9,7 @@
     - systems_user: 'deploy'
     - force_settings: false
     - running_on_server: false
+    - php_webserver: 'apache2'
```
acozine, line 12: Omit this for testing or change to nginx to match the role vars and also our overall strategy of "move away from apache and mysql when possible, use nginx and postgresql."

kayiwa (approved):
thanks so much for unblocking me @acozine! OK

PR #3371, opened 2022-09-28
follow globus certificate update

kayiwa:
convert the documented [steps here](https://docs.globus.org/globus-connect-server/v5.4/#upgrading_globus_connect_server) into a playbook to update globus certificates

### DIFF
--- a/playbooks/globus_certs.yml
+++ b/playbooks/globus_certs.yml
@@ -1,3 +1,4 @@
+# to run in production, pass '-e runtime_env=production'
 ---
 - name: update globus certificates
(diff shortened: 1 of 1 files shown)


carolyncole (approved):

playbooks/globus_certs.yml
```diff
@@ -0,0 +1,24 @@
+---
```
carolyncole, line 1:
```suggestion
# to run in production, pass '-e runtime_env=production'
---
```

carolyncole:
@kayiwa I added in the running in production comment.  🤞 you are ok with that... Once it is green again either you or I can merge it...

Issue #3117, opened 2022-06-23
completely uninstall all ruby


kayiwa:
The `ruby_s` role should completely uninstall all versions of ruby that are
still left on a vm. This is especially needed when one is trying to use `ruby_s`
for the ruby version 2

acozine:
This ticket was prompted by a problem updating lib-jobs.princeton.edu in mid-June, after changing the playbook to run the `ruby_s` role instead of the `ruby` role. The playbook ran fine, but the first deploy after the playbook run failed. On investigation, it turned out there were multiple versions of ruby installed on the box. We suspect that the ruby version passenger was using was not the newly installed version. Gist of the run: https://gist.github.com/pulbot/ad1588075adb2d5bca178211089cdbfb

acozine:
Related to #3114.

acozine:
I think this will be solved/fixed by 808a249c8c6dd213f82dd6b6b4b1b2981374e982, which is part of #3373.

acozine:
If we want to use the ruby_s role to upgrade/downgrade ruby versions, we also need to run `make uninstall` from the directory for the old ruby version before installing the new one.

acozine:
Closed (as "good enough") by #3414.

PR #3373, opened 2022-09-30
[Bibdata] Remove stale apt ruby versions in bibdata when running the ruby_s role. Install ruby 3.1

### LINKED ISSUE #3368: Upgrade bibdata to ruby 3.1

christinach:
closes #3368 
Update bundler to 2.3.3


acozine:
@christinach and I tried this branch against staging and it installed bunder 2.3.22 instead of the required bundler version of 2.3.3.

At this point it's difficult to revert to the old ruby because it was installed with the oldruby role, whereas the new ruby has been installed by the ruby_s role.

acozine:
Next steps: try changing the bibdata role to use `ruby_s` to install ruby 2.7.5 (change the install method, but not the ruby version) on staging. Then confirm that everything on the staging box is using the correct ruby and bundler versions from the correct paths. Then try to upgrade to ruby 3.x and see if we can fix the bundler version problem.

acozine:
@christinach and I will work on this again next week.

Also related to #3117.

christinach:
@acozine I downgraded the bundler version to 2.3.24 and was able to deploy ruby 3.1. I noticed the ruby_test.yml file you added only for bibdata-alma-worker-staging1. I can see `usr/lib/ruby/vendor_ruby/phusion_passenger` passenger in this machine. I cannot see `usr/lib/ruby/vendor_ruby/phusion_passenger` on the server machine bibdata-alma-staging1 and this is why I get a forbidden status only from this machine; I found this in the nginx logs `pulsys@bibdata-alma-staging1:~$ sudo tail -f /var/log/nginx/error.log`: 'directory index of "/opt/bibdata/current/public/" is forbidden,'. 
In the bibdata-alma-worker-stgaing1 machine I checked if passenger is loading by running `/usr/bin/passenger-config --root`. Then  I tunneled in bibdata-alma-worker-staging1 where you have the ruby_test to run. Passenger is there but fails to load because of a gem version that I need to figure out how to change it. I will try again to install a different bundler version that supports the specific default gem that throws the error or I will start the ruby 3.1 upgrade from scratch. See you next week!

acozine:
Idempotency fails because we are comparing the output of `ruby -v` to the bare version variable. Once we add the regex and tasks mentioned in 76795b0c461d3854f35a1bf200927fc90d4af502 this should pass CI and also be able to handle upgrading and downgrading ruby installed from source.

acozine:
The title of this PR is a bit misleading at this point - it's a refactor of the `ruby_s` role. The main changes are:
- puts the tasks that install ruby from source in a separate file
- only runs those tasks when ruby is missing or the wrong version
- gets rid of most `changed_when: false` params, since the new logic maintains idempotency without them
- adds a `make uninstall` step so the role can upgrade/downgrade ruby when it was previously installed from source
- changes the tasks that remove ruby installed by apt so we don't need to know what version is installed on a box
- moves the variables for bibdata into `group_vars` and creates a `group_vars/common.yml` file
Testing encouraged on projects beyond bibdata!

acozine:
Okay, it is time to remove the `update_cache` task from the common role. This PR is failing CI with:
```
   CRITICAL Idempotence test failed because of the following tasks:
*  => update cache
```

christinach:
rebase this when #3414 is merged

christinach:
Today when I re ran the playbook I noticed that even though we specify bundler version 2.3.24 , it installed 2.3.25 which was released today Nov 2nd. It didn't create any problem when I deployed the app.

kayiwa:
superseded by #3415

PR #3375, opened 2022-09-30
reduces number of mysql logs we send to datadog

acozine:
We have seen >800K logs sent from mysql to datadog, compared to <36K for other things we monitor. The additional logging may have been added for a debugging session at some point and never removed. We are not using it intensively enough (at all?) to justify the added cost.

The PR would stop sending everything except the error logs. @kayiwa does that seem like the right thing, at least for now?


carolyncole (approved):

kayiwa (approved):
OK!

PR #3374, opened 2022-09-30
tls renewal for multiple names

kayiwa:

PR #3376, opened 2022-10-03
cert_renewal playbook accepts a list of certs, with a list of SAN names

acozine:
Closes #3374.

Instead of passing extra vars at the command line, the revamped playbook uses a `vars:` list inside the playbook.

You can renew multiple certs in a one playbook run, including some with no SAN names, some with a single SAN name, and some with multiple SAN names.


playbooks/cert_renewal.yml
```diff
@@ -1,53 +1,61 @@
...
+  #         - DNS:foo.princeton.edu
+  #     - name: nuts
+  #     - name: plain
+  #       SAN_names:
+  #         - DNS:nut-free.princeton.edu
+  #         - DNS:allergy-friendly.princeton.edu
+  #   cert_path: /Users/foo/ssl_certs
   tasks:
+  - name: Make a subdirectory
+    file:
```
kayiwa, line 36: FQRN please

playbooks/cert_renewal.yml
```diff
@@ -1,53 +1,61 @@
...
+      path: "{{ cert_path }}/{{ item.name }}"
+      state: directory
+    loop: "{{ host_names }}"
+
   - name: Generate a new private key
     openssl_privatekey:
```
kayiwa, line 42: same here.

playbooks/cert_renewal.yml
acozine: Thanks for catching that. Addressed by b2c2f94fdabd468d66426badc6a4d3a33ec6588b.

playbooks/cert_renewal.yml
acozine: Also addressed by b2c2f94fdabd468d66426badc6a4d3a33ec6588b - thanks again @kayiwa.

Issue #2724, opened 2022-02-04
Support removing SSH keys


acozine:
We use the [authorized_key module](https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html) to add SSH keys. The current playbook for adding keys uses a loop, which is efficient and works well for adding keys. However, it only adds keys, it does not remove keys that are removed from the list. You can add set `exclusive: yes`, which removes all keys except the ones specified in the current task, but the action is "not loop aware". Meaning, it removes all keys each time, so if you run it on a list, you end up with only the current keys for the last entry in the list.

We like the efficiency of using a loop for adding keys. So we've decided to create a separate playbook for removing keys, one that will contain a separate task for each user key, with `exclusive: yes` set.

acozine:
We decided to build the `authorized_keys` file from a template instead of using the module and having a separate playbook. Now the `authorized_keys` file is regenerated fully on every run, so any keys we remove from our list will be removed from the `.ssh/authorized_keys` file on the next playbook run.

acozine:
Closed by #3377 and by https://github.com/PrincetonUniversityLibrary/infra/pull/30.

PR #3377, opened 2022-10-04
Builds deploy authorized_keys from a template

acozine:
Partial fix for #2724.

The deploy_user role pulls key content from GitHub and from files, then builds an authorized_keys file from a template instead of using the `authorized_key` module. We should be able to use this pattern to manage and track service accounts as well.


roles/deploy_user/tasks/main.yml
```diff
@@ -12,11 +12,37 @@
     home: "/home/{{ deploy_user }}"
     shell: "{{ deploy_user_shell }}"
 
-- name: deploy_user | add SSH keys
-  ansible.posix.authorized_key:
-    user: "{{ deploy_user }}"
-    key: "{{ item.key }}"
-  with_items: "{{ deploy_ssh_users }}"
+- name: deploy_user | get key content from github
```
carolyncole, line 15: It would be nice to have a comment here, that this is utilized in the template.

roles/deploy_user/tasks/main.yml
```diff
@@ -12,11 +12,37 @@
...
+- name: deploy_user | build authorized keys file
+  ansible.builtin.template:
+    src: authorized_keys.j2
+    dest: /home/{{ deploy_user }}/.ssh/authorized_keys
+    group: "{{ deploy_user }}"
+    owner: "{{ deploy_user }}"
+    mode: '0600'
+    backup: true
+
+# - name: deploy_user | add SSH keys
```
carolyncole, line 39: Is this needed?

carolyncole (changes requested):
@acozine looks nice! I just had two small comments

roles/deploy_user/tasks/main.yml
acozine: I think the commented task can go - I meant to remove it and just never got around to it.

@kayiwa remember that this is for the `deploy` user, so it's less urgent than for our usual SSH user. Also, the new task sets `backup: true` so when the `authorized_keys` file changes, we will always have the previous file available on the VMs.

roles/deploy_user/tasks/main.yml
acozine: addressed in 5691a35087107bf8fe04c3b967b4e5bfaf537aec

roles/deploy_user/tasks/main.yml
acozine: addressed in 5691a35087107bf8fe04c3b967b4e5bfaf537aec

sandbergja (approved):

PR #3381, opened 2022-10-06
add hc3 and hc3-test permanent redirects

kayiwa:
this will add new domain names under cdh


carolyncole (approved):
@kayiwa I'm unsure why the locations swapped positions, but I am assuming you did it for a reason.

PR #3385, opened 2022-10-10
New pudl certs

acozine:
Updates certs and keys for `pudl` and `pudltest`.

This is, in part, a test of whether or not we can use the combined certificate that comes back from OIT with their new process.


acozine:
Note: pudl has been decommissioned, but we are still redirecting the domain, so we need to maintain the certificates.

kayiwa (approved):
Thank you! OK!

PR #3389, opened 2022-10-12
renews SSL certs and keys for issue-tracker and ojs-dev

acozine:
Last two of the four test-cases for the new OIT process. So far so good!

All four are set to auto-renew in future.


kayiwa (approved):
Thanks! OK

PR #3392, opened 2022-10-13
create the ealapps role

kayiwa:
adds a deploy user with keys to be used to clone.
configures php and apache2


roles/ealapps/files/eal_app
```diff
@@ -0,0 +1,26 @@
+$ANSIBLE_VAULT;1.1;AES256
```
sandbergja, line 1: What does this private key do?  And can we add a `.key` to the filename?

roles/ealapps/molecule/default/verify.yml
```diff
@@ -0,0 +1,18 @@
...
+  gather_facts: false
+  tasks:
+  - name: check php curl status
+    ansible.builtin.package:
+      name: "{{ item }}"
+      state: present
+    check_mode: true
+    register: pkg_status
+    loop:
+      - php8.1-curl
```
sandbergja, line 13: Why are we just checking for php8.1-curl, and not the other php8.1 packages (like php8.1, php8.1-dev, php8.1-zip) here?

roles/ealapps/molecule/default/verify.yml
```diff
@@ -0,0 +1,18 @@
...
+
+  - name: test for ruby packages
```
sandbergja, line 15: What does this ruby packages test do?

roles/ealapps/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+    state: present
+
+- name: ealapps | install github deploy keys
+  ansible.builtin.template:
+    src: "{{ item }}"
+    dest: "/home/{{ deploy_user }}/.ssh/{{ item }}"
+    owner: "{{ deploy_user }}"
+    group: "{{ deploy_user }}"
+    mode: 0600
+  with_items: "{{ deploy_keys }}"
```
sandbergja, line 15: Where is the deploy_keys variable defined?

sandbergja (changes requested):
Thanks for your work on this, @kayiwa !  @christinach and I reviewed the first half and we'll continue tomorrow.

roles/ealapps/molecule/default/verify.yml
kayiwa: this come from the PHP role. https://github.com/pulibrary/princeton_ansible/blob/674f6dc4c41fcf91686fa30c03620340d1b1fe9a/roles/php/tasks/main.yml#L19 and I notice curl is already in there.

roles/ojs/templates/script.exp.j2
```diff
@@ -43,7 +43,7 @@ if {$force_conservative} {
 
 
 set timeout -1
-spawn php7.4 tools/install.php
+spawn php8.1 tools/install.php
```
christinach, line 46: @sandbergja was wondering if ojs should be protected at php7?

roles/php/tasks/configure-apache2.yml
```diff
@@ -16,18 +16,3 @@
...
-  notify: restart apache
-  when:
-    - running_on_server
-  ignore_errors: true
-
-- name: php | enable php apache module
-  apache2_module:
-    state: present
-    name: php{{ php_version }}
-  notify: restart {{ php_webserver }}
```
christinach, line 33: Why do we remove these tasks?

christinach (changes requested):
@kayiwa, @sandbergja and I added a couple more questions.

roles/php/tasks/configure-apache2.yml
kayiwa: ansible does the right thing and installs the module.

roles/ealapps/files/eal_app
kayiwa: allows us to clone a private repo.

roles/ealapps/tasks/main.yml
kayiwa: https://github.com/pulibrary/princeton_ansible/blob/eaf7664c9cb9d2cd3cac805498f4a203274ae744/roles/deploy_user/defaults/main.yml#L3

roles/ojs/templates/script.exp.j2
kayiwa: I got the blessing of RDSS to move on with this

kayiwa, line 46: can we take advantage of templates with "{{ php_version }}" for future us

playbooks/ealapps.yml
```diff
@@ -0,0 +1,19 @@
+---
+# by default this playbook runs in the staging environment
```
kayiwa, line 2: add this to the apache service

roles/drupal9/molecule/default/verify.yml
```diff
@@ -40,16 +40,25 @@
     register: file
     failed_when:
       - not file.stat.exists
+  - name: remove legacy php packages
```
kayiwa, line 43: this only allows it to pass on CI. What about the VMS

roles/ealapps/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
+=========
+
+A brief description of the role goes here.
+
```
kayiwa, line 5: add documentation about role

roles/ealapps/files/id_rsa.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC2RPcHrYslQj6AHe+AHai2HRgy8Z0qOo14xDbvO6g7yFJd2pfoG+UgbscgPwl8vRpD9TFuTwkcd9BE9hiBQW6oiVOdbADMAIobXK2AmNKN0vzHLs3UZRqvnq9loKUswxMKGUD/2sfCV+nSVpQYj8ZIwW+XmJinY7T2yttChXlWa3BLqTDt0T+A8vFbBYC03Kre278ypGDmuAkuvDpXI0QW9RXJxuyoxT39TaJmh0R0a66q9y2VC5ElOeheDyw9QP7hl7IYi2M3/GFO5mOjLzZNkcwJ+A29zxSXl0cnmZX6S9bxgohIOBinXhz3/g9DfB2wJfyBlGjiqE7Ut8M02P7E6Q/LaiQLK+St8oDTO0yFA/QsgvPgLXVynfiu1mjXix6to1O719jabwjs2XBFBD/zIoawMu3FAId8Y5Sqjm36LI1JZLdYLD+OxFJwlBauPwiIunrQ5sIEjxNAOfLC4jaj9i8+ol00ldi0lNVMXyZ6SKj+5Wbhm3Lw3A8/YPA5VNE= kayiwa@PU-F5KSY01XF9VN
```
kayiwa, line 1: we don't need both.

roles/ealapps/meta/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+galaxy_info:
+  role_name: ealapps
+  company: Princeton University Library
+  description: <Role Description>
```
kayiwa, line 5: add a description

roles/ealapps/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
+---
+# tasks file for ealapps
+- name: ealapps | install ealapps dependencies
+  ansible.builtin.apt:
+    name: ["php{{ php_version }}-curl"]
```
kayiwa, line 5: php role installs this already.

roles/ealapps/tests/inventory
```diff
@@ -0,0 +1,2 @@
+localhost
```
kayiwa, line 1: we don't need this from `molecule init role`

roles/ojs/templates/script.exp.j2
kayiwa, line 46: use "{{ php_version }}"

christinach (changes requested):
@kayiwa we need to install `php8.1-gd` and `php8.1-mysql` package in the drupal role. We tested this with @sandbergja and @kevinreiss.

christinach (changes requested):
@kayiwa we would also like to install `php8.1-mbstring` for the drupal role. Thanks!

sandbergja:
@kayiwa @christinach I think these lines can just be updated to reference {{ php_version }}, rather than hard-coded php7.4: https://github.com/pulibrary/princeton_ansible/blob/main/roles/drupal/tasks/main.yml#L7-L10

That will also bring in the `php8.1-xml` package, which is needed for the special_collections site.

roles/libwww/molecule/default/verify.yml
```diff
@@ -1,8 +1,9 @@
 ---
 # This is an example playbook to execute Ansible tests.
-
 - name: Verify
   hosts: all
+  vars:
+    - php_version: "8.1"
```
kayiwa, line 6: change this back to 7.4

playbooks/ealapps.yml
```diff
@@ -0,0 +1,19 @@
...
+# to run in production, pass '-e runtime_env=production'
+- name: build the ealapps
+  hosts: ealapps_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/ealapps/{{ runtime_env | default('staging') }}.yml
+    - ../group_vars/ealapps/vault.yml
```
sandbergja, line 10: This file isn't yet in source control, so the playbook fails with:

```
PLAY [build the ealapps] ***********************************************************************************************************************************
ERROR! vars file ../group_vars/ealapps/vault.yml was not found
Could not find file on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

sandbergja (approved):
Thanks so much, @kayiwa

christinach (approved):
Thanks @kayiwa !

PR #3395, opened 2022-10-14
Added key and env value

rladdusaw:
Resolves pulibrary/dss#283

### DIFF
--- a/group_vars/dss/staging.yml
+++ b/group_vars/dss/staging.yml
@@ -23,3 +23,5 @@ dss_db_user: "dss"
 dssdb_password: "{{ vault_dss_db_password }}"
 dss_host_name: "dss-staging1.princeton.edu"
 solr_url: "http://lib-solr8-staging.princeton.edu:8983/solr/dss-staging"
+dss_honeybadger_key: '{{ vault_dss_honeybadger_key }}'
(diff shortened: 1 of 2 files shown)


group_vars/dss/staging.yml
```diff
@@ -23,3 +23,5 @@ dss_db_user: "dss"
 dssdb_password: "{{ vault_dss_db_password }}"
 dss_host_name: "dss-staging1.princeton.edu"
 solr_url: "http://lib-solr8-staging.princeton.edu:8983/solr/dss-staging"
+dss_honeybadger_key: '{{ vault_dss_honeybadger_key }}'
+honeybadger_env: 'staging'
```
sandbergja, line 27: The HONEYBADGER_ENV environment variable didn't get set on dss-staging when this playbook was run.  I think it might need to be added to the rails_app_vars in roles/dss/vars/main.yml

sandbergja (changes requested):
Thanks, @rladdusaw !  The HONEYBADGER_API_KEY envvar got set correctly when this playbook was run, but the HONEYBADGER_ENV wasn't.

sandbergja (approved):
Thanks @rladdusaw !

PR #3407, opened 2022-10-21
renew certificates for

kayiwa:
milberg (now part of lib-static-prod SAN)
pulsearch (now part of catalog SAN)
point configuration of of catalog and lib-static-prod to reflect changes
above


acozine:
Looks good, thanks @kayiwa .

PR #3414, opened 2022-11-02
updates ruby_s role to support upgrades, ease idempotency

acozine:
Co-authored-by: Christina Chortaria <christinach@users.noreply.github.com>

Closes #3317.

This PR uses a single commit for a bunch of work that was previously done on #3373. We are disentangling the changes to the `ruby_s` role from the changes to bibdata, so we can upgrade ruby on other projects from the `main` branch.

Changes include:
- removes all apt packages that begin with `ruby`, so we don't have to know what version was installed by apt
- adds a check for the existing ruby version
- only installs from source when the existing ruby version is different from the desired ruby version
- moves the shell tasks (for building ruby from source) into the conditional tasks file so we can remove `changed_when: false` and still pass idempotency testing
- adds a make uninstall step so the role can upgrade/downgrade ruby when ruby was previously installed from source

### DIFF
--- a/roles/ruby_s/tasks/cleanup.yml
+++ b/roles/ruby_s/tasks/cleanup.yml
@@ -3,11 +3,15 @@
   community.general.gem:
     name: bundler
     state: absent
-  changed_when: false
 
-- name: ruby | remove installed ruby
+- name: ruby | check for apt-installed rubies
+  ansible.builtin.package_facts:
+    manager: apt
+  register: all_packages
+
+- name: ruby | remove any package that begins with 'ruby'
   ansible.builtin.apt:
-    name: ["{{ ubuntu_ruby_version }}", "{{ ubuntu_ruby_version }}-dev", "ruby-switch"]
+    name: "{{ ansible_facts.packages | list | select('match', '^ruby') }}"
     state: absent
 
 - name: ruby | remove the Brightbox repository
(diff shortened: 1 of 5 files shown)


christinach (approved):
We have tested this pr in bibdata-staging and bibdata-qa

christinach:
I meant to add this comment in this pr https://github.com/pulibrary/princeton_ansible/pull/3373#issuecomment-1301065281

acozine:
The ruby_s role does not reliably install the specified version of `bundler`, we're not sure why.

hackartisan:
I think the `closes` reference is pointing to the wrong number

roles/ruby_s/tasks/cleanup.yml
```diff
@@ -1,25 +1,28 @@
...
   community.general.gem:
     name: bundler
     state: absent
-  changed_when: false
 
-- name: ruby | remove installed ruby
+- name: ruby_s | check for apt-installed rubies
+  ansible.builtin.package_facts:
+    manager: apt
+  register: all_packages
```
hackartisan, line 10: It doesn't look like we're using this value? Is this to aid debugging? Maybe it should be removed, or commented.

acozine: the value gets used in line 14

acozine: oh, I see what you mean, we don't need to register the var because it becomes part of ansible_facts . . . I'll make an issue for this

PR #3417, opened 2022-11-03
Fix permissions issue for OJS file uploads

bess:
Fixes https://github.com/pulibrary/openpublishing/issues/56

Screen Shot 2022-11-03 at 1 37 36 PM
Screen Shot 2022-11-03 at 11 09 59 AM
Screen Shot 2022-11-03 at 1 51 35 PM


carolyncole:
https://github.com/pulibrary/princeton_ansible/pull/3419 May be a better way if it passes idempotence

PR #3419, opened 2022-11-04
Fix ojs permissions2

carolyncole:
Maybe this will work and be less mysterious


carolyncole:
Closing the other commit was merged.  May put this up as a separate commit.

PR #3428, opened 2022-11-10
Make the geoserver playbook non-destructive

### LINKED ISSUE #214: GeoServer playbook can't be re-run
- There is a problem with running deploy_user role. Deploy user already exists.
- Samba role can't be re-run. Shares already exist.
- I'm concerned that deploying the geoserver role might overwrite the geoserver app directory.

acozine:
Closes #214.

Makes the download step explicit.
Checks the version of the download.
Moves the install tasks into a separate tasks file.
Skips the install tasks unless the playbook downloads a different version of geoserver.

I'm making this a draft PR because I'm not sure which tasks were destructive. But this PR will address the obvious challenges I see in the role and playbook.

### DIFF
--- a/playbooks/geoserver_production.yml
+++ b/playbooks/geoserver_production.yml
@@ -1,5 +1,6 @@
 ---
-- hosts: geoserver
+- name: install geoserver
+  hosts: geoserver
   remote_user: pulsys
   become: true
   vars_files:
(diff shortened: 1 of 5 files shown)


acozine:
Closing as obsolete. We removed the geoserver playbook in #4335.

Issue #3356, opened 2022-09-21
Upgrade Discoveryutils to PHP 8.1


kevinreiss:

kevinreiss:
This app is part of the libwww role.

kevinreiss:
1. Run locally with PHP 8.1 with Lando
2. Run phpunit test suite. 
3. In tandem with upgrade of the main library website deploy to staging
4.  In tandem with upgrade of the main library website deploy to prod

christinach:
Can we test this in the library-staging website?

kevinreiss:
We will run this on php 8.0 since the main website which shares a server with this app requires it.

Issue #3350, opened 2022-09-21
Upgrade lib_www role to PHP 8.0


kevinreiss:

kevinreiss:
1. Run locally with PHP 8.1 - make sure site comes up and admin backend works
2. Refresh cache locally and the drupal cron task
3. Check the active "feeds" importers for libcal and staff directory
4. Update ansible role to use php 8.1
5. Deploy to staging
6. Deploy to prod

kevinreiss:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DACS-Maintenance-5da0db919c4ddf0001b2eb92/issues/pulibrary/princeton_ansible/3350?planning-poker) with Zenhub @sandbergja @maxkadel @leefaisonr

kevinreiss:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DACS-Maintenance-5da0db919c4ddf0001b2eb92/issues/pulibrary/princeton_ansible/3350?planning-poker) with Zenhub @christinach @regineheberlein @rladdusaw

sandbergja:
@maxkadel and I investigated this at length today.  Unfortunately, the Drupal 7 date_ical module relies on a very outdated version of the iCalcreator library, which is not compatible with PHP 8.

For context:

- The [date_ical module documentation](https://git.drupalcode.org/project/date_ical/-/blob/7.x-3.x/README.txt) says: "The author of iCalcreator made extenside backwards incompatible changes to the
library in the v2.22 release. Thus Date iCal does not support any version of
iCalcreator after v2.20.x."
- PHP 8 support is not mentioned in the iCalcreator release notes until https://github.com/iCalcreator/iCalcreator/blob/master/releaseNotes#L132-L155

The results on PHP 8 are:

- With iCalcreator 2.20.2 (the version currently in main), the libcal importer fails with an error about the curly brace array syntax (e.g. `$array{0}`)
- With iCalcreator 2.20.4, the libcal importer hangs for several minutes, before giving a database timeout error

kevinreiss:
We will run this on PHP 8.0 because this site will also struggle with 8.1 

see
https://github.com/pulibrary/special_collections/issues/726
https://github.com/pulibrary/special_collections/issues/725

PR #3435, opened 2022-11-15
Use php 8.0 for libwww

sandbergja:
Related to #3350 and #3356

### DIFF
--- a/roles/libwww/vars/main.yml
+++ b/roles/libwww/vars/main.yml
@@ -1,6 +1,5 @@
 ---
 # vars file for roles/libwww
-php_version: "7.2"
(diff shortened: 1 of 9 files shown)


roles/libwww/vars/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # vars file for roles/libwww
-php_version: "7.2"
+php_version: "8.0"
```
christinach, line 3: We have php_version: "8.0" both here and in group_vars. Do we need it in both places?

kayiwa: it will fail CI if you don't add it there.

PR #3437, opened 2022-11-16
add prosody-beta config add a new prosody-beta configuration

kayiwa:


roles/nginxplus/files/conf/http/cdh_prod_prosody.conf
```diff
@@ -0,0 +1,88 @@
...
+}
+
+server {
+    listen 80;
+    server_name prosody-beta.princeton.edu;
+
+    location / {
+        return 301 https://$server_name$request_uri;
+    }
+}
```
rlskoeser, line 29: do we need to listen to `ppa.princeton.edu` as well? would redirect to `prosody` as the canonical (but don't need this part yet)

rlskoeser (approved):

Issue #3337, opened 2022-09-20
Rename nginxplus production playbook


kayiwa:
remove the nginxplus production from playbooks

acozine:
Make sure the playbook can run in any environment (now that we are creating a staging system), but call it `nginxplus.yml` instead of `nginxplus_production.yml`.

aruiz1789:
List of configs that will be migrated to dev load balancers: 

- [ ] allsearch-api_staging.conf
- [ ] allsearch_staging.conf
- [ ] approvals_staging.conf
- [ ] datacommons_staging.conf
- [ ] dpul-staging.conf
- [ ] dss-staging.conf
- [ ] figgy-staging.conf
- [ ] imagecat_staging.conf
- [ ] lae-staging.conf
- [ ] lib-jobs-staging.conf
- [ ] lib-solr8d-staging.conf
- [ ] lib-solr9-staging.conf
- [ ] lockers-and-study-spaces-staging.conf
- [ ] maps-staging.conf
- [ ] oawaiver-staging.conf
- [ ] orcid_staging.conf
- [ ] pdc-describe_staging.conf
- [ ] pdc-discovery_staging.conf
- [ ] pulfalight-staging.conf
- [ ] repec-staging.conf
- [ ] slavery-staging.conf
- [ ] static_tables_staging.conf
- [ ] whichiso_staging.conf

aruiz1789:
```
ruiza@LIB-FMC6NRWR1L ~ % nslookup allsearch-api-staging.princeton.edu | grep Name:
Name:	adc-dev.lib.princeton.edu

ruiza@LIB-FMC6NRWR1L ~ % nslookup allsearch-staging.princeton.edu | grep Name:
Name:	adc-dev.lib.princeton.edu

ruiza@LIB-FMC6NRWR1L ~ % nslookup approvals-staging.princeton.edu | grep Name:
Name:	adc-dev.lib.princeton.edu

ruiza@LIB-FMC6NRWR1L ~ % nslookup datacommons-staging.princeton.edu | grep Name:
Name:	adc-dev.lib.princeton.edu

ruiza@LIB-FMC6NRWR1L ~ % nslookup whichiso-staging.princeton.edu | grep Name:
Name:	adc-dev.lib.princeton.edu
```

aruiz1789:
```
ruiza@LIB-FMC6NRWR1L ~ % nslookup dss-staging.princeton.edu | grep Name:
[...]

Issue #2556, opened 2021-11-04
Nginx playbook errors on new server


carolyncole:
need to manually run: `sudo mkdir /data/nginx/lockers-and-study-spaces-staging`

Sure would be nice not to have to do that on each server...

```
TASK [../roles/nginxplus : nginxplus | make sure nginx configs are valid] **************************************************************
fatal: [lib-adc2.princeton.edu]: FAILED! => {"changed": false, "cmd": ["/usr/sbin/nginx", "-t"], "delta": "0:00:00.257307", "end": "2021-11-04 11:11:12.533545", "msg": "non-zero return code", "rc": 1, "start": "2021-11-04 11:11:12.276238", "stderr": "nginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size\nnginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size\nnginx: the configuration file /etc/nginx/nginx.conf syntax is ok\nnginx: [emerg] mkdir() \"/data/nginx/lockers-and-study-spaces-staging/NGINX_cache/\" failed (2: No such file or directory)\nnginx: configuration file /etc/nginx/nginx.conf test failed", "stderr_lines": ["nginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size", "nginx: [warn] could not build optimal server_names_hash, you should increase either server_names_hash_max_size: 512 or server_names_hash_bucket_size: 64; ignoring server_names_hash_bucket_size", "nginx: the configuration file /etc/nginx/nginx.conf syntax is ok", "nginx: [emerg] mkdir() \"/data/nginx/lockers-and-study-spaces-staging/NGINX_cache/\" failed (2: No such file or directory)", "nginx: configuration file /etc/nginx/nginx.conf test failed"], "stdout": "", "stdout_lines": []}
```

acozine:
Looks like this is an issue as we add more sites - see https://serverfault.com/questions/419847/nginx-setting-server-names-hash-max-size-and-server-names-hash-bucket-size and https://stackoverflow.com/questions/13895933/nginx-emerg-could-not-build-the-server-names-hash-you-should-increase-server. 

[...]

PR #3439, opened 2022-11-17
initial installation for staging load balancer

kayiwa:


acozine:
This will be a good test of whether running the `nginx_production_rebuild.yml` playbook (which we should rename, see #3337) creates all the necessary cache directories on the staging load balancers. See [this comment](https://github.com/pulibrary/princeton_ansible/issues/2556#issuecomment-1367433679).

kayiwa:
superseded by #3638

PR #3446, opened 2022-11-22
add the procs tool

kayiwa:
the [procs](https://github.com/dalance/procs) tool allows us to see what
processes are running in a way that improves on the venerable `ps`
we alphabetize the packages


christinach (approved):
Thank you @kayiwa !

PR #3453, opened 2022-11-24
Add fqrns for sneakers_worker role

leefaisonr:


leefaisonr:
This role was a bit different from kakadu...can we review sometime (I understand if it can not be this week due to training)

kayiwa (approved):
thanks! OK

roles/sneakers_worker/meta/main.yml
```diff
@@ -7,13 +7,13 @@ galaxy_info:
...
 
-  min_ansible_version: 2.4.0.0
+  min_ansible_version: "2.4.0.0"
 
   platforms:
     - name: Ubuntu
       versions:
-        - 16.04
-        - 18.04
+        - xenial
```
acozine, line 15: Xenial is EOL, so we can remove this line. It looks like this is also the source of the "trailing whitespace" error in CI.

PR #3454, opened 2022-11-28
modify template for connection params

kayiwa:


christinach (changes requested):
There are conflicts. @kayiwa

PR #3458, opened 2022-11-28
the ruby_s role occasionally fails idempotence

### LINKED ISSUE #3455: ruby_s idempotence
the `ruby_s` role is broken in idempotence

kayiwa:
the role will occasionally run longer than the alloted 10 minutes and
will subsequently fail idempotence.

I am unable to find the pattern for this and certainly not on CI.

Connected to #3455.


acozine:
It's possible the role will continue to fail occasionally - when there's a change caught by a cache update. But let's try this and then track failures (or lack thereof).

PR #3461, opened 2022-11-28
add videoreserves-staging name

kayiwa:
the name videoreserves is registered and the TLS certificates added.


sandbergja (approved):
Thanks, @kayiwa !

PR #3490, opened 2022-12-02
Use more generalized variable name for unwanted versions of PHP, ensure they are uninstalled

### LINKED ISSUE #3489: Ensure unwanted PHP versions are uninstalled
See https://github.com/pulibrary/princeton_ansible/issues/3486 for some context. 

When downgrading PHP versions (e.g. from 8.1 to 8.0), the application is still looking at the newest version of PHP even after running the playbook. Once the playbook is run, the application should point to the version of PHP in the variables.

maxkadel:
Use a wildcard rather than list of packages.

Closes #3489


acozine:
Current failure is:
```
   TASK [php : php | remove packages that begin with 'php7.4'] ********************
  fatal: [instance]: FAILED! => {"changed": false, "msg": "No package(s) matching 'php7.4*' available"}
```

maxkadel:
> Current failure is:
> 
> ```
>    TASK [php : php | remove packages that begin with 'php7.4'] ********************
>   fatal: [instance]: FAILED! => {"changed": false, "msg": "No package(s) matching 'php7.4*' available"}
> ```

I'm trying to look at this and I'm confused - if there are no packages on the machine that match `php7.4*`, then shouldn't ansible just respond "ok", since the state of the machine is what we want? I feel like there's something I'm not understanding about this. 

Trying adding `fail_on_autoremove: false`, even though that's supposed to be the default

maxkadel:
Still has the same problem - 

```
 TASK [php : php | remove packages that begin with 'php7.4'] ********************
 fatal: [instance]: FAILED! => {"changed": false, "msg": "No package(s) matching 'php7.4*' available"}
```

maxkadel:
May be related to this issue - https://github.com/ansible/ansible/issues/62262

acozine:
Nice sleuthing @maxkadel! This is an excellent step forward.

PR #3492, opened 2022-12-02
allow PHP*-xml version for OJS to change

acozine:
With `php7.4-xml` hard-coded, the OJS role fails idempotency.

Since the default version of PHP for all other related modules is already 8.1, this PR updates the XML package to match (and also to handle future upgrades).

Related to #3490, where the failure first showed up.

### DIFF
--- a/roles/ojs/tasks/install_prerequisites.yml
+++ b/roles/ojs/tasks/install_prerequisites.yml
@@ -11,7 +11,7 @@
 
 - name: ojs | install php xml libraries
   apt:
-    name: ["php7.4-xml"]
+    name: ["php{{ php_version }}-xml"]
     state: present
 
 - name: ojs | install php postgres drivers


acozine:
The OJS role  idempotency check fails because the expect script runs again on the second pass. I get the same behavior locally. If I run `molecule converge` on my laptop, the expect script runs:
```
TASK [ojs : ojs | install ojs] *************************************************
included: /Users/alicia/git_repos/princeton_ansible/roles/ojs/tasks/install_ojs.yml for instance

TASK [ojs : ojs | copy config file] ********************************************
ok: [instance]

TASK [ojs : ojs | write expect script for OJS interactive installation] ********
changed: [instance]

TASK [ojs : ojs | run expect script for OJS interactive installation] **********
changed: [instance]

TASK [ojs : ojs | remove expect script] ****************************************
changed: [instance]
```
and the playbook succeeds overall:
```
PLAY RECAP *********************************************************************
instance                   : ok=82   changed=33   unreachable=0    failed=0    skipped=110  rescued=0    ignored=0

```
However, if I log into the test instance, the test for "Is OJS installed" fails:
```
root@instance:/var/www/ojs-3.3.0-8# grep 'installed =' config.inc.php
installed = Off
```
and the idempotence test fails with the same error as in CI.

bess:
This passes idempotence for me locally if I add all these `changed_when: false` statements.

bess (approved):

PR #3494, opened 2022-12-05
draft playbook for capistrano deploys

acozine:
We want to be able to deploy code from Tower.

First attempt, very much a draft.


acozine:
Per discussion today, we will not use Ansistrano, but will run existing cap tasks directly.

acozine:
Turns out a playbook has to be on the main branch before you can create a Template in Tower based on it. So I'm going to merge this tonight so we have a template to start next steps on tomorrow.

PR #3502, opened 2022-12-06
update ojs-staging TLS to include ojs-staging2

leefaisonr:
Co-authored-by: Francis Kayiwa <kayiwa@users.noreply.github.com>


acozine:
Need to re-do the private keys.

PR #3509, opened 2022-12-07
updates rapid7 after testing on sandbox vms

VickieKarasic:
We installed Rapid7 on a sandbox machine and tried rebooting it, and the Rapid7 agent restarted successfully. Removing the dummy tasks in the playbook.


kayiwa (approved):
Thanks! OK

Issue #3508, opened 2022-12-07
Use MySQL database for the slavery-dev1 environment


kevinreiss:
- [ ] Create a mysql database for the slavery-dev1 environment
- [ ] Change the playbook to set-up slavery-dev1 to use Mysql. 
- [ ] After updated playbook is run. Import the mysql dump file provided by contractor. Currently in /home/deploy on slavery-dev1.

PR #3510, opened 2022-12-07
Change pas-dev to run on mysql not pgsql

maxkadel:
Connected to #3508


roles/pas/defaults/main.yml
```diff
@@ -21,6 +21,7 @@ pas_db_schema: 'public'
 pas_db_table_prefix: 'craft_'
 pas_security_key: 'abc123'
 pas_default_site_url: 'https://localhost:8484'
+craft_version: 3
```
kevinreiss, line 24: The new version we are trying is Craft 4.

maxkadel: Would you like the default to be Craft 4, since that's what we're moving towards? We wanted it to be backwards compatible for production and staging, since we don't have Craft 4 working yet.

sandbergja (approved):
Thanks, @maxkadel !

PR #3512, opened 2022-12-07
Ealapps cron jobs

tventimi:
Uses the ansible.builtin.cron module to create cron jobs for ealapps.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/cron_module.html

All tests passed.  I want to confirm that the files were generated correctly before finalizing this (will do so first thing on Thu 12/8).


tventimi:
Confirmed that the files hishi/hishi.json and newtitles/files/newtitles.cron.xml were correctly generated by the overnight cron job.

maxkadel (approved):

PR #3521, opened 2022-12-12
update zookeeper fqrn parameters

leefaisonr:


kayiwa (changes requested):
the handlers/main.yml needs to have `ansible.builtin.service` 

https://github.com/pulibrary/princeton_ansible/blob/9a9efcb07a732283bbbb3cd0b40531f49c19f740/roles/zookeeper/handlers/main.yml#L3

kayiwa (approved):
thank you! OK

PR #3523, opened 2022-12-12
[Repec] upgrade ruby

sandbergja:
Also refactor playbook/group_vars to use the `runtime_env` and `common.yml` patterns.

Sibling PR: https://github.com/pulibrary/repecwp/pull/81


playbooks/repec.yml
```diff
@@ -1,9 +1,12 @@
...
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- hosts: repec_{{ runtime_env | default('staging') }}
   remote_user: pulsys
   become: true
   vars_files:
-    - ../group_vars/repec/repec_prod.yml
+    - ../group_vars/repec/repec_{{ runtime_env | default('staging') }}.yml
+    - ../group_vars/repec/common.yml
     - ../group_vars/repec/vault.yml
```
maxkadel, line 10: Should this playbook include the datadog role for production? Is it a conscious choice to leave it off?

maxkadel, line 10: Is there something that is supposed to happen with the assets repository or something after you run this playbook? Would that be good to add to the post_tasks?

maxkadel (approved):
Thanks for the upgrade and refactor!

Approved, but just double check me on the comments.

playbooks/repec.yml
sandbergja: Good idea!  I added the datadog role.

sandbergja: Bundler will handle pulling in the pul_assets repo.  The only thing to do is deploy the app, which doesn't seem worth a special `post_tasks` reminder to me.

maxkadel (approved):
Still approved! Thanks Jane!

PR #3536, opened 2022-12-14
cleanup stale ruby variables(unused) from qa and production.

christinach:
Move ruby vars to group_vars/common.yml


kayiwa (approved):
Thanks! OK!

PR #3539, opened 2022-12-15
ojs tls certs

kayiwa:
- add ojs-staging tls certs
- add ojs-prod tls certs


christinach (approved):

roles/nginxplus/files/conf/http/ojs-staging.conf
```diff
@@ -4,6 +4,7 @@ proxy_cache_path /data/nginx/ojs-staging/NGINX_cache/ keys_zone=ojs-stagingcache
 upstream ojs-staging1 {
     zone ojs-staging 64k;
     server ojs-staging1.princeton.edu:443;
+    #server ojs-staging1.princeton.edu:443;
```
acozine, line 7: Should this be ojs-staging2?

Also, should it be uncommented?

roles/nginxplus/files/conf/http/openpublishing.conf
```diff
@@ -4,6 +4,7 @@ proxy_cache_path /data/nginx/ojs-prod/NGINX_cache/ keys_zone=ojs-prodcache:10m;
 upstream ojs-prod1 {
     zone ojs-prod 64k;
     server ojs-prod1.princeton.edu:443;
+    #server ojs-prod2.princeton.edu:443;
```
acozine, line 7: Should this be uncommented?

acozine (commented):
Just a couple of questions.

roles/nginxplus/files/conf/http/ojs-staging.conf
kayiwa: It should be ojs-staging2 and it is commented because this will be deployed as A/B

roles/nginxplus/files/conf/http/openpublishing.conf
kayiwa: no it shouldn't. Only one end point is active.

PR #3540, opened 2022-12-15
Set memory limit 512M in apache2/php.ini

### LINKED ISSUE #3529: Maintain apache2 memory settings when upgrading PHP
After the upgrade of the Friends of PUL website, we saw problems loading the [event photos](https://fpul.princeton.edu/events/photos) page. A first look revealed a memory issue. This turned out to be caused (at least in part) by the `memory_limit` setting in `/etc/php/8.1/apache2/php.ini`.

The old configuration for PHP 7.4 set `memory_limit = 256M`, while the new 8.1 configuration set `memory_limit = 128M`. We manually updated the setting on the FPUL production VM to `memory_limit = 512M`, but we should fix this in the role and be aware of it on other upgrades.

christinach:
closes #3529

### DIFF
--- a/group_vars/friends_of_pul/common.yml
+++ b/group_vars/friends_of_pul/common.yml
@@ -2,8 +2,10 @@
 drupal_major_version: 7
 drupal_db_port: '3306'
 db_port: "{{ drupal_db_port }}"
-mysql_server: false
 db_is_mysql: true
+friends_pul_memory_limit: '512M'
+memory_limit: {{ friends_pul_memory_limit }}
+mysql_server: false
 mysql_host: "{{ db_host }}"
 mysql_databases:
     - name: "{{ drupal_db_name }}"
--- a/group_vars/friends_of_pul/common.yml
+++ b/group_vars/friends_of_pul/common.yml
@@ -4,7 +4,7 @@ drupal_db_port: '3306'
 db_port: "{{ drupal_db_port }}"
 db_is_mysql: true
 friends_pul_memory_limit: '512M'
-memory_limit: {{ friends_pul_memory_limit }}
+php_memory_limit: {{ friends_pul_memory_limit }}
 mysql_server: false
 mysql_host: "{{ db_host }}"
 mysql_databases:
--- a/group_vars/libwww/common.yml
+++ b/group_vars/libwww/common.yml
@@ -10,6 +10,8 @@ drupal_base_dir: '/var/www'
 drupal_project: 'library_cap'
 
 ddtrace_repo_version: '0.56.0'
+libwww_memory_limit: '256M'
+php_memory_limit: {{ libwww_memory_limit }}
 
 mysql_host: "{{ db_host }}"
 mysql_root_password: "{{ root_db_password }}"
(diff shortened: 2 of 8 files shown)


roles/friends_of_pul/tasks/main.yml
```diff
@@ -1,2 +1,7 @@
 ---
 # tasks file for roles/friends_of_pul
+- name: friends_of_pul | ensure memory limit is set in apache2 php.ini
```
acozine, line 3: Could we do this in the PHP role instead? That way it will be available to all PHP applications, not just FPUL.

I'd put the task here: https://github.com/pulibrary/princeton_ansible/blob/main/roles/php/tasks/mod-php.yml, making the memory limit a variable so each PHP project can select its own level. Then I'd add `512M` to the FPUL group_vars and set a default value in the PHP role.

acozine (commented):
Thanks for taking this on @christinach! I have a couple of suggestions for making the fix more general.

roles/php/tasks/mod-php.yml
```diff
@@ -4,3 +4,9 @@
     dest: "/etc/php/{{ php_version }}/apache2/php.ini"
     regexp: "date.timezone ="
     line: "date.timezone = {{ server.timezone }}"
+
+- name: php | ensure memory limit is set in apache2 php.ini
+  ansible.builtin.lineinfile:
+    dest: "/etc/php/{{ php_version }}/apache2/php.ini"
+    regexp: "memory_limit ="
+    line: "memory_limit = {{ memory_limit }}"
```
acozine, line 12: Per discussion in Ansible Open House, maybe use `php_memory_limit` as the var, so the association is clear.

group_vars/friends_of_pul/common.yml
```diff
@@ -2,8 +2,10 @@
 drupal_major_version: 7
 drupal_db_port: '3306'
 db_port: "{{ drupal_db_port }}"
-mysql_server: false
 db_is_mysql: true
+friends_pul_memory_limit: '512M'
+memory_limit: {{ friends_pul_memory_limit }}
```
acozine, line 7: Use `php_memory_limit` here?

group_vars/libwww/common.yml
```diff
@@ -10,6 +10,8 @@ drupal_base_dir: '/var/www'
 drupal_project: 'library_cap'
 
 ddtrace_repo_version: '0.56.0'
+libwww_memory_limit: '256M'
+memory_limit: {{ libwww_memory_limit }}
```
acozine, line 14: and here

group_vars/pas/common.yml
```diff
@@ -8,5 +8,8 @@ apache:
     servername: '{{ ansible_hostname }}'
     directory_options: '+Multiviews'
 
+pas_memory_limit: '256M'
+memory_limit: {{ pas_memory_limit }}
```
acozine, line 12: and here

roles/php/defaults/main.yml
```diff
@@ -5,6 +5,7 @@ ddtrace_repo_version: '0.44.1'
 datadog_enabled: false
 datadog_service_name: test
 datadog_environment: none
+memory_limit: '256M'
```
acozine, line 8: and here

maxkadel (approved):
Thanks for this, it's a big improvement! 🎉

acozine (approved):

PR #3545, opened 2022-12-16
Cache python environment (including pip dependencies) in github actions

sandbergja:
This shaves a solid 90 seconds off of each job run (except the very first one to run with a given python version or requirements.txt, of course).  Hopefully this will translate into more available runners on busy days with a lot of PRs. 🤞


kayiwa (approved):
WHOA! Thanks! OK

PR #3555, opened 2022-12-19
update how much of journal logs we retain

kayiwa:
we are restricting the caps to 100MB and keeping no more than 7 days
worth of journal logs


christinach (approved):

acozine (commented):
I like this approach. Could we make the two settings variables, and set them in `group_vars`, so different projects can select different settings if they want to, and so we can change the limits whenever we want?

acozine:
Per discussion in Ansible Open House, we would like to standardize on these numbers. Some logs were growing into gigabytes. If we see a need to keep larger logs, or keep them longer, we can adjust in future.

PR #3559, opened 2022-12-20
[geaccirc] Upgrade to Ruby 3.1

maxkadel:
- successfully run on geaccirc-staging
- Changing order of vars_files so precedence is given to environment-specific file

Connected to pulibrary/geaccirc#16


kayiwa (approved):
Thanks! OK

PR #3561, opened 2022-12-20
add qa vms to inventory

kayiwa:
the qa vms needed to be included in our environments for `Patch Tuesday` reasons


acozine:
We'll run patches on `runtime_env=qa` as well as `staging`.

PR #3576, opened 2022-12-22
also remove the config that refers to lib-prod cert and key

acozine:
Follow up to #3568.

Removes all reference to `lib-prod.princeton.edu`.


sandbergja (approved):
Thanks, @acozine ! 🐎

PR #3580, opened 2022-12-28
Configure sitemap cacheing in nginx for orangelight

### LINKED ISSUE #3436: Cache orangelight requests under /sitemap
### Current behavior

The sitemaps can take quite some time to generate (we were seeing up to 7 minutes!).  I don't think they are cached at the nginx level, so orangelight has to do its work every time they are requested.

### Desired behavior

Nginx should cache requests like this:
* https://catalog.princeton.edu/sitemap
* everything of pattern https://catalog.princeton.edu/sitemap/1234

And then expires the cache daily.

### Related work

We put [this fix](https://github.com/pulibrary/orangelight/issues/3253) into place, which reduces the size of each individual sitemap, so that they are quicker for orangelight to generate.

acozine:
Closes #3436.

WIP. Configures cacheing for the sitemap for the catalog/orangelight.

### DIFF
--- a/roles/nginxplus/defaults/main.yml
+++ b/roles/nginxplus/defaults/main.yml
@@ -156,7 +156,7 @@ nginx_main_template:
   http_enable: true
   http_settings:
     keepalive_timeout: 65
-    cache: false
+    cache: true
     rate_limit: false
     keyval: false
   stream_enable: false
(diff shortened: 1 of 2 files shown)


acozine:
This seems like the wrong approach. Needs more research.

PR #3581, opened 2022-12-29
removes campus_access directory from bibdata build

### LINKED ISSUE #3174: [bibdata] Bibdata should not set up the campus_access directory anymore
This is an old feature that is no longer needed

acozine:
Closes #3174.

Removes unused `campus_access` directory from the bibdata tasks and variables.

Running the playbook with this change will not remove the directory from existing infrastructure, but it will stop checking it and will not recreate it if/when it's removed.


sandbergja (approved):
Thanks, @acozine !

PR #3582, opened 2022-12-29
removes unused 'solr_version' variable

### LINKED ISSUE #2708: solr_version is wrong in many roles
Many roles specify a solr version of 8.2, but the actual solr version is 8.4.1
I'm not sure if this has any impact...
https://github.com/pulibrary/princeton_ansible/search?q=solr_version

acozine:
Closes #2708.

Grepping through the repo turned up these places where `solr_version` was being set (often incorrectly) but no place where it was being used (in other words, no instances of `{{ solr_version }}`). 

This PR removes the unused variables.


hackartisan (approved):
Thanks for finding this!

PR #3584, opened 2023-01-03
add the number of servers this runs

kayiwa:
this will bump up the time it takes to run all the updates with a
minimal disruptive penalty.
An odd number is also preferable because occasionally the playbook fails
when paired servers are included.


leefaisonr (approved):
Looks good, thanks!

PR #3599, opened 2023-01-05
Documentation for rekeying the vault

sandbergja:
Co-authored-by: Anna Headley <hackartisan@users.noreply.github.com>
Co-authored-by: Carolyn Cole <carolyncole@users.noreply.github.com>
Co-authored-by: Francis Kayiwa <kayiwa@users.noreply.github.com>
Co-authored-by: Robert-Anthony Lee-Faison <leefaisonr@users.noreply.github.com>

PR #3601, opened 2023-01-06
Remove old documentation for rekeying the vault

sandbergja:
new docs were introduced in #3599


kayiwa (approved):
Thanks @sandbergja OK!

PR #3603, opened 2023-01-06
patch lib-monitor

kayiwa:
add the lib-monitor server to patch Tuesday group


acozine:
Per discussion today we'll include more of the `orphan` groups. See https://github.com/pulibrary/princeton_ansible/blob/main/inventory/all_projects/_orphans.

inventory/by_environment/production
```diff
@@ -48,8 +50,10 @@ repec_production
 researchdata_production
 sftp_production
 shxco_production
+smb_serve
 solr8cloud_production
 solr_mirror_production
 special_collections_production
+studio_processors
```
acozine, line 57: Can we update the comment about `studio_processors` and `studio_proc` in the orphans file to mention which one we're using here? That way if we ever remove one of the two groups, we have a better chance of remembering to make sure we update this file if necessary.

acozine (commented):
Great catch, thanks @kayiwa!

I see one way to help Future Us here, otherwise looks good.

Issue #3328, opened 2022-09-13
can we mirror geoserver?


kayiwa:
can we mirror geoserver to prevent our CI from failing
add it to solr-mirror

acozine:
Related to #3292.

acozine:
Closing in favor of #3292, which has more specific info.

PR #3608, opened 2023-01-10
make the nodejs role use fqrn

kayiwa:


acozine:
Failure is from an outage on NodeJS downloads. We will mirror Node going forward to avoid this problem. Tangentially related to #3328.

sandbergja (approved):
Thanks, @kayiwa !

PR #3609, opened 2023-01-10
make the omp role use ansible fqrn

kayiwa:


acozine (approved):
Looks awesome, thanks!

Issue #3319, opened 2022-09-06
Run nginx -t on the inactive adc before applying a potentially destructive change


sandbergja:
We should add a pre_task to the nginx plus playbook that does the following:

- Deploy the config to the inactive load balancer
- Run `nginx -t` to make sure that the config is okay
- Stop the playbook if the config is not okay

The goal is to prevent us from running the nginx plus playbook with bad config on the active lib-adc, thereby taking all of our sites down.

acozine:
See also the documentation in the pul-it-handbook:
https://github.com/pulibrary/pul-it-handbook/blob/main/services/nginxplus.md

acozine:
Ansible gathers the IP address(es) of all hosts during fact gathering. We can use this to control the logic of the nginxplus playbook. For example, this playbook:
```
- name: test safer nginxplus approach
  hosts: nginxplus_production
  remote_user: pulsys
  strategy: linear
  become: true
  vars_files:
    - ../group_vars/nginxplus/main.yml
  tasks:
    - name: check on IP addresses
      debug: var=ansible_all_ipv4_addresses

    - name: ping only the inactive VM
      ping:
      when: "'128.112.203.146' not in ansible_all_ipv4_addresses"
```
Runs the second task only on the inactive load balancer:
```
% ansible-playbook --ask-vault-password playbooks/nginxplus-test.yml
Vault password: 

PLAY [test safer nginxplus approach] ****************************************************************************

TASK [Gathering Facts] ****************************************************************************
ok: [lib-adc2.princeton.edu]
ok: [lib-adc1.princeton.edu]

TASK [check on IP addresses] ****************************************************************************
ok: [lib-adc1.princeton.edu] => {
    "ansible_all_ipv4_addresses": [
        "128.112.203.145",
        "128.112.203.146"
    ]
}
ok: [lib-adc2.princeton.edu] => {
    "ansible_all_ipv4_addresses": [
        "128.112.203.144"
    ]
}

TASK [ping only the inactive VM] ****************************************************************************
skipping: [lib-adc1.princeton.edu]
ok: [lib-adc2.princeton.edu]

PLAY RECAP ****************************************************************************
lib-adc1.princeton.edu     : ok=2    changed=0    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0   
lib-adc2.princeton.edu     : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```
[...]

PR #3614, opened 2023-01-11
Nginx playbook should fail if run without '--limit'

acozine:
Partial fix for #3319.

Adds a failure task at the beginning of the nginx playbook.

I didn't add it to the larger `nginx_production_rebuild` playbook, and I'm not sure if we want to add it there. That playbook rebuilds the load-balancing infrastructure from scratch.


kayiwa (approved):
thanks @acozine OK!

PR #3622, opened 2023-01-13
remove all occurences of slavery-dev

kayiwa:
we no longer use the slavery-dev server


acozine:
Thanks @kayiwa. And hooray for completing the upgrade of the PAS site to Craft CMS 4!

Issue #3633, opened 2023-01-17
add ansible-tower to patch tuesday group


kayiwa:
add ansible and redhat family os'es to be upgrade

PR #3634, opened 2023-01-17
add ansible-tower1 to patch Tuesday

kayiwa:
we had ansible-tower and any redhat based OSes to be eligible for patch
tuesday

### DIFF
--- a/playbooks/os_updates.yml
+++ b/playbooks/os_updates.yml
@@ -4,7 +4,7 @@
 # to run in production, pass '-e runtime_env=production'
 # remember that the loadbalancing machines are not included
 # and will need to be manually updated
-- name: update the Operating System apt packages 
+- name: update the Operating System packages 
   hosts: "{{ runtime_env | default('staging') }}"
   remote_user: pulsys
   serial: 3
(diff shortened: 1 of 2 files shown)


kayiwa:
closes #3633

playbooks/os_updates.yml
```diff
@@ -63,20 +63,46 @@
...
+        name: "*" 
+        state: latest
+        update_cache: true
+        update_only: true
+      register: yum_update_status
+      when: ansible_os_family == "RedHat"
+
+    - name: RedHat | remove unneeded packages
+      ansible.builtin.yum:
+        autoremove: true
```
acozine, line 84: Oh, interesting, does this work without a package list?

playbooks/os_updates.yml
```diff
@@ -63,20 +63,46 @@
...
+      ansible.builtin.reboot:
+        msg: "Reboot initiated by Ansible because of yum updates."
+        connect_timeout: 5
+        reboot_timeout: 600
+        pre_reboot_delay: 0
+        post_reboot_delay: 15
+        test_command: whoami
+      when: 
+        - ansible_os_family == "RedHat"
+        - yum_update_status.changed
```
acozine, line 103: We need to register something here and then add a task to reboot the CentOS/RH boxes when required.

acozine (commented):
This is a great addition. There are a few bits to polish.

playbooks/os_updates.yml
kayiwa: If I read this right. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#parameter-autoremove

playbooks/os_updates.yml
kayiwa: I thought reboot happens when this register status changes. https://github.com/pulibrary/princeton_ansible/blob/be341d4f54912eb8c6254bbeb34e1596cfaa15a7/playbooks/os_updates.yml#L79

acozine: Oh! You're absolutely right. I was reading the `name:` of the task and not what the task actually does.

playbooks/os_updates.yml
```diff
@@ -63,20 +63,46 @@
...
+      when: ansible_os_family == "RedHat"
+
+    - name: Ubuntu | Check if reboot required
       ansible.builtin.stat:
         path: /var/run/reboot-required
       register: reboot_required_file
       when: ansible_os_family == "Debian"
 
-    - name: os_updates | Reboot if required
+    - name: RedHat | Check if reboot required
```
acozine, line 93:
```suggestion
    - name: RedHat | Reboot if required
```

PR #3638, opened 2023-01-18
i3299 nginxplus WAF

kayiwa:
- add appprotect Web Application Firewall (WAF)
- rewrite nginxplus to use the Web Application Firewall


playbooks/nginxplus.yml
```diff
@@ -21,10 +21,12 @@
 
   # updates existing load balancer
   roles:
-    - role: ../roles/nginxplus
+    - role: ../roles/nginxplusprotect
+    - role: ../roles/datadog
```
carolyncole, line 25: Is there a reason this is now needed for staging?  Should it only be run in the production env?

roles/nginxplus/README.md
```diff
@@ -488,7 +488,7 @@ nginx_stream_template:
 Dependencies
 ------------
 
-None
+This role install `nginx-ha-keepalived`. On initial installation you will need to [configure both devices](https://docs.nginx.com/nginx/admin-guide/high-availability/ha-keepalived/) concurrently with the VRRP (protocol 2) provided virtual IP from OIT.
```
carolyncole, line 491: ❤️

roles/nginxplus/files/conf/http/abid-staging.conf
```diff
@@ -29,8 +29,8 @@ server {
     ssl_prefer_server_ciphers  on;
 
     location / {
-        app_protect_enable on;
-        app_protect_security_log_enable on;
+#        app_protect_enable on;
```
carolyncole, line 32: can these be deleted?  Is there a reason for leaving them?

roles/nginxplus/tasks/plus/delete-license.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: "Delete NGINX Plus License"
-  file:
+- name: nginxplus | NGINX Plus License
```
carolyncole, line 2: Should this still have Delete?

carolyncole (changes requested):
@kayiwa this PR is Huge!  There are a number of commented out lines in the configs.  What is the reason for commenting them, vs deleting them?
```
#        app_protect_enable on;
#        app_protect_security_log_enable on;
```
I added a couple of questions about other things too.  I will say this was a very cursory review, since there were som many changes.  Others may want to put some eyes on it...

roles/nginxplus/files/conf/http/abid-staging.conf
kayiwa: Heya @carolyncole we are leaving them off until we get training from f5/nginx on how to set them up. I suppose we can delete them and enable them after the training but I elected to add them.

playbooks/nginxplus.yml
kayiwa: we only install datadog on staging. So if the playbook is run datadog will not be installed.

carolyncole (approved):
Looks ok to me.  Should someone else review?  If not feel free to hit the big green button

roles/nginxplusprotect/defaults/main.yml
```diff
@@ -0,0 +1,132 @@
+---
+# defaults file for app_protect
+# Specify whether or not this role should install the NGINX App Protect WAF product.
+nginx_app_protect_waf_enable: true
+
+# Specify whether or not this role should install the NGINX App Protect DoS product.
+nginx_app_protect_dos_enable: false
```
acozine, line 7: Heh, I first read this as "DOS" as in "MS-DOS".

acozine (commented):
This is a lot of work, and a huge step forward! Thanks @kayiwa.

I have a few questions and some suggestions that may help streamline things a little. I confess, I did not review the `nginxplusprotect` role as closely as I might have . . . could those files be pulled into a separate PR?

PR #3648, opened 2023-01-23
Updates slack messages for nginx_rebuild and openbooks playbooks

acozine:
Partial fix for #2852.

Updates the slack message in the nginx_production_rebuild playbook to be more informative. 

Also tidies up the consolidation of the openbooks playbooks.


kayiwa (approved):
thanks @acozine OK!

PR #3649, opened 2023-01-23
More slack message updates

acozine:
Partial fix for #2852.

Updates the slack messages in the `pdc_describe` and `pdc_discovery` playbooks.

Also updates the slack message and play name in the `repec` playbook.


kayiwa (approved):
thanks @acozine OK!

PR #3653, opened 2023-01-24
update and rename libstatic playbook

VickieKarasic:
Partial fix for #2852


playbooks/libstatic.yml
```diff
@@ -0,0 +1,20 @@
+---
+# by default this playbook runs in the staging environment
+# to run in qa, pass '-e runtime_env=qa'
```
sandbergja, line 3: There is no qa environment for libstatic.

sandbergja (changes requested):
Thanks, @VickieKarasic !  It will be really nice to have the unified playbook and the improved slack message!

I'm not sure this will work with the production `runtime_env`.  Currently, the production playbook specifies a separate `vars_file` and a dependency on the datadog role.  Would you be able to bring those into this playbook as well, and delete the old `libstatic_prod.yml` playbook as part of this PR?

playbooks/libstatic.yml
```diff
@@ -0,0 +1,20 @@
...
+# to run in production, pass '-e runtime_env=production'
+- name: Build libstatic infrastructure
+  hosts: libstatic_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  roles:
+    - role: libstatic
+      when: runtime_env | default('staging') == "production"
```
sandbergja, line 11: Do we need this `when`?  I think this playbook should always run the `libstatic` role.

playbooks/libstatic.yml
VickieKarasic: Thanks, @sandbergja! I removed that comment from the playbook.

playbooks/libstatic.yml
VickieKarasic: You're right! @acozine and I added the datadog role and moved the `when` to that.

VickieKarasic:
Thanks for all these helpful changes, @sandbergja! I think I caught all of them, if you'd like to take another look when you get a chance.

sandbergja (approved):
This looks awesome!  Thanks, @VickieKarasic ! 🎉

PR #3659, opened 2023-01-24
add libserv39 to patch tuesday

kayiwa:
adding libserv39 to be patched every tuesday


acozine:
Looks good, thanks @kayiwa.

Issue #3423, opened 2022-11-04
Allow upgrading bundler version alone


acozine:
Right now the `ruby_s` role only changes the version of bundler if/when you are also changing the version of ruby.

We would like to have the ability to change the version of bundler while keeping the existing version of ruby in place.

This likely involves:
- [ ] pulling the bundler-related tasks into a separate tasks file
- [ ] adding a check in tasks/main.yml to see if the installed bundler version matches the desired bundler version
- [ ] figuring out how to force-install a specific version of bundler

christinach:
The ruby interpreter will install a specific bundler version. In order to change this you need to change the ruby version and use a different bundler version. 
Locally when I want to use a different bundler version to bundle the gems:
1. I install the bundler version I want and then I run `bundle _theversionnumber_ install`
2. If the previous installed version is marked as default step 1 will not work. Then I need to look in the `specifications/default` directory and remove it from there. 
3. `gem env` provides the installation path which is `INSTALLATION DIRECTORY: /usr/local/lib/ruby/gems/3.1.0` . You can check what is installed as default in `/usr/local/lib/ruby/gems/{ruby_version}/specifications/default/`. In the bibdata-alma-staging case it indicates that bundler default is `bundler-2.3.25.gemspec`. In order to change the default : 
4.  i) `rm  /usr/local/lib/ruby/gems/{ruby_version}/specifications/default/bundler-2.3.25.gemspec` , ii) reinstall the desired ruby version as default ,  iii) uninstall the previous installed version.  
This needs to be well tested as there might be more than one bundler defaults in the default directory.

kayiwa:
closed by this #3700

PR #3667, opened 2023-01-25
[ruby_s] Move bundler out of install into main

### LINKED ISSUE #3666: [DSS] Upgrade bundler
https://github.com/pulibrary/princeton_ansible/blob/b4a8eb0a31886aefa921e2edc8d4c266be627234/group_vars/dss/common.yml#L6 
Upgrade bundler to https://github.com/pulibrary/DSS/blob/4e139bca07ebf73286c6b9658901f0ee7a9634c2/Gemfile.lock#L473

maxkadel:
When trying to upgrade the bundler for DSS, I found that running the playbook did not update the bundler after updating `bundler_version`, because I was not also changing the Ruby version. I've moved the bundler steps back into main.yml, since people are likely to upgrade bundler on their applications more often than they upgrade ruby.

- People are likely to upgrade bundler much more often than they upgrade Ruby
- [dss] Upgrade bundler to match application

Closes #3666


roles/ruby_s/tasks/main.yml
```diff
@@ -25,3 +25,17 @@
...
   when:
     - (installed_ruby.stderr is search("not found")) or (installed_ruby.stdout | regex_search('\d+\.\d+\.\d+') != ruby_version | regex_search('\d+\.\d+\.\d+'))
+
+- name: ruby_s | install global bundler, specific version
+  community.general.gem:
+    name: bundler
+    state: present
+    version: "{{ bundler_version }}"
+    user_install: false
+  when: bundler_version is defined
```
kayiwa, line 35: community gem will use the Operating Systems ruby which we go not want.

roles/ruby_s/tasks/install_from_source.yml
```diff
@@ -60,17 +60,6 @@
   ansible.builtin.shell: cd {{ install_path }}/{{ ruby_version }} && make install
   become: true
 
-- name: ruby_s | uninstall wrong version of bundler
```
kayiwa, line 63: why are we removing this?

roles/ruby_s/tasks/main.yml
maxkadel: Switched back to ansible.builtin.command

roles/ruby_s/tasks/install_from_source.yml
maxkadel: Because it does not work anywhere near consistently, and does not report that it does not work, and in the rare cases it does work it makes the role fail idempotence

acozine:
Related to #3423

acozine:
See also https://bundler.io/doc/troubleshooting.html#other-problems for tips on getting rid of wrong versions of bundler.

kayiwa:
apply the pattern in nodejs

kayiwa:
superseded by #3700

PR #3669, opened 2023-01-26
[libwww] Move Database list redirects for libguides into Production.

kevinreiss:
Testing in staging, this PR moves the new redirect rules for prod. It also removes them for staging to at the point of go live staff can still refer to the old list for a period of time to help resolve any outstanding discrepancies that remain post migration.

### DIFF
--- a/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
+++ b/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
@@ -3,9 +3,6 @@
     rewrite ^/research-data(.*)$ https://researchdata-staging.princeton.edu redirect;
     rewrite ^/AssignedSpaceApplication(.*)$ https://lockers-and-study-spaces-staging.princeton.edu/ redirect;
     rewrite ^/AssignedSpaces https://lockers-and-study-spaces-staging.princeton.edu/ redirect;
-    rewrite ^/research/databases https://libguides.princeton.edu/az.php redirect;
-    rewrite ^/research/databases/subjects https://libguides.princeton.edu/az.php redirect;
-    rewrite ^/resource/(\d+)$ https://libguides.princeton.edu/resource/$1 redirect;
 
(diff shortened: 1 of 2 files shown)


roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
```diff
@@ -3,9 +3,6 @@
     rewrite ^/research-data(.*)$ https://researchdata-staging.princeton.edu redirect;
     rewrite ^/AssignedSpaceApplication(.*)$ https://lockers-and-study-spaces-staging.princeton.edu/ redirect;
     rewrite ^/AssignedSpaces https://lockers-and-study-spaces-staging.princeton.edu/ redirect;
-    rewrite ^/research/databases https://libguides.princeton.edu/az.php redirect;
```
carolyncole, line 6: @kevinreiss did you mean to remove them from staging?

carolyncole (commented):
Looks ok to me, but are you sure you do not want staging rewrites? Or maybe put them here: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass.conf

roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
acozine: The deleted ones look like they redirected staging URLs to production libguides . . . I think? Do we have a staging site for libguides?

kevinreiss: Yes @carolyncole I removed them from staging intentionally. We are going to have a short period where Librarians can review the "old" entries in staging if needed in case something has been missed. There is not a staging site for libguides @acozine. We only have a single environment Springshare hosts for us.

acozine:
Thanks @kevinreiss, looks good.

PR #3672, opened 2023-01-30
[ansible] Ubuntu certificates will occasionally expire

kayiwa:
rather than specifically hunt for the individual expired ones we now
renew all certificates.
also updating this when running the OS updates which will fail to run


acozine:
Awesome. We can probably remove lines 8-15 of the first playbook.

PR #3673, opened 2023-01-31
[nginxplus] re-refactor nginxplus role

kayiwa:
in preparation for the WAF role re-factor nginxplus. Incorporates comments from @carolyncole and @acozine in #3638


carolyncole (approved):

roles/nginxplus/meta/main.yml
```diff
@@ -1,18 +1,19 @@
 ---
 galaxy_info:
-  role_name: nginxplud
+  role_name: nginxplus
```
acozine, line 3: Heh, woopsie. Thanks for fixing the typo!

roles/nginxplus/tasks/main.yml
```diff
@@ -1,48 +1,53 @@
...
             or nginx_rest_api_enable
 
-    - import_tasks: conf/setup-status.yml
+    - ansible.builtin.import_tasks: conf/setup-status.yml
       when: nginx_status_enable
 
-    - import_tasks: conf/debug-output.yml
+    - ansible.builtin.import_tasks: conf/debug-output.yml
       when: nginx_debug_output
   when: nginx_enable
```
acozine, line 46: I don't see this set anywhere in the group_vars. Is it meant to be `nginx_something_enable`? Or is this a variable you set manually when the playbook/role is misbehaving? If that's the use case, I'd rename this to something like:
```suggestion
  when: nginx_debug
```

roles/nginxplus/tasks/conf/debug-output.yml
```diff
@@ -1,9 +1,9 @@
...
-  command: nginx -T
+- name: nginxplus | Register NGINX configuration
+  ansible.builtin.command: nginx -T
   changed_when: false
   register: nginx_configuration
 
-- name: "Setup: Print NGINX configuration"
-  debug:
+- name: nginxplus | Print NGINX configuration
+  ansible.builtin.debug:
```
acozine, line 8: Does this run in testing? If so, does it fail idempotency? The documentation isn't clear, but my bet is that the `debug` module always returns `changed`. You've already got `changed_when: false` on the prior task . . .

roles/nginxplus/tasks/keys/apt-key.yml
```diff
@@ -1,14 +1,10 @@
...
-- name: "Install: Set NGINX Signing Key URL"
-  set_fact:
-    keysite: "{{ nginx_signing_key | default(default_keysite) }}"
-
-- name: "Install: Add APT NGINX Signing Key"
-  apt_key:
+- name: nginxplus | Add NGINX signing key
+  ansible.builtin.apt_key:
+    id: 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62
+    keyring: /usr/share/keyrings/nginx-archive-keyring.gpg
```
acozine, line 9: I don't entirely understand this change, and enquiring minds want to know . . . No objections, I'm just curious.

roles/nginxplus/tasks/localhost.yml
```diff
@@ -8,7 +14,7 @@
     mode: 0755
   when: running_on_server
 
-- name: add pul svg file
+- name: nginxplus | add pul svg file
```
acozine, line 17: This is very nitpicky, but for clarity and ease of reading, I might change:
```suggestion
- name: nginxplus | add pul logo
```

roles/nginxplus/tasks/main.yml
```diff
@@ -1,48 +1,53 @@
...
+- name: nginxplus | adjust limits
+  ansible.builtin.import_tasks: adjust_ulimit.yml
   when: nginx_type == "plus"
 
-- import_tasks: keys/apt-key.yml
+- name: nginxplus | add keys
+  ansible.builtin.import_tasks: keys/apt-key.yml
 
 - name: Install NGINX
   block:
```
acozine, line 13: Aha! Here are the tasks with no name.

roles/nginxplus/tasks/modules/install-xslt.yml
```diff
@@ -1,18 +1,18 @@
...
+- name: nginxplus | Install NGINX Plus XSLT Module
+  ansible.builtin.apt:
     name: nginx-plus-module-xslt
     state: present
   when: nginx_type == "plus"
 
-- name: "Setup: Load NGINX XSLT Module"
-  lineinfile:
+- name: nginxplus | Load NGINX XSLT Module
+  ansible.builtin.lineinfile:
```
acozine, line 15: We've got a lot of `lineinfile` tasks for the Nginx config file. How about using a template instead of editing and re-editing the file on the server?

roles/nginxplus/tasks/plus/install-plus.yml
```diff
@@ -1,13 +1,37 @@
...
+- name: nginxplus | NGINX Plus repository
+  ansible.builtin.apt_repository:
+    filename: nginx-plus
+    repo: deb {{ nginx_signed_statement }} https://pkgs.nginx.com/plus/ubuntu {{ ansible_distribution_release }} nginx-plus
+    update_cache: false
+    state: present
+    mode: 0644
+  when: nginx_manage_repo | bool
+
+- name: nginxplus | {{ nginx_setup | capitalize }} NGINX Plus
```
acozine, line 29: This is a cool construction.

In `group_vars/common.yml` I see `nginx_setup: install` - what other values could `nginx_setup` have? Would we set it to `upgrade` or `downgrade` somehow when changing the version? I'm trying to envision what this output will look like.

roles/nginxplus/vars/main.yml
```diff
@@ -1,2 +1,64 @@
 ---
 # vars file for roles/nginxplus
+nginx_signed_statement: "[signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg]"
```
acozine, line 3: How did you decide what should go here and what should go into `group_vars/common.yml`?

roles/nginxplus/tasks/plus/setup-debian.yml
```diff
@@ -1,17 +1,17 @@
...
+- name: nginxplus | Add NGINX Plus Repository
+  ansible.builtin.apt_repository:
+    repo: deb {{ nginx_signed_statement }} https://pkgs.nginx.com/plus/ubuntu {{ ansible_distribution_release }} nginx-plus
     filename: nginx-plus
     update_cache: false
 
-- name: "Install: Verify NGINX Plus License"
-  blockinfile:
+- name: nginxplus | Verify NGINX Plus License
+  ansible.builtin.blockinfile:
```
acozine, line 9: This is the second `blockinfile` task for the `/etc/apt/apt.conf.d/90nginx` file. Could we use a template instead?

acozine (commented):
Great stuff! I added some questions and comments.

PR #3681, opened 2023-02-02
adds honeybadger key to lib_jobs staging, follows pattern for HB usage

### LINKED ISSUE #3658: [lib-jobs] Honeybadger key missing from lib_jobs staging
When I deploy to lib-jobs staging, I get a honeybadger API key error: https://gist.github.com/pulbot/adeb41d0bd18123e59b929f070b7a8ad#file-stdout-L255-L274

The error doesn't appear when deploying to production: https://gist.github.com/pulbot/31ba0279216aa7d23a28698a1c955aa7#file-stdout-L249-L254

We should add the honeybadger API key, and also set the `HONEYBADGER_ENV` environment variable to staging for this box.

acozine:
Closes #3658.

This PR makes three changes. Do they represent the best way to accomplish this goal? 

1. Changes the vaulted variable name to `vault_lib_jobs_honeybadger_key`.
2. Sets the environment for Honeybadger to the value of `rails_env` for all cases.
3. Sets a rails app var for the Honeybadger key in the staging variables.

Production already has the Honeybadger key but no rails app var for it. Do we want to add `HONEYBADGER_API_KEY` as a rails app var for production as well?


group_vars/lib_jobs/common.yml
```diff
@@ -22,3 +22,7 @@ passenger_ruby: /usr/local/bin/ruby
 ruby_version_default: "ruby-3.1.2"
 ubuntu_ruby_version_default: "ruby-3.1.2"
 desired_nodejs_version: v16.19.0
+# shared rails variables
+rails_app_vars:
+  - name: HONEYBADGER_ENV
```
sandbergja, line 27: Thanks, @acozine !  After I ran this on staging, the `HONEYBADGER_ENV` var wasn't set.  I wonder if `rails_app_vars` is getting overriden?

sandbergja (commented):
Looks good!  Just one question about the `HONEYBADGER_ENV`

group_vars/lib_jobs/common.yml
maxkadel: It looks like lib_jobs is defining the rails_app_vars in `roles/lib_jobs/vars/main.yml`, and I think when you put it in `common.yml`, it doesn't just overwrite one value, it overwrites the whole list.

acozine:
Ah, yes, role vars (from `role_name/vars/main.yml`) override group_vars. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable. I've updated this to put all the `rails_app_vars` into group vars, as that is our general pattern.

group_vars/lib_jobs/staging.yml
```diff
@@ -38,3 +38,9 @@ voucher_feed_recipients: '{{ vault_lib_jobs_staging_voucher_feed_recipients }}'
 # Aspace
 app_lib_jobs_aspace_url: 'https://aspace-staging.princeton.edu/staff/api'
 app_lib_jobs_aspace_staging_url: 'https://aspace-staging.princeton.edu/staff/api'
+# Honeybadger
+lib_jobs_honeybadger_key: '{{ vault_lib_jobs_honeybadger_key }}'
+rails_app_vars:
+  - name: HONEYBADGER_API_KEY
+    value: '{{ vault_lib_jobs_honeybadger_key }}'
```
maxkadel, line 45: I think you'll need to remove this stanza - otherwise it will override all the `rails_app_vars` in the `group_vars/common.yml` with just the `HONEYBADGER_API_KEY` name / value pair.

group_vars/lib_jobs/common.yml
```diff
@@ -13,12 +13,114 @@ application_db_name: '{{ app_db_name }}'
...
 install_ruby_from_source: true
-ruby_version_override: "ruby-3.1.2"
-bundler_version: "2.3.26"
 passenger_ruby: /usr/local/bin/ruby
 ruby_version_default: "ruby-3.1.2"
+ruby_version_override: "ruby-3.1.2"
 ubuntu_ruby_version_default: "ruby-3.1.2"
-desired_nodejs_version: v16.19.0
+# shared rails variables
+rails_app_vars:
```
maxkadel, line 26: Good call moving these to common.yml, matches our other apps better.

sandbergja:
Thanks, @acozine !  I just pushed a small commit that got this working for me.

christinach (approved):

PR #3682, opened 2023-02-02
updates SSL certs and keys for four sites

acozine:
Updates the cert and key for these four sites:
- annotations
- library
- mudd
- openbooks-staging
Note that we set the cert for `openbooks-staging` to expire in 30 days so we can test the auto-renew feature of OIT's current certificate request and renewal process.


kayiwa (approved):
thanks @acozine! OK!

PR #3683, opened 2023-02-04
[server maintenance] add libserv31 

kayiwa:
include this server to the patch tuesday group


acozine:
Looks good, thanks @kayiwa.

PR #3686, opened 2023-02-06
[ansible operations] add pulmirror vm

### LINKED ISSUE #3685: Add pulmirror box to production group
Should we add the [pulmirror](https://github.com/pulibrary/princeton_ansible/blob/main/inventory/by_cloud/google_cloud#L15) group as a child of the production group? That way it will be included in Patch Tuesday.

kayiwa:
the pulmirror vm is not included in the patch tuesday group
this adds it

will close #3685


acozine (approved):
Excellent, thanks!

PR #3691, opened 2023-02-09
Add collection reqs

acozine:
This list of collection requirements in our repo tells Tower/Controller to load the listed collections onto the default execution environment. The list should probably be longer, but for now, this change got me past the error:
```
ERROR! couldn't resolve module/action 'community.general.sysupgrade'. This often indicates a misspelling, missing collection, or incorrect module path.
```


acozine:
Add a comment to the file that it's related to Tower and EEs.

kayiwa (approved):
thanks @acozine watched this work! OK

PR #3700, opened 2023-02-15
Allow bundler version to change independent of ruby

carolyncole:
Run for putting a new version of bundler on tigerdata-staging


acozine:
Failing idempotence when the `TASK [update cache]` in the ruby_s role reports changed on the second run.

kayiwa (approved):
OK! thanks @carolyncole

PR #3703, opened 2023-02-16
Next steps on code-deploy playbook

acozine:
I merged #3494 so I could test this in Tower.

After the first commit on this branch, the Tower job runs but fails with `bundler: command not found`. Looking forward to picking this work up in the working group tomorrow!


acozine:
Put the public key for Ansible Tower into the [keys](https://github.com/pulibrary/princeton_ansible/tree/main/keys) directory, then add it into the deploy_keys rotation. See [figgy group vars](https://github.com/pulibrary/princeton_ansible/blob/b2c1788a07c213a9e91ecf7efaf0d5d9b023aa31/group_vars/figgy/production.yml#L177) for an example.

acozine:
I made a fresh key on the appdeploy server, put the matching public key on the rest of the VMs, and updated the `.ssh/config` files for both the `pulsys` and `deploy` users on the appdeploy server, but deploying still fails with `/opt/appdeploy/shared/bundle/ruby/2.6.0/gems/net-ssh-6.1.0/lib/net/ssh.rb:268:in 'start': Authentication failed for user deploy@lockers-and-study-spaces-staging1.princeton.edu (Net::SSH::AuthenticationFailed)`. I get the same error when I try it manually and from the playbook.

acozine:
As of today, this works on all DLS and DACS projects. We can update as needed if any RDSS projects don't deploy.

maxkadel (approved):

PR #3713, opened 2023-02-23
[pulfalight] don't serve findingaids site from findingaids-beta anymore

hackartisan:
refs pulibrary/pulfalight#1220


kayiwa (approved):
OK! Thanks @hackartisan

PR #3722, opened 2023-02-28
[ansible] update paramiko

### LINKED ISSUE #3694: CryptographyDeprecationWarning: Blowfish has been deprecated   "class": algorithms.Blowfish
When running molecule I get the above warning.

Do we need to do something about it?

kayiwa:
paramiko library has deprecated use of blowfish

closes #3694


carolyncole:
Already done by dependabot

PR #3724, opened 2023-03-01
[dss] update to jammy jelly

kayiwa:
by default python 2 is not available in jammy jellyfish and we do not
really need it on bionic beaver


carolyncole (approved):
Tested by running the tigerdata playbook with this PR.

PR #3740, opened 2023-03-07
[imagecat-staging] add TLS configuration

kayiwa:


hackartisan (changes requested):
I think we wanted this locked down to VPN; here's the example: https://github.com/pulibrary/princeton_ansible/blob/f00c9cf5789c81f20c08c6bcbc1a5af6d9c6f4d5/roles/nginxplus/files/conf/http/catalog-staging.conf#L46-L49

leefaisonr:
Thank you for the tip, @hackartisan!

PR #3741, opened 2023-03-07
WIP Proxy pass template

acozine:
Supersedes #2776. I opened that PR off of a clone of the repo that no longer exists.

I've rebased the branch today so it will run on our new CI infrastructure. I'll close the other PR but please look there for earlier conversations on this topic!

The template we've been using for proxy pass configuration is more of a file than a template.

This PR pulls the data for proxy passes into its own (alphabetized) vars file, makes the template a template, and adds a task to fill out the template. Separating the data out doesn't reduce maintenance, but it does make the entries more readable, so folks can more easily add or edit entries for sites that need proxy passes.


acozine:
Current CI failure is in the libwww role:
```
  TASK [libwww : libwww | ensure MaxRequestWorkers is set for system use] ********
  An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NoneType: None
  fatal: [instance]: FAILED! => {"changed": false, "msg": "src and dest are required"}
```

acozine:
Less of an issue now that Drupal sites are externally hosted. Won'tifx.

PR #3744, opened 2023-03-08
Ealapps remove hishi

tventimi:
Removes all references to the Hishi app


kayiwa (approved):
OK! thank you @tventimi

PR #3750, opened 2023-03-14
Use 3.1.3 in .tool-versions

christinach:
Use ruby 3.1.3 in tool-versions

`asdf list all ruby` indicates that 3.2.0 is a dev release `3.2.0-dev`


christinach:
I updated the asdf ruby plugin and this was fixed. Thanks @sandbergja

PR #3751, opened 2023-03-17
Update arks-redirect config, use auto-renew certs

acozine:
The certificate for the arks site expired today. I used playbooks/incommon_certbot.yml to renew the key and cert with ACME.

This PR updates the config file for the site to point to the new key and cert and removes the old keys and certs from this repo.


kayiwa (approved):
OK! thanks @acozine

PR #3752, opened 2023-03-20
[nginxplus TLS] renew certs

kayiwa:
lib-adc.princeton.edu

remove old TLS certs to use ACME


VickieKarasic (approved):
Looks good - thanks, Francis

PR #3753, opened 2023-03-20
jammy jellyfish uses https

kayiwa:
add a new scheme for ppa path


kayiwa:
upstream mariadb uses mariadb 10.6

PR #3757, opened 2023-03-22
[appdeploy] add heaven machine to patch tuesday

### LINKED ISSUE #3756: app_deploy (aka pulbot, aka heaven, I think) not participating in Patch Tuesdays
While looking into some ssh errors while trying to deploy to new environments, I noticed that it looked like it had not been updated in some time, then investigated and saw that it is not included in any of the "by_environment" inventory files, meaning it's not getting automatically updated.

kayiwa:
this changes the playbook to follow the pattern for playbooks
adds the app_deploy inventory to production

closes #3756

### DIFF
--- a/inventory/all_projects/app_deploy
+++ b/inventory/all_projects/app_deploy
@@ -1,2 +1,2 @@
-[app_deploy]
+[app_deploy_production]
 appdeploy1.princeton.edu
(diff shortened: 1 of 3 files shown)


maxkadel (approved):
Thanks!

PR #3763, opened 2023-03-27
[mariadb] upgrade mariadb to jammy

kayiwa:
jammy jellyfish upstream repository only uses MySQL 8
we update to mariadb 10.6 which is compatible
remove systemd calls which only run on servers


acozine:
Depends on changes to the applications using mariadb. See https://github.com/PrincetonUniversityLibrary/pas-craft3/pull/91 for an example.

acozine:
Related to #3962

inventory/by_environment/dev
```diff
@@ -0,0 +1,2 @@
+[dev:children]
+mysql_dev
```
acozine, line 2: We also would not need this. Much tidier.

playbooks/mysql.yml
```diff
@@ -1,6 +1,8 @@
 ---
 # by default this playbook runs in the staging environment
 # to run in production, pass '-e runtime_env=production'
+# to run in dev, pass '-e runtime_env=dev'
+#
```
acozine, line 5: This wouldn't be needed either.

roles/mysql/tasks/main.yml
```diff
@@ -1,19 +1,44 @@
...
+
+- name: MySQL | ensure keyrings directory exist
+  ansible.builtin.file:
+    path: /etc/apt/keyrings
+    state: directory
+    mode: "0755"
+
 - name: MySQL | Install MariaDB key
-  ansible.builtin.apt_key:
+  ansible.builtin.get_url:
```
acozine, line 17: Possibly relevant here - see the first example in the [docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html#ansible-collections-ansible-builtin-apt-key-module) on `apt_key`.

roles/mysql/tasks/main.yml
```diff
@@ -37,9 +62,10 @@
 
 - name: MySQL | Restart MySQL
   ansible.builtin.service:
-    name: mysql
+    name: "{{ mysql_name }}"
```
acozine, line 65: Double-check that we give this var a default setting in the mysql role and/or set it for all applications to avoid the `variable undefined` error.

roles/mysql/vars/main.yml
```diff
@@ -33,4 +33,5 @@ mysql_configuration_options:
 
 mysql_configuration_destination: /etc/mysql/mariadb.conf.d
 
-mariadb_repo: "http://sfo1.mirrors.digitalocean.com/mariadb/repo/10.3/ubuntu"
+mariadb_repo: "https://deb.mariadb.org"
+mariadb_repo_version: 10.6
```
acozine, line 37: Jammy runs mariadb 10.6. During the upgrade process, we want prod and staging to stay on 10.3.

We might want to set this in `roles/mysql/defaults/main.yml` so it will be overridden by any other setting.

acozine (commented):
This is a great step forward!

PR #3764, opened 2023-03-29
[figgy] update figgy to focal

kayiwa:
- updating figgy to use ruby from source
- update vips
- use mediainfo apt repo configuration
- use rabbitmq apt repo configuration


kayiwa:
we will be using jammy jellyfish when #3783 gets into main

PR #3765, opened 2023-03-29
change slack channel for alerts from 'server-alerts' to 'ansible-alerts'

acozine:
We have a new channel for `sensu-alerts`, we no longer get alerts from netdata or anything else . . . so it might be nice to use `ansible-alerts` instead of `server-alerts` in Slack. This is more specific, more descriptive, and easier to understand if you're browsing Slack channels.

This PR updates all instances of alerts sent to `server-alerts` on the `main` branch. We probably have a couple more on branches/PRs.


hackartisan (approved):
Great idea to have its own channel. Maybe if we want to change this again in the future it would be good to use a top-level variable.

PR #3766, opened 2023-03-30
New appdeploy

kayiwa:


roles/towerdeploy/meta/main.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  license: MIT
+
+  min_ansible_version: 2.9
+
+  platforms:
+    - name: Ubuntu
+      versions:
+        - jammy
+dependencies:
```
maxkadel, line 16: Can we remove this dependency? Otherwise it will run the ruby_s role twice, I think - my understanding is that this was the purpose of having the roles in the playbook, rather than the meta.yml.

roles/towerdeploy/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+# tasks file for towerdeploy
+
+- name: towerdeploy | install capistrano
+  ansible.builtin.command: gem install capistrano --version "{{ cap_version }}"
+  register: gemoutput
+  changed_when: gemoutput.rc != 0
+  failed_when: gemoutput.rc !=0
+
+- name: towerdeploy | install capistrano dependencies
+  ansible.builtin.command: gem install "{{ item }}" airbrussh
```
maxkadel, line 11:
```suggestion
  ansible.builtin.command: gem install "{{ item }}"
```

maxkadel:
It looks like we'll need to add [capistrano-composer](https://rubygems.org/gems/capistrano-composer) for discovery utils - see [failed ansible tower run output](https://ansible-tower.princeton.edu/#/jobs/playbook/310/output)

roles/towerdeploy/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+- name: towerdeploy | install capistrano
+  ansible.builtin.command: gem install capistrano --version "{{ cap_version }}"
+  register: gemoutput
+  changed_when: gemoutput.rc != 0
+  failed_when: gemoutput.rc !=0
+
+- name: towerdeploy | install capistrano dependencies
+  ansible.builtin.command: gem install "{{ item }}"
+  register: capgemoutput
+  changed_when: capgemoutput.rc != 0
```
maxkadel, line 13: Is it possible to make this more sensitive? When I added capistrano-composer it did not reflect that there was a change.

roles/towerdeploy/meta/main.yml
kayiwa: we will be unable to test this role if we don't have `ruby_s`

roles/towerdeploy/tasks/main.yml
kayiwa: heya @maxkadel the `ansible.builtin.command` worked more reliably `community.general.gem` in all our tests. We you expecting output?

kayiwa: I have issued this https://github.com/pulibrary/princeton_ansible/issues/3781

PR #3776, opened 2023-04-05
[libjob] create an open marc records mount

kayiwa:
- add nfs file servers to inventory
- enable creation of mounts through a loop
- deploy role addition to nfs
- create the open_marc_records mount

### DIFF
--- a/playbooks/nfsserver.yml
+++ b/playbooks/nfsserver.yml
@@ -8,7 +8,6 @@
   vars_files:
     - ../group_vars/nfsserver/{{ runtime_env | default('staging') }}.yml
   roles:
-    - role: roles/deploy_user
     - role: roles/nfsserver
 
   post_tasks:
--- a/playbooks/nfsserver.yml
+++ b/playbooks/nfsserver.yml
@@ -1,0 +1,19 @@
+---
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- name: configure nfsserver
+  hosts: lib_fs_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/nfsserver/{{ runtime_env | default('staging') }}.yml
+  roles:
+    - role: roles/deploy_user
+    - role: roles/nfsserver
+
+  post_tasks:
+    - name: tell everyone on slack you ran an ansible playbook
+      community.general.slack:
+        token: "{{ vault_pul_slack_token }}"
+        msg: "Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}"
+        channel: #server-alerts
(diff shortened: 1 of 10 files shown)


christinach (approved):
Thanks @kayiwa !

roles/nfsserver/meta/main.yml
```diff
@@ -14,4 +14,4 @@ galaxy_info:
       versions:
         - bionic
 dependencies:
-  - role: "common"
+  - role: deploy_user
```
maxkadel, line 17: Can we use the new pattern where we put the dependencies in the playbook roles stanza, rather than the meta/main.yml? 

This will help us move towards having less redundancy in our runs and having both runs and tests run a lot faster.

maxkadel (commented):
I'm excited to have a staging nfsserver! This is looking great - just the question on the meta/main.yml, and on lib-jobs-staging1 the mount owner and group look weird to me

```zsh
deploy@lib-jobs-staging1:~$ ls -la /mnt/nfs/open_marc_records/
total 8
drwxr-xr-x 2 nobody 4294967294 4096 Apr  5 13:46 .
drwxr-xr-x 3 root   root       4096 Apr  5 13:45 ..
```

roles/nfsserver/meta/main.yml
maxkadel: it seems to me like if you changed the playbook to have:
```yaml
  roles:
    - role: roles/deploy_user
    - role: roles/nfsserver
```
then that task would not fail, because once the nfsserver role ran, the deploy_user role would already have run

acozine: Yes, if we remove stuff from `meta/main.yml` we need to add it in the playbook.

acozine: The idea is to include the dependencies under `roles:` in the playbook instead of under `dependencies` in `meta/main.yml`. We'd need them in both the operational playbook and the testing playbook for each project.

acozine: If the playbooks have `- deploy_user` listed in the `roles:` section, this task would succeed.

playbooks/nfsserver.yml
```diff
@@ -0,0 +1,18 @@
...
+# by default this playbook runs in the staging environment
+# to run in production, pass '-e runtime_env=production'
+- name: configure nfsserver
+  hosts: lib_fs_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/nfsserver/{{ runtime_env | default('staging') }}.yml
+  roles:
+    - role: roles/nfsserver
```
acozine, line 11: Related to the comments on the `meta/main.yml` below, this is where you'd add the deploy user role: ```suggestion
    - role: roles/deploy_user
    - role: roles/nfsserver
```

maxkadel (approved):
Looks good! No need for the new pattern to be a blocker.

PR #3787, opened 2023-04-12
[figgy] update figgy to jammy image

kayiwa:
- add figgy-web-staging-1 to inventory and loadbalancer
- remove unneeded kakadu role
- make medianrole compatible with keyring mode for repositories
- upgrade vips version
- use quay.io image and unpin gdal

### DIFF
--- a/roles/figgy/meta/main.yml
+++ b/roles/figgy/meta/main.yml
@@ -7,15 +7,15 @@ galaxy_info:
 
   license: MIT
 
-  min_ansible_version: 2.2
+  min_ansible_version: 2.9
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - bionic
+        - jammy
 dependencies:
   - role: "redis"
-  - role: "kakadu"
   - role: "mediainfo"
   - role: "vips"
   - role: "imagemagick"
--- a/roles/figgy/tasks/main.yml
+++ b/roles/figgy/tasks/main.yml
@@ -209,8 +209,8 @@
     name: "{{ item }}"
     state: present
   loop:
-    - "libgdal-dev=2.4.2+dfsg-1~bionic0"
-    - "gdal-bin=2.4.2+dfsg-1~bionic0"
+    - "libgdal-dev"
+    - "gdal-bin"
 
 - name: figgy | Install cogeo-mosaic apt prerequisites
   ansible.builtin.apt:
--- a/roles/mediainfo/tasks/main.yml
+++ b/roles/mediainfo/tasks/main.yml
@@ -1,12 +1,16 @@
 ---
 # tasks file for mediainfo
-- name: medianinfo | Install mediaserver repository
+- name: mediainfo | download mediainfo repo config
+  ansible.builtin.get_url:
+    url: "https://mediaarea.net/repo/deb/repo-mediaarea_1.0-21_all.deb"
+    dest: /tmp/repo-mediaarea_1.0-21_all.deb
+
+- name: mediainfo | Install mediainfo repo
   ansible.builtin.apt:
-    deb: "https://mediaarea.net/repo/deb/repo-mediaarea_{{ mediaarea_repo_version }}_all.deb"
-    state: present
-  when: running_on_server
+    deb: /tmp/repo-mediaarea_1.0-21_all.deb
 
 - name: mediainfo | Install mediainfo
   ansible.builtin.apt:
     name: mediainfo
     state: present
+    update_cache: true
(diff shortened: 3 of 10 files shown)


roles/mediainfo/tasks/main.yml
```diff
@@ -1,12 +1,16 @@
...
-    deb: "https://mediaarea.net/repo/deb/repo-mediaarea_{{ mediaarea_repo_version }}_all.deb"
-    state: present
-  when: running_on_server
+    deb: /tmp/repo-mediaarea_1.0-21_all.deb
 
 - name: mediainfo | Install mediainfo
   ansible.builtin.apt:
     name: mediainfo
     state: present
+    update_cache: true
```
acozine, line 16: Is this going to fail idempotency? We might want to add a cache valid time.

roles/mediainfo/tasks/main.yml
```diff
@@ -1,12 +1,16 @@
 ---
 # tasks file for mediainfo
-- name: medianinfo | Install mediaserver repository
+- name: mediainfo | download mediainfo repo config
```
acozine, line 3: I'm confused by `config` here - isn't this the full mediainfo repo?
```suggestion
- name: mediainfo | download mediainfo repo
```

roles/figgy/tasks/main.yml
```diff
@@ -209,8 +209,8 @@
     name: "{{ item }}"
     state: present
   loop:
-    - "libgdal-dev=2.4.2+dfsg-1~bionic0"
-    - "gdal-bin=2.4.2+dfsg-1~bionic0"
+    - "libgdal-dev"
```
acozine, line 212: This change makes me so happy! No more Bionic cruft.

acozine (commented):
Overall this looks great. It will feel wonderful to get Figgy running on Jammy boxes. I had a couple of questions, see below.

roles/mediainfo/tasks/main.yml
kayiwa: I imagined it would have but wasn't. I've removed it.

roles/figgy/meta/main.yml
```diff
@@ -7,15 +7,15 @@ galaxy_info:
...
 
-  min_ansible_version: 2.2
+  min_ansible_version: 2.9
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - bionic
+        - jammy
```
acozine, line 16: Heh. A giant leap!

roles/figgy/tasks/main.yml
acozine, line 212: Yay!

roles/figgy/molecule/default/verify.yml
```diff
@@ -21,9 +21,3 @@
     failed_when:
       - not figgy_mount.stat.exists
       - not figgy_mount.stat.isreg
-  - name: test for cogeo-mosaic command
```
acozine, line 24: Purely out of curiosity - what was this for?

roles/mediainfo/tasks/main.yml
```diff
@@ -1,10 +1,13 @@
...
+  ansible.builtin.get_url:
+    url: "https://mediaarea.net/repo/deb/repo-mediaarea_1.0-21_all.deb"
+    dest: /tmp/repo-mediaarea_1.0-21_all.deb
+
+- name: mediainfo | Install mediainfo repo
   ansible.builtin.apt:
-    deb: "https://mediaarea.net/repo/deb/repo-mediaarea_{{ mediaarea_repo_version }}_all.deb"
-    state: present
-  when: running_on_server
+    deb: /tmp/repo-mediaarea_1.0-21_all.deb
```
acozine, line 10: This looks weird. I checked the docs and the `state` param is not required, but I would vote to keep it for readability.

roles/mediainfo/tasks/main.yml
```diff
@@ -1,10 +1,13 @@
...
 
 - name: mediainfo | Install mediainfo
```
acozine, line 12: Huh. Is this the same task as the one above? I mean, does it do anything different?

acozine (commented):
Overall looks good. See below for a few questions/comments.

roles/figgy/molecule/default/verify.yml
kayiwa: I am not sure. Running python in a container proved monumentally difficult for me.

roles/mediainfo/tasks/main.yml
kayiwa: no it isn't. The former installs software that configures the repository at `/etc/apt/sources.list.d/mediaarea.list` which then allows us to install the mediaarea software with the latter. Which reminds me I need to put the update cache so the `apt` updates its repos

acozine:
Addressed in #4014.

PR #3094, opened 2022-06-14
add serial running of playbook

kayiwa:
The PR will prevent the playbook from locking up machines
Co-authored-by: Alicia Cozine<acozine@users.noreply.github.com>
Co-authored-by: Jane Sandberg <sandbergja@users.noreply.github.com>
Co-authored-by: Ryan Laddusaw<rladdusaw@users.noreply.github.com>
Co-authored-by: James R. Griffin III <jrgriffiniii@users.noreply.github.com>
Co-authored-by: Max Kadel <maxkadel@users.noreply.github.com>

PR #3789, opened 2023-04-14
Make deploy_user playbook & role quicker

acozine:
Because we want folks to rotate their SSH keys periodically, we want to make it easy to deploy new keys. This PR:
- Adds two refinements to the deploy_user role and playbook:
  - use `run_once` on the task that pulls keys from GitHub (which takes a long, long time)
  - adds a tag called `update_keys` to the two tasks in the `deploy_user` role - when we only need to update keys, this will make the process much quicker, because we won't be running all the other tasks (creating the deploy user, allowing the `authorized_keys` file, etc.) on existing/built VMs
- Updates the documentation in the playbook
- Comments out the `serial: 2` setting
  - when run with `serial`, Ansible runs any `run_once` tasks once per group, so with `serial: 2` we would get at most a 50% speed improvement; I hope that with this task set to `run_once`, the playbook will be much less disruptive; we should test to confirm

### DIFF
--- a/playbooks/deploy_user.yml
+++ b/playbooks/deploy_user.yml
@@ -1,10 +1,19 @@
 ---
-# Only run this playbook on all hosts in a maintenance window, because no one can deploy while the playbook is running.
-# If you are running it outside the maintenance window please run it on one box at a time using a limit flag.
-# For example: `ansible-playbook playbooks/deploy_user.yml --limit slavery-prod1.princeton.edu`
+# If you run the full deploy_user role
+# no-one can deploy until the playbook stops!
+# So only run the full playbook:
+#  - during a maintenance window on all hosts, or
+#  - with `serial: 2` uncommented, or
+#  - on one box at a time using a limit flag, for example:
+#    `ansible-playbook playbooks/deploy_user.yml --limit slavery-prod1.princeton.edu`
+#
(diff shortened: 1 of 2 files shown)


playbooks/deploy_user.yml
```diff
@@ -1,10 +1,19 @@
 ---
-# Only run this playbook on all hosts in a maintenance window, because no one can deploy while the playbook is running.
```
hackartisan, line 2: Which is / was the step that makes it so that no one can deploy? Is there a version of this playbook the deletes the authorized keys file altogether or something?

acozine: I'm not sure. We added the warnings and the `serial` setting last year, before we changed the way we started using a template to manage authorized keys:
Warnings in May 2022: #3030
Serial in June 2022: #3094
Template in October 2022: #3377
Before October, we used the [authorized_key module](https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html#ansible-collections-ansible-posix-authorized-key-module) to manage keys.

My instinct would be to test the updates on a small group of staging servers and see how it behaves.

bess (approved):

bess:
Tested on staging servers with @acozine , @leefaisonr and @VickieKarasic

PR #3794, opened 2023-04-17
[ansible] make roles use fqrn

kayiwa:
- make the bibdata_sqs_poller use fqrn
- add blacklight app to use ansible fqrn
- make cantaouple role use fqrn


leefaisonr (approved):
Looks good to me - approved

PR #3809, opened 2023-04-20
Update install location for mailcatcher

bess:
We upgraded to ruby 3.1.0 so we need to update the mailcatcher location to match

Fixes:
* https://github.com/pulibrary/pdc_describe/issues/1085


hectorcorrea (approved):

hectorcorrea:
@bess great catch!

PR #3810, opened 2023-04-21
[lib-sftp] add aspace directory

kayiwa:
make sure directory is owned by same user
make the playbook have documentation on how to add a user


regineheberlein (approved):
Thanks so much Francis!

PR #3814, opened 2023-04-21
[lib_jobs] Add SCSB S3 connection vars to lib_jobs

### LINKED ISSUE #3807: [lib_jobs] New variables for SCSB AWS S3 connection
- [x] Get the SCSB production and staging values from the bibdata vault and copy them in the lib_jobs vault. 

- [x] Copy the SCSB_S3_UPDATES path https://github.com/pulibrary/princeton_ansible/blob/e92d8babcd9166b00e14b90a7bafceb5cbb5366b/group_vars/bibdata/production.yml#L84 in lib_jobs group vars staging and production.

These values will be used for submit collection and the connection to SCSB AWS S3 bucket.

maxkadel:
Run on lib-jobs-staging1 successfully `ansible-playbook playbooks/lib_jobs.yml --tags site_config`

Closes #3807


group_vars/lib_jobs/production.yml
```diff
@@ -80,3 +80,7 @@ datadog_checks:
 # Aspace
 app_lib_jobs_aspace_url: 'https://aspace.princeton.edu/staff/api'
 app_lib_jobs_aspace_staging_url: ''
+# SCSB S3
+scsb_s3_access_key: '{{ vault_scsb_s3_access_key_local }}'
```
christinach, line 84: @maxkadel instead of `.._local` can we rename them to `..._production`?

christinach (changes requested):

group_vars/lib_jobs/production.yml
maxkadel: What's confusing is that the `vault_scsb_s3_access_key_prod` is used for the `SCSB_S3_PARTNER_ACCESS_KEY` in staging, qa, and production in Bibdata.

I think there are two distinctions we need to make - the Princeton vs. Partner access keys, and the access keys by environment. It looks like the partner access keys are the same regardless of environment, whereas the Princeton access keys are different between the different environments. I'll try a commit and see what you think.

christinach: @maxkadel, after I submitted my review I remembered that there is an inconsistency in the names of the envs in bibdata. I came across this in the past but forgot to create an issue. 

For now and in this lib_jobs PR we should not include the vars/values with the 'local' name vault_scsb_s3_access_key_local and vault_scsb_s3_secret_access_key_local
These refer to the alma-events and in the AWS webhook but for some reason they were given the wrong name and they are not needed in lib-jobs. 
I will create a different princeton-ansible ticket so that we can address this issue.

christinach (approved):

PR #3820, opened 2023-04-24
Updates firewall playbook

acozine:
We will update this PR incrementally as we learn to automate tasks on the test firewall.

This playbook can only be run in Ansible Tower.


acozine:
Current state: the `Read-only Network Rules` template (based on the  `firewall_update` playbook) successfully authenticates to the test firewall. It fails on the `Get a list of all security rules` task - see [this task output](https://ansible-tower.princeton.edu/#/jobs/playbook/604/output). Next steps:
- get that task working
- extract the IP address of a VM based on its DNS name
- view network rules for that specific IP

acozine:
I tried moving the tasks that install dependencies out of the playbook and into a custom EE build script, but that failed. For now I've reverted the commit; we can try debugging it later.

acozine:
85 commits is too many!

Closing in favor of #4870.

PR #3826, opened 2023-04-25
Updates deploy role to run in Tower

acozine:
When running the `deploy_user.yml` playbook with the tag `update_keys` in Ansible Tower, the task that replaces the  deploy user's `authorized_keys` file failed with this error:
```
Failed to get information on remote file (/home/deploy/.ssh/authorized_keys): Permission denied
```
The Tower Template had Privilege Escalation turned on, but that was not enough.

Setting `become: true` on the task was not enough either.

Changing the task to run as the `deploy` user fixed the issue.

### DIFF
--- a/roles/deploy_user/tasks/main.yml
+++ b/roles/deploy_user/tasks/main.yml
@@ -39,6 +39,7 @@
     mode: '0600'
     backup: true
   become: true
+  become_user: deploy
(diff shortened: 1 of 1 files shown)


acozine:
The problem turned out to be with the Tower Credentials. The playbook now runs in Tower as expected.

PR #3860, opened 2023-05-02
[dataspace/oar] move to nginx

kayiwa:
- add task to install nginx
- add configuration and files for nginx


acozine (approved):
Looks good, thanks!

Issue #3840, opened 2023-04-28
manage what adc logs go to datadog


kayiwa:
nginx provides conditional "not a regex" logging. This should lower the amount of logs to datadog saying "OK!"

PR #3862, opened 2023-05-03
[nginxplus/datadog] conditional logging

kayiwa:
- create a "normal" and an "abnormal" map
- conditional logging for loadbalancer


roles/nginxplus/templates/nginx.conf.j2
```diff
@@ -54,7 +54,14 @@ http {
                       '$status $body_bytes_sent "$http_referer" '
                       '"$http_user_agent" "$http_x_forwarded_for"';
 
-    access_log  /var/log/nginx/access.log  main;
+    # access_log  /var/log/nginx/access.log  main;
+    map $status $loggable {
+        ~^[23]  0;
+        default 1;
```
acozine, line 60: It isn't obvious what this does. Can you put in a comment, or a link to the docs (in the PR if not in the code)?

acozine (commented):
Would this close #3840, or is there more to do on that?

roles/nginxplus/templates/nginx.conf.j2
kayiwa: Yes it would close #3840

PR #3878, opened 2023-05-07
[nginxplus TLS] renew TLS certs

kayiwa:
cicognara.org


kayiwa:
run this on both loadbalancers successfully

PR #3880, opened 2023-05-08
[special collections] add a keepalive configuration

### LINKED ISSUE #2603: Persistent "upstream timed out" errors on lib-adc1/2
These errors result in an error page for users, but don't get marked as "failures" in the load balancer for some reason. This results in the load balancer dashboard saying the site's healthy, but users are failing to access content. It seems to effect several sites.

We're going to try increasing the proxy_timeouts for the special collections site first, as that one's a very common problem. You can see the logs for special collections here:

https://app.datadoghq.com/logs?query=%22upstream%20timed%20out%22%20%22special-collections%22&agg_m=count&agg_q=&agg_t=count&index=&sort_m=&sort_t=&stream_sort=time%2Cdesc&top_n=&top_o=&viz=stream&from_ts=1639002007899&to_ts=1639088407899&live=true

kayiwa:
we are not using round robin connections and keepalive is recommended
closes #2603

### DIFF
--- a/roles/nginxplus/files/conf/http/lib-sc-prod.conf
+++ b/roles/nginxplus/files/conf/http/lib-sc-prod.conf
@@ -9,6 +9,7 @@ upstream lib-sc-prod {
           create=$upstream_cookie_lib-sc-prodcookie
           lookup=$cookie_lib-sc-prodcookie
           zone=lib-sc-prodclient_sessions:1m;
+     keepalive 16;
 }
 
 server {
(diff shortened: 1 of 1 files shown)


kayiwa:
Documentation found here http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

tpendragon:
I'm not sure if this would work - this seems like it's a setting for connections that stay alive indefinitely, and this keeps some of those connections active at all times - for things like memcached or websocket connections. Special collections is just a normal request-response site, I'm not sure what this would do when the response eventually returns.

kayiwa:
special collections has moved to a hosted environment.

PR #3883, opened 2023-05-09
[pas] add mailcatcher

### LINKED ISSUE #3593: Set up Mail Catcher on PAS Staging and Dev Roles
Set up mail catcher in these environments so we can test email distribution with sending out real messages.

kayiwa:
this will close #3593


sandbergja (approved):
Thanks, @kayiwa !

PR #3895, opened 2023-05-12
[patch Tuesday] split openbsd playbook

kayiwa:
doas replaced sudo from OpenBSD base. doas does not propagate
the environment variables from the original session. Ansible has a way
to specify environment variables per task. This workaround allows us to
automatically update our BSD servers


leefaisonr:
LGTM. Thanks!

PR #3913, opened 2023-05-17
[Figgy] Cache HEAD requests separately from GET requests.

tpendragon:
The load balancer was converting HEAD requests to GET requests, resulting in Uppy not being able to do resumable uploads for Figgy (it uses a HEAD request to see the state of the previous upload.)

Pulled from https://serverfault.com/questions/530763/nginx-proxy-cache-key-and-head-get-request and tested.


tpendragon:
Deployed.

PR #3918, opened 2023-05-18
Updates keys list for deploy and pulsys users

acozine:
Removes the key for `heaven`, which is now retired.
Removes the key for Taylor.


sandbergja (approved):
Thanks, @acozine !

PR #3919, opened 2023-05-18
More updates to SSH keys

acozine:
Follow-up to #3918.
- Deletes the actual public key for the Heaven service, along with all other references to it.
- Consolidates GitHub keys list.
- Attempts to solve the question of how to put required machine-to-machine SSH keys onto the correct machines. 

Needs testing.


acozine:
Managing "one-off" SSH keys for individual groups or machines is a challenge when we run a single playbook against an entire environment at once, given our current `group_vars` setup.

I've done a small proof-of-concept here for the simplest idea in #3925 - adding a file called `/group_vars/figgy_production.yml` meant those variables automatically loaded into the playbook for machines in the `figgy_production` inventory group. We definitely need a broader discussion about this, though.

acozine:
Closing in favor of #3970.

PR #3923, opened 2023-05-19
adds filtering for Datadog APM logs

acozine:
Building on @bess's great work in #3922 and #3921.

This change filters out:
- health checks from the load balancers to VMs
- sidekiq "got any jobs" checks
from the logging for Application Performance Monitoring on Datadog for all applications that are configured to use it.

### DIFF
--- a/group_vars/bibdata/staging.yml
+++ b/group_vars/bibdata/staging.yml
@@ -141,7 +141,7 @@ postgresql_is_local: false
 
 datadog_api_key: "{{ vault_datadog_key }}"
 datadog_config:
-  tags: "application:bibdata, environment:production, type:webserver"
+  tags: "application:bibdata, environment:staging, type:webserver"
   apm_enabled: "true"
   log_enabled: true
   process_config:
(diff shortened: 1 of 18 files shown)


hectorcorrea (approved):

bess (approved):
🎉

group_vars/bibdata/staging.yml
```diff
@@ -141,14 +141,16 @@ postgresql_is_local: false
```
maxkadel, line 141: Should datadog even be configured on this machine? This doesn't need to block this PR, just confused.

bess: @maxkadel Great point. I am also seeing several RDSS projects that don't need DataDog at all and I'll be removing it via a separate PR.

tpendragon (approved):

maxkadel (approved):
🎉  - confused why there's datadog stuff in bibdata-staging, but that's not the fault of this PR.

carolyncole (approved):

PR #3932, opened 2023-05-24
Add heaven key back

sandbergja:
Deploying via ansible tower still depends on the heaven key.  We should remove the heaven key eventually, but we're re-adding it now so that we can keep deploying.

Previously, if the deploy_user role had been run on a box to remove the heaven key, appdeploy could no longer ssh into the box to deploy.  This affected lib-jobs production, bibdata production, pdc_describe, and possibly figgy.

To get your app deployable again after this is merged: `ansible-playbook playbooks/deploy_user.yml -e runtime_env="[your_env]" --limit=[your_group_of_servers]`

### DIFF
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -90,7 +90,6 @@ library_github_keys:
   - https://github.com/rladdusaw.keys
   - https://github.com/sandbergja.keys
   - https://github.com/sdellis.keys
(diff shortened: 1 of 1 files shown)


christinach (approved):

acozine:
Oops. Sorry about that. I thought we could tidy that away, but I was wrong.

PR #3946, opened 2023-05-27
Remove duplicate vars and rename vars wrongly named as scsb

### LINKED ISSUE #3828: [Bibdata] Remove duplicate vars and rename them
The following vars even though they have scsb in the name they are not related to scsb but to Princeton aws account.

1. Remove lines : https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/production.yml#L72-L73 
```
- name: SCSB_S3_ACCESS_KEY
   value: '{{ vault_scsb_s3_access_key_local }}'
```

2. keep lines https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/production.yml#L104-L105
```
- name: AWS_ACCESS_KEY_ID
   value: "{{ vault_scsb_s3_access_key_local }}"
```

3. In `group_vars/bibdata/production` rename `vault_scsb_s3_access_key_local` to `vault_aws_access_key update the vault

4. Remove https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/production.yml#L74-L75
```
name: SCSB_S3_SECRET_ACCESS_KEY
value: '{{ vault_scsb_s3_secret_access_key_local }}'
```

5. Keep https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/production.yml#L106-L107
```
name: AWS_SECRET_ACCESS_KEY
value: "{{ vault_scsb_s3_secret_access_key_local }}"
```
6. rename vault_scsb_s3_secret_access_key_local to vault_aws_secret_access_key

7. In https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/staging.yml 
and https://github.com/pulibrary/princeton_ansible/blob/cd3368bfdc0468f758c0ddb4ae52c4338b2ab9f7/group_vars/bibdata/qa.yml
rename: vault_scsb_s3_access_key_local to vault_aws_access_key
rename: vault_scsb_s3_secret_access_key_local to vault_aws_secret_access_key

update vault

christinach:
closes #3828


group_vars/bibdata/production.yml
```diff
@@ -104,9 +100,9 @@ rails_app_vars:
   - name: AWS_REGION
     value: "us-east-1"
   - name: AWS_ACCESS_KEY_ID
-    value: "{{ vault_scsb_s3_access_key_local }}"
+    value: "{{ vault_aws_access_key }}"
```
acozine, line 103: Do we have only a single set of AWS access and secret access keys for all of PUL? If not, we might want to include `bibdata` in these var names so we don't get confused across projects.

acozine (commented):
It's great to remove `scsb` from these, but generally, do we need to differentiate these from other projects' keys?

acozine (commented):
Looking at this PR sparked a more general discussion about key management. For this PR, let's change the names of the vars to make it clear that they are coming from the `group_vars/bibdata/vault.yml` file and not from `group_vars/all/`. 

Longer-term, let's look at https://github.com/pulibrary/princeton_ansible/blob/main/architecture-decisions/0004-key-scope and see if we want to revise it. Depending on how the connection to AWS works, we may or may not be able to differentiate keys for each environment.

acozine (approved):
@christinach and I confirmed that we currently use different keys for each project, but not for each environment.

PR looks good, the variable names make the current state clear.

PR #3950, opened 2023-05-30
[pulmap] replace bad google credentials file

hackartisan:
pulibrary/pulmap#1141 is still needed, but this solidifies the earlier quick fix done on these machines, which was lost during a recent playbook run.

closes pulibrary/pulmap#1159

This has been run in production.


acozine:
The CI failures on this PR are related to Molecule itself, not to the change made to the `pulmap` role. The Python stacktrace led us to two possible issues:
https://github.com/ansible-community/molecule/issues/2540
and
https://github.com/ansible-community/molecule/issues/3903
It seems 3903 was the issue here, though I'm not sure why we didn't see this failure on other tests. We should keep this in mind as we do our upgrades to everything - the problem with `ansible-compat` has been fixed in the latest versions of molecule.

Related to #3722 and #3782.

PR #3951, opened 2023-05-31
[nginxplus] renew TLS certs

kayiwa:
- remove expired certificates
- renew certificates using ACME protocol for


kayiwa:
run successfully on both loadbalancers

PR #3952, opened 2023-05-31
Add Isha to Princeton ansible

ishasinha1:


group_vars/all/vars.yml
```diff
@@ -45,6 +46,7 @@ sidekiq_netids:
   - eliotj # Eliot Jordan
   - hc8719 # Hector Correa
   - heberlei # Regine Herberlein
+  - is4684 # Isha Sinha
```
sandbergja, line 49: It looks like this list is only used for pulmap and pulfalight, so I don't think you will need to be on this list.

group_vars/all/vars.yml
```diff
@@ -78,6 +80,7 @@ library_github_keys:
   - https://github.com/escowles.keys
   - https://github.com/hackartisan.keys
   - https://github.com/hectorcorrea.keys
+  - https://github.com/ishasinha1.keys
```
sandbergja, line 83: This list gives access to the `pulsys` user, and you will just need access to the `deploy` user, so I don't think you will need to be on this list either.

sandbergja (changes requested):
Just two changes -- apologies for not catching these before, @ishasinha1 !

sandbergja (approved):
Thanks, @ishasinha1 !

PR #3953, opened 2023-05-31
[postgresql] upgrade to postgresql 15

kayiwa:
- remove all stand-alone replication references
- set default to postgresql 15
- add new vm for postgresql staging
- remove references in tasks to replication


group_vars/postgresql/production.yml
```diff
@@ -1,46 +1,43 @@
...
     priv: "ALL"
     role_attr_flags: "SUPERUSER"
-  - name: replicant
-    password: "{{ vault_replicant_password }}"
-    db: "replicant"
-    priv: "ALL"
-    role_attr_flags: "SUPERUSER"
-postgres_leader: "{{ leader_db_ip }}"
-postgres_standby: "{{ standby_db_ip }}"
+  - name: datadog
```
carolyncole, line 19: This seem like a strange addition?

group_vars/postgresql/production.yml
```diff
@@ -1,46 +1,43 @@
...
-postgres_host: "{{ leader_db_ip }}"
 postgres_port: 5432
 postgres_admin_user: "{{ postgresadmin }}"
 
 datadog_api_key: "{{ vault_datadog_key }}"
 datadog_config:
   log_enabled: true
 datadog_typed_checks:
-  - type: file
+  - type: postgres
```
carolyncole, line 32: 🎉

roles/figgy/molecule/default/verify.yml
```diff
@@ -21,9 +21,3 @@
     failed_when:
       - not figgy_mount.stat.exists
       - not figgy_mount.stat.isreg
-  - name: test for cogeo-mosaic command
```
carolyncole, line 24: Is this no longer needed?

roles/postgresql/tasks/create_db.yml
```diff
@@ -20,7 +20,7 @@
     state: "present"
     login_user: "{{ postgres_admin_user }}"
     owner: "{{ item.owner | default(postgres_admin_user) }}"
-  delegate_to: '{{ postgres_host }}'
+  #  delegate_to: '{{ postgres_host }}'
```
carolyncole, line 23: Is this no longer needed for creating databases from other roles?

carolyncole (commented):
I'm a bit concerned about approving this PR.  Does this mean all the applications in the PR are running jammy?  If not and the servers explode will be be able to get them back up and running on bionic?

A few other random questions.  Not trying to block you...

group_vars/postgresql/production.yml
kayiwa: you need to create a datadog user to monitor postgresql.

roles/postgresql/tasks/create_db.yml
kayiwa: that is passed using the create_db module here. https://github.com/pulibrary/princeton_ansible/blob/5168670a3909a28377bce1dc76d773d32cbe75ba/roles/postgresql/tasks/create_db.yml#L6

roles/figgy/molecule/default/verify.yml
hackartisan: ?

carolyncole (approved):

PR #3963, opened 2023-06-06
[byzantine] add nfs mounts

kayiwa:
the byzantine vms did not have NFS mounts added


acozine:
As we start running multiple VMs for our PHP applications (including the byzantine project), we need to use NFS mounting so all data is written to the shared mount from all VMs.

Issue #3925, opened 2023-05-19
Could we rename group vars files for auto-loading?


acozine:
Ansible loads group_vars files by matching the name of the group to the name of a sub-directory or file in a directory called `group_vars` (see the [docs on inventory](https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html#organizing-host-and-group-variables)).  Our `group_vars` directory has sub-directories, but they are named for the projects, not the inventory groups. This means we have to explicitly load the group variables in playbooks.

This is a challenge if we try to run a playbook on, say, all the staging infrastructure, and different staging groups need different variable values in that playbook. Auto-loading group variables would assign the correct values to each group where they differed (for example, different keys on different groups).

I think we can get group variables to load automatically and still have a separate `vault.yml` file for each project (see https://github.com/pulibrary/princeton_ansible/blob/main/architecture-decisions/0002-group-vars.md). This may involve:
- [ ] creating groups in inventory that match the project names (for example, a `figgy` group with `figgy_production` and `figgy_staging` as children)
- [ ] storing the production and staging group vars in files named `<project>_<env>.yml` to match our group names (for example, `group_vars/figgy_staging.yml` or possibly even `group_vars/figgy/figgy_staging.yml`, though I'm not sure this latter one would work)

PR #3970, opened 2023-06-07
Read "one-off" local SSH deploy keys from group_vars files

acozine:
Supersedes #3919.
Related to #3925.

This PR creates auto-loading `group_vars` files for those groups that need group-specific keys added to the `deploy` user's authorized_keys file.

Tested in check mode. We should be able to run the deploy_keys playbook across our infrastructure from this branch and retain those "one-off" keys on the correct machines.


acozine:
Not happening.

PR #3971, opened 2023-06-08
[svn] add libjobs user

### LINKED ISSUE #3920: create new svn user libjobs_svn
We need a new user so we can send finding aid data from lib_jobs to svn as described in [this ticket](https://github.com/pulibrary/lib_jobs/issues/509). 

The user needs to do the same kinds of things the figgysvn user used to do (authenticate/update/commit).

kayiwa:
alphabetize order of users

this will close #3920


kayiwa:
run the playbook on staging and production

regineheberlein (approved):
Thanks so much @kayiwa !

PR #3981, opened 2023-06-15
removes confluence role, playbook, tests, and vars

acozine:
Removing the role, playbook, test dependency, and group variables for building confluence locally.

We are still running a staging server, however, so the confluence staging needs to stay in the inventory so it participates in patch Tuesdays.


kayiwa (approved):
OK! Thank you @acozine

PR #3982, opened 2023-06-15
Update geoserver fqrn

leefaisonr:


roles/geoserver/tasks/main.yml
```diff
@@ -28,7 +28,7 @@
     state: present
 
 - name: install postgis software
```
VickieKarasic, line 30: Need to add geoserver | before this task (install postgis software)

PR #3984, opened 2023-06-15
make python version general again

acozine:
Our CI system has stopped running on every push. Not sure this is hte problem, but it's a recent change, so we're trying this fix first.


kayiwa:
Sorry, I didn't see this PR when I was working on this https://github.com/pulibrary/princeton_ansible/pull/3988

PR #3985, opened 2023-06-15
Update hr_share fqrn

leefaisonr:


VickieKarasic (changes requested):
Add ansible.builtin.[task] before task names for each role

acozine (commented):
See https://docs.ansible.com/ansible/latest/collections/ for an index of all Ansible collections. 

To check or confirm which collection a module is in, you can go to the command line and type `ansible-doc <module_name>` - for example, `ansible-doc file`. The first line of the docs output will include the fully qualified collection name or FQCN. In this case:

```
% ansible-doc file
> ANSIBLE.BUILTIN.FILE    (/Users/alicia/.local/share/virtualenvs/princeton_ansible-HZ9wli-i/lib/python3.11/site-packages/ansible/modules/file.>

        Set attributes of files, symlinks or directories. Alternatively, remove files, symlinks or directories.
        Many other modules support the same options as the `file' module - including [ansible.builtin.copy],
        [ansible.builtin.template], and [ansible.builtin.assemble]. For Windows targets, use the
        [ansible.windows.win_file] module instead.
. . . 
```

leefaisonr:
See #4021

PR #3991, opened 2023-06-19
[php] use jammy image

kayiwa:


VickieKarasic (approved):
Looks good thanks @kayiwa

PR #3992, opened 2023-06-20
Geoserver2 fqrn

leefaisonr:


roles/geoserver/tasks/main.yml
```diff
@@ -1,48 +1,48 @@
...
+- name: geoserver | Create symlinks
+  ansible.builtin.file:
     src: '{{ item.src }}'
     dest: '{{ item.link }}'
     owner: '{{ deploy_user }}'
     group: '{{ deploy_user }}'
     state: 'link'
     force: true
-  with_items:
+  loops:
```
acozine, line 45:
```suggestion
  loop:
```

acozine (commented):
Lines 18 and 22 also need to change - we need FQCNs there as well.

roles/geoserver/tasks/main.yml
```diff
@@ -1,48 +1,48 @@
 ---
 - name: geoserver | stop tomcat server
-  service:
+  apt.builtin.service:
```
acozine, line 3:
```suggestion
  ansible.builtin.service:
```

PR #4001, opened 2023-06-21
Added proxy pass rule for sales catalogs

rladdusaw:


roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
```diff
@@ -47,6 +47,11 @@
         proxy_pass https://firestone-renovations-staging.princeton.edu/versailles/;
     }
 
+    rewrite ^/libraries/marquand/resources/sale(.*)$ /marquand_catalogs/$1 redirect;
```
kevinreiss, line 50: The current URL is https://library.princeton.edu/marquand/resources/sales-catalogs the new root URL is just /marquand_catalogs. We want to capture any bookmarked browsing patterns and send them to the new home, however I'm not sure we need the $1 at the end of the redirect. Don't we just want everything to go to "/marquand_catalogs as a landing page no matter what the params are? Or does new location attempt to do anything with any params passed over?

kevinreiss (changes requested):

roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf
rladdusaw: The WAF is blocking requests to the redirected url. We will either need to turn it off or determine why it's being blocked.

maxkadel (approved):

kevinreiss (approved):

PR #4004, opened 2023-06-21
[oawaiver] use jammy image

kayiwa:
- add linting to OJS role. Make the role user fqrn
- make oawaiver use jammy image


kayiwa:
closed by this https://github.com/pulibrary/princeton_ansible/pull/4260

PR #4015, opened 2023-06-22
[pgbouncer] remove role

kayiwa:
remove references to it


acozine (commented):
If we are no longer using `pgbouncer` to manage connections to the postgres database, then +1 to this.

PR #4016, opened 2023-06-22
[postgres] remove unused roles

kayiwa:
- remove the `psql` role which was there for confluence as a standalone
- ~~remove the `postgres` role into a playbook~~
- move the `add-readonly-user` tasks from the `postgres` role into the `postgresql` role
- remove the `postgres` role
- remove the `figgy` role's dependency on the `postgres` role


acozine:
CI is broken because the `figgy` role depends on the `postgresql` role - see https://github.com/pulibrary/princeton_ansible/blob/main/roles/figgy/meta/main.yml#L27.

Is figgy the only reason we need the new "create a read-only user" playbook? If this is a pattern we want to move away from, we could incorporate those tasks into the figgy role as `roles/figgy/tasks/readonly_pg_user.yml` instead of having a playbook for them.

acozine:
@hackartisan @tpendragon @eliotjordan can we delete the figgy dependency on the obsolete `postgres` role? Also, when/why do we create readonly database users for figgy?

roles/figgy/meta/main.yml
```diff
@@ -24,4 +24,3 @@ dependencies:
   - role: "rails_app"
   - role: "ffmpeg"
   - role: "extra_path"
-  - role: "postgres"
```
acozine, line 27: Do we need to add a dependency on the postgresql role (now that it's the only db-related role for postgres)?

hackartisan (approved):
Looks good from figgy's perspective

PR #4017, opened 2023-06-22
[Redis] Collect metrics for datadog

tpendragon:
There were a couple config options missing for datadog to collect metrics on on our redis infrastructure.


tpendragon:
This only does lib-redis-prod1 which looks to be unused. It looks like lib-redis doesn't have datadog.

PR #4022, opened 2023-06-23
Update figgy_pubsub_worker role with fqrn

VickieKarasic:
Co-authored-by: Robert-Anthony Lee-Faison
<leefaisonr@users.noreply.github.com>

### DIFF
--- a/roles/figgy_pubsub_worker/tasks/main.yml
+++ b/roles/figgy_pubsub_worker/tasks/main.yml
@@ -1,7 +1,7 @@
 ---
 ## Google Cloud Worker
 - name: Install startup script for PubSub
-  template:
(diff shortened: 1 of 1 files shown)


VickieKarasic:
We had questions with updating the `with items` part of the task with `loop` instead and want to verify that we did that correctly. Thanks!

acozine (approved):
Looks good. In most places where we've used `with_items`, we pass a list of items - nothing fancy. You can change these to `loop` just like you could change `with_list` to `loop`. See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#migrating-from-with-x-to-loop for more details.

VickieKarasic:
Thanks @acozine!

PR #4024, opened 2023-06-23
update memcached role with fqrn

VickieKarasic:


roles/memcached/tasks/main.yml
```diff
@@ -1,21 +1,21 @@
...
 - name: Define memcached_user.
-  set_fact:
+  ansible.builtin.set_fact:
     memcached_user: "{{ __memcached_user }}"
   when: memcached_user is not defined
 
 # Setup/install tasks.
 - name: Install Memcached.
-  package: name=memcached state=present
+  ansible.builtin.package: name=memcached state=present
```
acozine, line 13: While you're updating this role, it would be great to use "new" syntax for all tasks. This isn't something we've discussed before . . . but if you look at the next task in this file, you'll see that all the parameters are on separate lines below the module name, indented two spaces, and use `:` instead of `=` as the separator. This is "modern" (since I think Ansible 2.0, maybe earlier) syntax, and I find it more readable. For this task, the update would look like this:
```suggestion
  ansible.builtin.package:
    name: memcached
    state: present
```

roles/memcached/tasks/main.yml
```diff
@@ -24,4 +24,4 @@
   notify: restart memcached
 
 - name: Ensure Memcached is started and set to run on startup.
-  service: name=memcached state=started enabled=yes
+  ansible.builtin.service: name=memcached state=started enabled=yes
```
acozine, line 27: Same thing here. Move the param values (name, state, enabled) onto separate lines, all indented 2 spaces from the module name.

acozine (commented):
Looks good - you made all the changes we talked about. There's one more bit of modernization we could do on this file - see details below.

roles/memcached/tasks/main.yml
VickieKarasic: Will do - thanks!

acozine (approved):
Fantastic, thanks @VickieKarasic.

roles/memcached/tasks/main.yml
```diff
@@ -1,21 +1,23 @@
...
 - name: Define memcached_user.
-  set_fact:
+  ansible.builtin.set_fact:
     memcached_user: "{{ __memcached_user }}"
   when: memcached_user is not defined
 
 # Setup/install tasks.
 - name: Install Memcached.
-  package: name=memcached state=present
+  ansible.builtin.package:
```
kayiwa, line 13: we only use ubuntu. Can this be `apt`

VickieKarasic: See #4032 - tried to fix this. Thanks!

PR #4037, opened 2023-06-29
Updates inventory, esp. orphans

acozine:
We have quite a few mysterious entries in our inventory. Can we clarify what these machines are, what they do, and how important they are?

This PR begins by noting the current use of libserv144.

We might also consider adding "hosted on" information to 
https://lib-confluence.princeton.edu/display/IT/Service+Catalog+Data as part of this effort.


kayiwa (approved):
OK! Thanks @acozine

Issue #4039, opened 2023-06-30
[orangelight] Do not create Blacklight cron jobs


maxkadel:
Having the cron job on all the servers leads to database locks (see [orangelight issue](https://github.com/pulibrary/orangelight/issues/3610)). We will use the Whenever gem to run these jobs on a single server, so we don't run into database locks, and remove the cron job.

### Manually remove cron job from following servers
- [x] catalog-staging1.princeton.edu
- [x] catalog-staging2.princeton.edu
- [x] catalog-qa1.princeton.edu
- [x] catalog1.princeton.edu
- [x] catalog2.princeton.edu
- [x] catalog3.princeton.edu
- [x] catalog-indexer1.princeton.edu
- [x] catalog-indexer2.princeton.edu
- [x] catalog-indexer3.princeton.edu

maxkadel:
Resolved by https://github.com/pulibrary/princeton_ansible/pull/4040. Old cron job removed from servers 6/30/2023

PR #4040, opened 2023-06-30
[orangelight] Use rails_app role not blacklight_app

maxkadel:
- This allows us to skip the creation of a cron job that was causing issues.
- In order for it to take full effect, you need to ssh onto each box and delete the cron job manually

```
ssh deploy@one-of-the-servers
crontab -e
# Use nano or vim to delete the blacklight:delete_old_searches cron job and save
# to verify the lines have been deleted:
crontab -l
```

Connected to #4039 
Connected to https://github.com/pulibrary/orangelight/issues/3610

### DIFF
--- a/roles/orangelight/meta/main.yml
+++ b/roles/orangelight/meta/main.yml
@@ -2,7 +2,7 @@
 galaxy_info:
   role_name: orangelight
   company: Princeton University Library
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
🦄

PR #4042, opened 2023-06-30
[orangelight] Enhance nginx monitoring via datadog

maxkadel:
Use typed checks to make sure datadog uses nginx status page

### DIFF
--- a/group_vars/orangelight/production.yml
+++ b/group_vars/orangelight/production.yml
@@ -38,24 +38,27 @@ datadog_config:
       orangelight|rack.request: 1
     filter_tags:
       reject: ["http.useragent:nginx/1.23.4 (health check)", "operation:heartbeat", "operation:job_fetch", "operation:scheduled_push", "operation:scheduled_poller_wait"]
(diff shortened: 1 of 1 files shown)


hackartisan:
This looks interesting, if this is a new best practice are there also documentation updates we could make?

maxkadel:
@hackartisan  - where do you think an appropriate place to document would be? It does seem to give more / better nginx insight, which is helpful, especially for the types of issues we've been seeing on Orangelight. [See the NGINX dashboard - now with Orangelight data!](https://app.datadoghq.com/dash/integration/21/nginx---overview?from_ts=1689164831182&to_ts=1689168431182&live=true)

hackartisan:
@maxkadel there are some playbook examples in https://github.com/pulibrary/princeton_ansible/blob/main/roles/datadog/README.md -- maybe even including this work already? If not, that might be a good place to document?

maxkadel:
@hackartisan  - thanks, yes, it looks like it's included there.

PR #4052, opened 2023-07-06
Datadog role example gathers the correct nginx information

sandbergja:
Update the datadog role's documentation, as recommended by @hackartisan in https://github.com/pulibrary/princeton_ansible/pull/4042#issuecomment-1618314237


hackartisan (approved):
thanks!

PR #4061, opened 2023-07-07
[recap] add jammy vm to loadbalancer

kayiwa:
point vms to mariadb-dev with running on jammy


kayiwa:
will need to review this first. https://github.com/pulibrary/recap/pull/248

maxkadel (approved):

PR #4062, opened 2023-07-07
[solr] change our upstream algorithm

kayiwa:
the loadbalancer keeps favoring one of our boxes because we've been
defaulting on round robin. we will use the least_time instead

load balancing method where a request is passed to the server with the least average response time and least number of active connections, taking into account weights of servers
incomplete requests are also taken into account with the inflight
parameter


jrgriffiniii (approved):

jrgriffiniii:
https://www.nginx.com/blog/choosing-nginx-plus-load-balancing-techniques/ confirms to me that this is valid

PR #4065, opened 2023-07-10
[manifold] add nginx config

kayiwa:


acozine:
Manifold config still needs work, but this is a good step forward!

PR #4081, opened 2023-07-25
[ansible] update cryptography

kayiwa:
- add rapid7 ip to the allow list
- update cryptography library


acozine:
duplicate of #4320

PR #4087, opened 2023-07-27
[lib_jobs] Do not log SVN secrets

maxkadel:
Previously the task was outputting each user's secret to the console, this makes it so that it does not log those secrets.


acozine:
Thanks @maxkadel !

Issue #2227, opened 2021-04-27
[nginx] Nginxplus should pass through the error status code provided by the upstream server instead of a generic 200.


hackartisan:
During an outage of many of our machines, nginx appears to have returned the stock error page with a 200 status. The underlying error appears to be: 

```
2021/04/26 15:03:09 [error] 22494#22494: *12797085 no live upstreams while connecting to upstream, client: 128.112.202.197, server: findingaids-beta.princeton.edu, request: "GET /catalog/CO2_trunks_E-B+.json HTTP/1.1", upstream: "http://pulfalight-prod/catalog/CO2_trunks_E-B+.json", host: "findingaids-beta.princeton.edu"
```

This should return as a non-200. If it had returned a non-200 figgy would have appropriately interpreted it as an error (see https://github.com/pulibrary/figgy/blob/186fadca59e3dfea86068d53cfbb669334373d57/app/services/pul_metadata_services/client.rb#L26) However, it tried to parse it as JSON. Relevant honeybadger issue: https://app.honeybadger.io/projects/53391/faults/79044416

some datadog logs: https://app.datadoghq.com/logs?context_event=AQAAAXkOtKOwCb6NGQAAAABBWGtPdEt5TUFBQVBYR09RaVVYbWJnQVI&event=AQAAAXkOtJ_ICb6NDQAAAABBWGtPdEt5TUFBQVBYR09RaVVYbWJnQUY&from_ts=1618931463021&index=&live=false&messageDisplay=inline&query=host%3Alib-adc1%20service%3Anginxplus%20json&stream_sort=desc&to_ts=1619449393000&viz=

rlskoeser:
I'm seeing similar behavior with CDH staging sites that are IP restricted, and suspect it may also be happening with errors being served to google and bing search bots (google search console is reporting a suspiciously large number of "soft 404s" for the CDH website, meaning content is not found but it was not served with a 404 or other status).

You can easily duplicate the problem by using curl on an IP restricted site while not on VPN, e.g.:
> curl -I https://test-geniza.cdh.princeton.edu
```
HTTP/2 200
server: nginx/1.19.10
date: Mon, 07 Jun 2021 16:43:00 GMT
content-type: text/html
content-length: 2489
last-modified: Fri, 04 Jun 2021 20:39:29 GMT
etag: "60ba8f81-9b9"
accept-ranges: bytes
```

I see that both prod and staging maintenance.conf `@maintenance` locations include a `=502` directive.

[...]

PR #4089, opened 2023-07-28
Pass HTML error codes for abid-staging through the LB

acozine:
This is part of our work on #2227. It still does not work (though we had one glorious moment where we did get a 500 error passed through). For the benefit of Future Us, we memorialized our attempted changes on a branch. Obviously this will stay a draft PR until it DOES work.


acozine:
Today we got this partially working. The necessary fixes/changes include:
1. add the `proxy_intercept_errors on;` directive in the `location` block of the site config (our test site config is `abid-staging`)
2. remove or comment out the inclusion of `/etc/nginx/conf.d/templates/staging-maintenance.conf` - this was active and working but uses the old syntax, and you can't have two `error_page` definitions 
3. include `/etc/nginx/conf.d/templates/errors.conf` - not sure if this should be in the `location` block or in the `server` block, right now it's in both
4. on the load balancer, copy `/var/local/www/default/index.html` (which is the deployed file created from the `roles/nginx/files/www/error.html` file) to `/var/local/www/default/error.html` (which is what the errors.conf file is looking for), with the correct permissions
5. add a specific error code you want to return, using `=XXX`, after the list of error codes to capture, in `errors.conf` (on the LB I set it to `=503`, in the git branch I've set it to `=500`)

With these changes, I consistently get the right results:
 - at the command line:
```
alicia@PULaptop ~ % curl -I https://abid-staging.princeton.edu
HTTP/2 503 
server: nginx/1.23.4
date: Mon, 31 Jul 2023 22:51:53 GMT
content-type: text/html
```
- in the browser:
Screenshot 2023-07-31 at 6 05 52 PM

Still not sure how to get the LB to pass "whatever the upstream error code is", but we're closer!

hackartisan:
Awesome, look forward to hearing more about this. For step 5, does pass-through work if you use the plain `=` instead of, e.g., `=503`?

roles/nginxplus/files/conf/http/abid-staging.conf
kayiwa, line 1: 🎉

kayiwa (approved):
OK! Thanks @acozine

Issue #4094, opened 2023-07-31
Manage certs transition


acozine:
These certs have already moved to the load balancer. Revoke these in ServiceNow if they still exist there:

- [ ] abid
- [ ] datacommons
- [ ] iiif-staging
- [ ] lib-sc-prod
- [ ] pdc-describe-prod
- [ ] abid-staging
- [ ] figgy
- [ ] imagecat-prod
- [ ] lib-sc-staging
- [ ] pdc-describe-staging
- [ ] annotations-staging
- [ ] figgy-staging
- [ ] imagecat-staging
- [ ] locator-staging
- [ ] pulfa
- [ ] approvals-staging
- [ ] findingaids-qa
- [ ] lib-adc
- [ ] manifold-staging
- [ ] rbsc
- [ ] arks-redirect
- [ ] findingaids-staging
- [ ] lib-jobs-staging
- [ ] maps
- [ ] bibdata
- [ ] geoserver
- [ ] lib-monitor
- [ ] oawaiver-staging
- [ ] recap-staging
- [ ] bibdata-qa
- [ ] geoserver-staging
- [ ] library-qa
- [ ] openbooks-staging
- [ ] researchdata
- [ ] bibdata-staging
- [ ] iiif
- [ ] library-staging
- [ ] openpublishing
- [ ] researchdata-staging

Move these certs to the load balancer:
1. Revoke the old certs in ServiceNow
2. Run the incommon_certbot playbook to create the letsencrypt certs
4. Create a PR to princeton_ansible removing the old certs and updating the nginx config
5. Run the nginx playbook to deploy the letsencrypt certs

*** delete all interm.pem files!

- [ ] annotations_princeton_edu
- [ ] ansible-tower_princeton_edu
- [ ] appdeploy_princeton_edu
- [ ] approvals_princeton_edu
- [ ] bibdata-alma-staging_princeton_edu - can we delete this altogether?
- [ ] bibdata-alma_princeton_edu - can we delete this?
- [ ] bluemountain_princeton_edu
- [ ] byzantine-tsp-prod_princeton_edu
- [ ] byzantine-tsp-staging_princeton_edu
- [ ] catalog-alma-qa_princeton_edu - can we delete this altogether?
- [ ] catalog-alma_princeton_edu - can we delete this altogether?
- [ ] catalog-prod_princeton_edu - can we delete this altogether?
- [ ] catalog-qa_princeton_edu
- [ ] catalog-staging_princeton_edu
[...]

PR #4095, opened 2023-07-31
Delete old revoked certificates

VickieKarasic:
Point config files to acme-generated certs for annotations.princeton.edu


VickieKarasic:
Relates to #4094

PR #4097, opened 2023-07-31
annotations-dev is retired, remove certs and conf

acozine:
Partial fix for #4094.

Removes config, cert and key for annotations-dev site. The machine has already been decommissioned and the site no longer exists.


kayiwa (approved):
OK! Thank you @acozine

PR #4103, opened 2023-08-01
[pdc-describe] update postgresql version

### LINKED ISSUE #4092: migrate to postgresql 15
upgrade databases for pdc-describe and pdc-discovery

- [ ] update pdc-discovery-staging
- [ ] update pdc_describe-staging

kayiwa:
upgrade postgresql version
this will close #4092

### DIFF
--- a/group_vars/pdc_describe/main.yml
+++ b/group_vars/pdc_describe/main.yml
@@ -1,5 +1,5 @@
 ---
-postgres_version: 13
+postgres_version: 15
 postgresql_is_local: false
 postgres_port: "5432"
 postgres_admin_user: 'postgres'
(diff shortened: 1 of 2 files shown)


group_vars/pdc_describe/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-postgres_version: 13
+postgres_version: 15
```
acozine, line 2: Won't this change production as well as staging? I don't think we're ready to change both.

acozine (commented):
The ticket says to upgrade pdc_describe_staging and pdc_discovery_staging. Has pdc-discovery already been done? Also, do we need to have separate values for the psql version in staging and prod (see below)?

acozine:
FYI for folks upgrading to 15 in development environments, see https://github.com/pulibrary/figgy/pull/5918#issuecomment-1660571628 for a gotcha (with fix).

group_vars/pdc_describe/main.yml
kayiwa: this only updates the client. This https://github.com/pulibrary/princeton_ansible/issues/4105 will document how this role functions.

acozine:
The values got updated for pdc_discovery in #4100.

PR #4129, opened 2023-08-09
[Figgy] Upgrade VIPs

tpendragon:
This makes sure it uninstalls an old version of vips if it exists, and sets the default higher.

New vips has faster PDF processing. Work towards pulibrary/figgy#5937


tpendragon:
Deployed.

PR #4137, opened 2023-08-10
Update the value for production env var scsb_s3_bucket_name

### LINKED ISSUE #4136: [Lib-jobs] Update scsb_s3_bucket_name
Update production value for var scsb_s3_bucket_name
It should be 'scsb-prod'

christinach:
closes #4136


christinach:
Ran the playbook in production to update the value of the s3 bucket.

PR #4145, opened 2023-08-14
increase serial number to allow playbook to get through inventory

acozine:
The Patch Tuesday job in Tower has been failing on several VMs. With the `serial` set to `3`, three unreachable or failing VMs will make the entire playbook fail and some VMs will not get updates at all.

This PR creates a variable called `concurrent_vms` for the `serial` value and sets a default of `5`. In Tower, we can set the template to `Prompt on launch` for that variable value when we need to override it.


kayiwa (approved):
OK! Thank you @acozine

PR #4146, opened 2023-08-14
[abid] move loadbalancer to abid-staging2

kayiwa:
comment out bionic VM on loadbalancer and inventory


acozine:
The old bionic VM (abid-staging1) still exists, but the load balancer is only routing traffic to the new jammy VM (abid-staging2).

Going forward our plan is to having one Staging VM running at all times for all projects, and to have a second Staging VM built for each project, but to power it down except while devs are actively working on the project.

PR #4163, opened 2023-08-21
[nginx] add config stanza for logo

### LINKED ISSUE #4134: [nginx] PUL logo does not load on error pages
While testing #4122, we noticed that the PUL logo is not displaying in our custom error pages. Those pages are served up from the load balancer's default site. It's possible, however, that nginx is still proxying the request for the image. Perhaps we could test this theory by copying the image into the root directory for the catalog site, then triggering rate limiting - if the image reappears, that suggests that nginx is looking in the proxied site for it.

We do have alt text for the image, and otherwise our custom error handling is working.

Image details:
In the repo: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/pul-logo-new.svg
On the server: `/var/local/www/default/pul-logo-new.svg`

Default site config:
In the repo: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/001-default.conf
On the server: `/etc/nginx/conf.d/001-default.conf`

acozine:
Closes #4134.

No idea how the old config actually worked. But this seems to do the trick.

Keeps the `error` and `ratelimit` locations private, but adds a non-private location stanza for the image.

### DIFF
--- a/roles/nginxplus/files/conf/http/templates/errors.conf
+++ b/roles/nginxplus/files/conf/http/templates/errors.conf
@@ -13,3 +13,8 @@ error_page 429 /ratelimit.html;
         internal;
         root /var/local/www/default;
     }
+
+    location = /pul-logo-new.svg {
+        ssi on;
+        root /var/local/www/default;
+    }


sandbergja (approved):
🎉  Thanks, @acozine

PR #4172, opened 2023-08-24
[libsftp] pause the system-ldap playbook

### LINKED ISSUE #3774: system_ldap prompt on role
can we add a prompt on the lib-sftp role which will allow us to only run the playbook once.

the prompt will allow us to create the steps listed in the [README](https://github.com/pulibrary/princeton_ansible/blob/main/roles/system_ldap/README.md)

acozine:
Closes #3774.

Needs testing, but this should pause the playbook that builds the libsftp box while the manual steps from the README are run, then resume once the user acknowledges the message.

### DIFF
--- a/roles/system_ldap/tasks/main.yml
+++ b/roles/system_ldap/tasks/main.yml
@@ -35,6 +35,12 @@
     line: '# AllowUsers pulsys'
   notify: restart sshd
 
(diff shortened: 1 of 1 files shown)


acozine:
Closing as obsolete. We replaced the system-ldap role with the new sssd_ad role.

PR #4179, opened 2023-08-24
add CDH error handling on test sites

acozine:
Closes https://github.com/Princeton-CDH/cdh-ansible/issues/60

This is the start of custom error handling for CDH sites on the library load balancers.

Remaining work to be done:
- [ ] ~~confirm that we want all these sites to pass upstream errors and use custom error pages~~
- [x] add the CDH logo (confirm that it's available as `cdh-logo.svg`)
- [x] confirm what links should go on CDH error pages: email to a CDH support list? link to CDH home page? keep library links?


acozine:
Note that this will return that "Are you on the VPN?" page for all `403 Forbidden` errors for these sites. We may want to add more info if that is likely to happen for other reasons beyond "Nope, i was not on the VPN."

rlskoeser:
Thank you for working on it. General approach seems good to me.

My initial thought was to keep the PUL logo (as shown on other error pages - since this is running on PUL infrastructure) but also add a CDH logo, maybe underneath it - as a way of signalling to people whether or not they are in the right place (if they intended to access a CDH site).  Does that makes sense?

I'm not sure what links should be there - we don't have CDH specific support lists currently; my preference would be to use the existing lsupport mailing address if we can (as you have here).

A couple variants of the CDH logo are available here https://github.com/Princeton-CDH/cdh-web/tree/main/sitemedia/img/cdh-logo-assets 
There are more variants in a google drive folder, but I think the default one should work here.  Do you want to have a local copy because we don't want to introduce any possible errors while serving the error page?

My main question about the functionality is how can we test it and how do we make sure this doesn't override app-specific error pages.  We typically implement custom 404 and 500 pages; some apps may have an app-specific 403 that should not be handled (e.g. for REST API type requests). Can a 403 triggered by nginxplus IP restriction be differentiated from an app 403?

Could try this out on CDH staging sites first before rolling it out to the production sites?

acozine:
@rlskoeser thanks for the review - I updated the branch in light of your comments. The current changes:
- add the CDH logo and properly reference it
- reinstate the PUL logo above the CDH one
- keep the lsupport link
- leave 404 and 500 errors out of the CDH-errors.conf list - I hope that this will preserve the application's custom pages for those errors, but I have not tested this
- leave production site configs unchanged

I do not know of a way to differentiate 403 errors by cause, but we could experiment - is there a staging site that needs this special handling?

rlskoeser (approved):
These revisions look good! I think this set of changes would be sufficient to test out on CDH staging sites and see if we need any adjustments before applying to production sites.

rlskoeser:
Thanks for the revisions!

I've been thinking about how to test this; I can add logic to force a 500 error on a staging site, so we can easily check 404 and 500 behavior. We can think about how to trigger other errors for testing purposes (e.g., shutting down nginx on the app server).  

I'll have to check the application code bases to find out which ones use or generate any other error codes besides these two. Once I have that list, I should be able to figure out how we can test them.

> I do not know of a way to differentiate 403 errors by cause, but we could experiment - is there a staging site that needs this special handling?

I think a 403 is possible in an internal API used for annotations in the Geniza code base; we'd want the user to see the error message if they're not on vpn but get the app-specific 403 if they are on vpn. I'll think about how to test this when I look for specifics for possible error codes.

acozine:
Tested and approved for dev/test/staging sites today. We will continue to refine.

PR #4180, opened 2023-08-24
moves lae-staging cert to ACME system

acozine:
Partial fix for #4094.
Related to #4178. 

The certificate for lae-staging had expired. This PR moves cert management for this site over to our new ACME system and removes the old key and cert.


kayiwa (approved):
OK! Thanks @acozine

PR #4183, opened 2023-08-28
[nginxplus] repec-prod TLS certs

kayiwa:
- remove all revoked certificates
- point repec-prod.princeton.edu name to ACME generated certificates


christinach (approved):
Thanks @kayiwa

Issue #4018, opened 2023-06-23
Failing/unreachable inventory hosts in staging group on Tower


acozine:
If I run the key-update playbook from Tower, I get 5 unreachable hosts:

```
fatal: [prds-dataspace-staging-endpoint1]: UNREACHABLE! => {. . . Permission denied (publickey). . .}
fatal: [staging-bastion.pulcloud.io]: UNREACHABLE! => {. . . Could not resolve hostname staging-bastion.pulcloud.io: Name or service not known . . . }
fatal: [smoketest.docnow.io]: UNREACHABLE! => { . . . Permission denied (publickey). . . }
fatal: [bastion-staging.pulcloud.io]: UNREACHABLE! => {. . . Connection timed out . . .}
fatal: [oawaiver-staging1.princeton.edu]: UNREACHABLE! => {. . . Connection timed out . . .}
```

We should either configure working connections for these machines or move them out of the `staging` group in our inventory.

acozine:
Related to #3897.

acozine:
As of today we have some VMs that fail the Patch Tuesday playbook and more that are unreachable:

Failing:

```
[...]

PR #4184, opened 2023-08-28
Exclude unreachable cloud boxes

acozine:
We've seen persistent `unreachable` failures on our cloud resources when running the "Patch Tuesday" playbook on Ansible Tower. This is one potential fix.

This PR:
- adds env-level (staging, production) groups for boxes on Google Cloud (gcp) - this is likely to be a handy change
- excludes gcp and aws boxes from the apt update playbook - this begs the question, how else would be update the apt packages on our cloud resources?

Related to #4018.


acozine:
Discussed today in Open House. This does not seem like the best approach.

PR #4195, opened 2023-08-30
[inventory] adds group for LBs

acozine:
The pulsys user playbook was not running on the load balancers. I suspect neither was the apt update playbook.

Root cause was that those playbooks run by environment, as defined by the files in `inventory/by_environment`, like `inventory/by_environment/production.yml`. The load balancers were not in the production group. 

This PR adds a group for the load balancers in the `_orphans` file and adds the `load_balancers` group to 'production' meta-group.

### DIFF
--- a/inventory/by_environment/production
+++ b/inventory/by_environment/production
@@ -29,7 +29,6 @@ libstatic_production
 lib_fs_production
 lib_svn_production
 libwww_production
-load_balancers
 loadtest_production
 locator_production
 lockers_and_study_spaces_production
--- a/inventory/by_environment/production
+++ b/inventory/by_environment/production
@@ -29,6 +29,7 @@ libstatic_production
 lib_fs_production
 lib_svn_production
 libwww_production
+load_balancers
 loadtest_production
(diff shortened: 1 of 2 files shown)


acozine:
Discussed today in Ansible Open House. We want to include the load balancers when we run the key-update playbooks, but not when we run the apt-update playbook.

Maybe change the key-update playbooks to include the load balancer group, but not add that group to the `production` group? Or possibly create a `danger` group? Investigate.

acozine:
Keep the `load balancers` group, remove it from the `production` group of groups.

inventory/by_environment/production
```diff
@@ -29,6 +29,7 @@ libstatic_production
 lib_fs_production
 lib_svn_production
 libwww_production
+load_balancers
```
carolyncole, line 32: Suggesting removing this change...
```suggestion
```

carolyncole (approved):

kayiwa (approved):
OK! Thanks @acozine

Issue #4194, opened 2023-08-30
datacommons.princeton.edu/describe should point to pdc-describe-prod/describe


bess:
... instead it redirects to pdc-discovery-prod/discovery

- [x] Figure out why
- [x] Fix it
- [x] Set up honeybadger uptime monitoring at the URL so if it breaks again we'll know

PR #4198, opened 2023-08-31
Use full URLs in datacommons nginxplus config

bess:
Ref https://github.com/pulibrary/princeton_ansible/issues/4194


kayiwa (approved):
OK! Thanks @bess

PR #4206, opened 2023-09-04
I4193 backup to gce

kayiwa:


acozine:
The restic role installs the [restic package](https://restic.net/), which manages backups to Google Cloud Environment.

We need to add the new geoserver databases to the list of servers for this change.

PR #4224, opened 2023-09-06
[DSS] Send all traffic to dss-prod1 during jammy upgrade

sandbergja:


kayiwa (approved):
OK! thanks @sandbergja

PR #4231, opened 2023-09-07
Use a template for our simplest nginx configs

acozine:
DRAFT ONLY - includes changes for testing that must not be merged.

This approach would convert a majority of our nginx config files to variables that run through a template. Currently addresses our two most common use cases:
 - public/production sites with a single `location` definition
 - private/staging sites with a single `location` definition that are restricted to the VPN


acozine:
A quick look at a few config files uncovers a few differences among sites. With these changes:
- [ ] add a var for the HTTP [load-balancing technique](https://www.nginx.com/blog/choosing-nginx-plus-load-balancing-techniques/) (see annotations-prod)
- [ ] add a section for enabling/disabling `app_protect` (see abid-staging)
- [ ] allow more than one included `restrict` file (see bibdata-qa)
- [ ] regularize an `ansible-managed` comment
we can try using the template for these sites:
- [ ] abid
- [ ] annotations
- [ ] ansible-tower
- [ ] approvals
- [ ] bibdata
- [ ] byzantine-tsp
- [ ] cicognara
- [ ] eal-apps
. . . and probably more.

Once the changes are ready, we can put the vars in place and create the templated files in `/tmp/`, then diff the files against the actual nginxplus config files to identify any changes/drift.

acozine:
Add a variable for the health check page (see Drupal sites for examples).

acozine:
Current CI failure is:
```
  TASK [nginxplus : Setup | Upload NGINX HTTP Configuration Files] ***************
  fatal: [instance]: FAILED! => {"msg": "Invalid data passed to 'loop', it requires a list, got this instead: conf/http/*.conf. Hint: If you passed a list/dict of just one element, try adding wantlist=True to your lookup invocation or use q/query instead of lookup."}
```

acozine:
Closing in favor of #4397

PR #4256, opened 2023-09-12
i4255: Update hostname if it doesn't match the inventory hostname

### LINKED ISSUE #4255: Update the hostname using ansible
Instead of changing the hostname manually: `sudo hostnamectl set-hostname <hostname>`

Do sth like:  https://github.com/codylane/ansible-playbook-change-hostname/blob/master/change_hostname.yaml#L5-L10

sandbergja:
closes #4255

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -1,4 +1,10 @@
 ---
+- name: common | Confirm that hostname matches inventory
+  ansible.builtin.hostname:
+    name: "{{ inventory_hostname }}"
+  when:
+    "ansible_hostname not in inventory_hostname"
+  tags: "update_hostname"
 - name: common | Add apt HTTPS capabilities.
   ansible.builtin.apt:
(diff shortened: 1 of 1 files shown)


kayiwa:
successfully ran this with `lib-mudd-prod2`

roles/common/tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: common | Confirm that hostname matches inventory
+  ansible.builtin.hostname:
+    name: "{{ inventory_hostname }}"
```
acozine, line 4: Is this going to add the `.princeton.edu` suffix to the hostname? If so, it would make for a long shell prompt.

sandbergja: It add .princeton.edu to /etc/hostname, but not to the shell prompt thankfully!  I added the steps that @christinach and I used to exercise this branch to this PR's description, in case you'd like to try it out.

Issue #3476, opened 2022-11-30
Node 18 binaries are not compatible with Ubuntu bionic


sandbergja:
The nodejs role uses the pre-built binary tarballs from the nodejs web site.  As of Node 18, [these binaries assume glibc 2.28 or later](https://nodejs.org/de/blog/announcements/v18-release-announce/#toolchain-and-compiler-upgrades), which bionic does not have (focal and higher do).

As a result, we can't use the nodejs role to upgrade to any node LTS after node 16.

### Steps to replicate

1. Edit group vars for an application to have `desired_nodejs_version: "v18.12.1"`
2. Run the playbook on a staging box
3. SSH into the box
4. Run `node -v`
5. Note that, instead of getting the expected node version, you get an error message:

``` 
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)`
```

### Create a new Staging VM
- [ ] Log into VSphere and find the jammy template
  * Right Click and Select New VM from the template
  * If  repec-staging1 one exists create a new VM called repec-staging2 and save it in discovered machines
- [ ] Create firewall Rules for your new VM
- [ ] Modify the node values at `group_vars/<project_name>/staging.yml` sometimes `group_vars/<project_name>/common.yml` or `group_vars/<project_name>/main.yml` to point to v18.17.0
- [ ] Modify the postgresql values `group_vars/<project_name>/staging.yml` to point to `lib-postgres-staging1` for the database and the `postgresql_version` to 15
- [ ] Modify `inventory/all_projects/<projectnam>` to comment out bionic vm and add new vm
- [ ] Run a database dump/restore from `lib-postgres-staging3` to `lib-postgres-staging1` using [playbooks/postgresql_db_migration.yml](playbooks/postgresql_db_migration.yml)
- [ ] run the playbook for the role
- [ ] run a deploy of the application
- [ ] modify the roles nginxplus config to point to the new VM

VM's updated to jammy

### staging

- [x] approvals
- [x] abid
- [x] bibdata
- [x] dss
- [ ] dpul
- [ ] figgy
- [x] geaccirc
- [x] lib-jobs
- [x] lockers_and_study_spaces
- [x] mudd-dbs
- [x] oawaiver-staging
- [x] orangelight
- [ ] pulfalight
- [x] repecwp
- [ ] pulmap

### production

- [x] approvals
- [ ] abid
- [x] bibdata
- [x] dss
- [ ] dpul
[...]

PR #4261, opened 2023-09-14
I3476 catalog staging update

kayiwa:
related to #3476

### DIFF
--- a/inventory/all_projects/orangelight
+++ b/inventory/all_projects/orangelight
@@ -17,9 +17,13 @@ catalog-indexer3.princeton.edu
 catalog-qa1.princeton.edu
 
 [orangelight_staging]
-# catalog-staging1.princeton.edu
+[orangelight_staging:children]
+orangelight_staging_webservers
+orangelight_staging_indexers
+
+[orangelight_staging_webservers]
+catalog-staging1.princeton.edu
 catalog-staging2.princeton.edu
-catalog-indexer-staging1.princeton.edu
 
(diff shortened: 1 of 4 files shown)


inventory/all_projects/orangelight
```diff
@@ -19,3 +12,14 @@ catalog3.princeton.edu
...
 catalog-indexer2.princeton.edu
 catalog-indexer3.princeton.edu
+
+[orangelight_qa]
+catalog-qa1.princeton.edu
+
+[orangelight_staging]
+# catalog-staging1.princeton.edu
+catalog-staging2.princeton.edu
+catalog-indexer-staging1.princeton.edu
```
christinach, line 22: I understand that we dint't have catalog-indexer-staging1 and this is the first staging indexer, but  can we name it to catalog-indexer-staging2 for consistency with the web server name?

inventory/all_projects/orangelight
```diff
@@ -19,3 +12,14 @@ catalog3.princeton.edu
...
+
+[orangelight_staging_indexers]
+catalog-indexer-staging1.princeton.edu
```
christinach, line 25: same here

inventory/all_projects/orangelight
kayiwa: I will be making catalog-staging1 jammy shortly. So we will have two webservers and one indexer. I left catalog-staging1 in the event that things went kablooie

christinach (approved):

PR #4267, opened 2023-09-15
[Bibdata-qa] update redis host to be lib-redis-prod1.princeton.edu

christinach:


acozine:
Closing because we will use the local bibdata redis instead. To upgrade local redis installations:
```
sudo add-apt-repository ppa:redislabs/redis
sudo apt-get update
sudo apt upgrade redis
```

PR #4282, opened 2023-09-18
[Postgresql] Move postgresql common role to playbook.

tpendragon:
It's both a "set a server up" playbook and a "apps include this" playbook - for "set a server up" it should be in the playbook, for "apps include this" common not being a dependency will allow us to optimize setting up required databases.

This also tags the postgresql role includes with "postgresql." By tagging postgresql we can run playbooks and just have it set up access/databases/users. This is useful for migrations. (see step 3 here: https://github.com/pulibrary/pulfalight/issues/1319)

### DIFF
--- a/playbooks/cicognara.yml
+++ b/playbooks/cicognara.yml
@@ -14,7 +14,7 @@
     - role: roles/ruby
     - role: roles/deploy_user
     - role: roles/passenger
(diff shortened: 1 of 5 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

PR #4284, opened 2023-09-18
Upgrade redis to 7.2.1

carolyncole:
Co-authored-by: Anna Headley <hackartisan@users.noreply.github.com>


carolyncole:
This did not work on pdc-describe.  It is possible if the template is updated to systemd [here](https://github.com/pulibrary/princeton_ansible/blob/main/roles/redis/templates/redis.conf.j2#L140) before running that it might work.  I'm going to lose the PR, since I have no system to test this theory upon.

hackartisan:
These changes make the redis upgrade work, you just have to run your app's playbook to get the right redis config, then run the upgrade playbook.

tpendragon (approved):

PR #4291, opened 2023-09-19
[ojs] update postgresql

kayiwa:
- update to jammy postgres
- point loadbalancer to jammy vm


kayiwa:
Impossible to determine if the error is OJS, the loadbalancer or something else entirely

PR #4298, opened 2023-09-19
fixes key typo - should be .pem not .key

acozine:
Small fix to the nginx files.


VickieKarasic:
Thanks for your help with this @acozine!

PR #4301, opened 2023-09-19
removes unused cert and key for appdeploy

acozine:
Follows up on #4293.

The new towerdeploy config uses ACME certs on the load balancers, so we don't need the old appdeploy cert and key.


sandbergja (approved):
Thanks, @acozine !

PR #4302, opened 2023-09-20
[catalog] update production

kayiwa:
- migrate to new postgresql server
- comment out bionic vms

related to #3476


christinach (approved):
Thanks @kayiwa

PR #4324, opened 2023-09-22
Allow nodejs to be downloaded from alternative urls

### LINKED ISSUE #4314: Upgrade Pulmap to Node 18
Do this by compiling nodejs from source so we are not bound VM operating machine upgrades.

eliotjordan:
Closes #4314

### DIFF
--- a/group_vars/pulmap/production.yml
+++ b/group_vars/pulmap/production.yml
@@ -85,6 +85,8 @@ rails_app_vars:
     value: '/home/{{deploy_user}}/cloud_config/google_cloud_credentials.json'
   - name: GOOGLE_CLOUD_PROJECT
     value: 'pulibrary-figgy-storage-1'
+  - name: NODE_OPTIONS
+    value: '--openssl-legacy-provider'
 sneakers_workers: GeoblacklightEventHandler
 sneakers_worker_name: pulmap-sneakers
 datadog_api_key: "{{ vault_datadog_key }}"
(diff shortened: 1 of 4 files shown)


group_vars/pulmap/production.yml
```diff
@@ -11,7 +11,8 @@ passenger_app_env: "production"
 passenger_server_name: "pulmap.princeton.edu"
 passenger_ruby: "/usr/local/bin/ruby"
 bundler_version: "2.3.18"
-desired_nodejs_version: "v16.15.0"
+nodejs_release_url_alt: "https://github.com/pulibrary/nodejs-binaries/releases/download/"
```
acozine, line 14: I'd vote to use the existing variable instead of adding a separate variable and task for the alternative URL.

Do we want to put build Node binaries on our mirror instead of on GitHub?

eliotjordan: I wasn't able to get variable override to happen because in the [order of variable precedence](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence), task variables (like nodejs_release_url) are higher than variables defined in playbook var files. Now that I'm looking at it though, we might be able to move `nodejs_release_url` from `nodejs/vars/main.yml` to `nodejs/defaults/main.yml`

eliotjordan: As for the binaries on the mirror, how would we get them on to the server in an automated fashion? Is there an FTP service?

eliotjordan: Moving the `nodejs_release_url` variable didn't work. It still has a higher precedence.

eliotjordan: Fixed.

acozine (approved):
Looks good, thanks @eliotjordan.

PR #4326, opened 2023-09-25
[allsearch-api] Rate limit requests / sec

maxkadel:
Closes https://github.com/pulibrary/allsearch_api/issues/45

Based on #4072 - using a lower burst rate, because this application does not download a bunch of assets when first loading the page, as the Catalog does.


kayiwa (approved):
OK! Thanks @maxkadel

PR #4334, opened 2023-09-27
[geoserver] move to new dbserver

### LINKED ISSUE #4332: get geoserver database
Migrate geoserver from lib-postgres3 physical hardware
New versions of postgres use `scram-sha-256` and the jdbc for geoserver is not capable of using it. As a stop gap. Set up a database server that uses `md5`

- [x] geoserver-db-prod1

kayiwa:
point geoserver to the new database server

closes #4332


acozine:
Currently the geoserver application is on its own postgresql server because the code required md5 database auth. It's running psql 15 but with custom config.

PR #4335, opened 2023-09-28
[geoserver] remove role and playbook

kayiwa:
prevent the belief that this playbook can be run.


eliotjordan:
Remade commit to delete all GeoServer references

Issue #4327, opened 2023-09-26
Create Playbook for All Search Front-end


kevinreiss:
### Dependencies

- [x] Node v. 18 or higher
- [x] Ruby (for capistrano, use our current default)

### Role
See [Create a new role instructions in Readme](https://github.com/pulibrary/princeton_ansible#create-a-new-role)
- [ ] Molecule tests
- [ ] Put variables required to get tests to pass in 
  - roles/allsearch_frontend/default/main.yml 
  - group_vars/allsearch_frontend/[common | production | staging]
 
### Questions
Do we need a new role, or can we just create a new playbook that uses the nodejs and ruby_s roles?

PR #4337, opened 2023-09-28
[allsearch-frontend] add a very basic playbook

sandbergja:
helps with #4327 

I ran this on staging on 28 September, 2023.


christinach (approved):
Thanks @sandbergja 🍄

PR #4341, opened 2023-09-29
do not restart nginx when creating new certs

acozine:
Adding a cert is necessary but not sufficient for adding or updating a site on the load balancer. To avoid unexpected behavior, we should not restart nginx when we add the cert - we should only restart nginx when we deploy the new/updated config that uses the cert.

### DIFF
--- a/playbooks/incommon_certbot.yml
+++ b/playbooks/incommon_certbot.yml
@@ -7,8 +7,6 @@
   remote_user: pulsys
   become: true
 
-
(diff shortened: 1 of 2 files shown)


maxkadel (approved):
🎉  Thanks @acozine!

carolyncole (approved):

kayiwa (approved):
OK! thanks @acozine

PR #4342, opened 2023-09-29
Remove prds jammy vm from the inventory

christinach:


kayiwa (approved):
OK! thanks @christinach and @sandbergja

PR #4370, opened 2023-10-05
Add connection upgrade headers to CDH sandbox proxy location

rlskoeser:
I have an app running on the sandbox that apparently requires connection upgrade headers when proxied by nginx. I have these settings in the nginx config on the sandbox server and when I access that server via ssh port forwarding the app works; when I disable these, I get an error about a web socket connection. When it is working in sandbox nginx and I access it through the load balancer it does not work. I'm hoping this configuration will allow it to work through lib-adc.


kayiwa (approved):
OK! thanks @rlskoeser

PR #4373, opened 2023-10-06
[PDC-ORCID] Creating a playbook for the pdc-orcid servers

### LINKED ISSUE #4363: Create Ansible playbook for new PDC ORCID Servers
Start with: https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/pdc_describe.yml

carolyncole:
fixes #4363


carolyncole:
There is a playbook, it runs, but the code does not run.

Bundle is getting an error 
```
Bundler::PermissionError There was an error while trying to write to `/usr/local/lib/ruby/gems/3.2.0/cache/bigdecimal-3.1.4.gem`. It is likely that you need to grant write permissions for that path.
```

Something to deal with on Monday...

PR #4374, opened 2023-10-06
Add location to serve as proxy for simulating risk solara app

rlskoeser:


kayiwa (approved):
OK! thanks @rlskoeser

PR #4377, opened 2023-10-06
update service and sourcecategory for pgsql in datadog

acozine:
We had these two datadog config params backwards. This PR resets them to do what we want for our postgres logs:

- `service` can be any string; this is what shows up in Datadog UI as a way to filter the logs view - we set it to `postgres_cluster`, which is more descriptive than `database`
- `sourcecategory` is the type of log - we set this to `database`, which is more accurate than `postgresql_logs`


sandbergja (approved):
Thanks, @acozine !

PR #4382, opened 2023-10-09
notify when nginx playbooks start, with tags

### LINKED ISSUE #4381: Notify when the nginxplus playbook starts
For easier tracking when a sensitive playbook fails, notify at the beginning of the playbook run as well as at the end (on success).

- [x] include any tags that are passed

Include `nginxplus.yml` - what other playbooks should we add this to?

acozine:
Closes #4381.

If no tag is passed at runtime, the output of the notification will say:

Ansible is now running `update loadbalancer configuration` with the `all` tag on lib-adc2.

Are there other playbooks we want to add this functionality to?

### DIFF
--- a/playbooks/nginxplus.yml
+++ b/playbooks/nginxplus.yml
@@ -19,6 +19,13 @@
     when: ansible_limit is not defined
     run_once: true
 
+  - name: tell slack the playbook is starting
+    slack:
+      token: "{{ vault_pul_slack_token }}"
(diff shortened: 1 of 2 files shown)


maxkadel (approved):

kayiwa (approved):
OK! 🥇 Thanks @acozine

kayiwa (approved):

PR #4390, opened 2023-10-10
Cert cleanup

acozine:
Part of #4094.

As far as I can tell, these certs are no longer in use. If anyone knows any obscure reason why we need to keep them, please speak up soon!

- cpanda.princeton.edu no longer exists
- ~~dataspace-staging.princeton.edu no longer exists~~
- dataspace-staging.princeton.edu runs on GCP, not on our load balancers
- lib-analytics.princeton.edu never really existed
- ~~oar-staging.princeton.edu no longer exists~~
- oar-staging.princeton.edu runs on GCP, not on our load balancers
- oarbeta.princeton.edu no longer exists
- pulsearch.princeton.edu is configured in catalog-prod.conf and the cert for catalog.princeton.edu has a SAN name for pulsearch
- republic-staging.princeton.edu is configured in libwww-proxy-pass-staging.conf and does not need a cert


jrgriffiniii (changes requested):
I apologize, the following still do exist:

- https://dataspace-staging.princeton.edu/
- https://oar-staging.princeton.edu/

If there is a need to migrate from these and use other domain names, I can definitely look to support whatever is needed. Please let me know how best to proceed.

acozine:
Thanks @jrgriffiniii! I will take those off the chopping block.

jrgriffiniii (approved):
I am extremely sorry, please know that @kayiwa has clarified that those certs are managed on the Google Cloud Platform

acozine:
Original comment edited to reflect the clarification.

acozine:
Per today's discussion, we will move keys for sites that still exist but don't run on our load balancer into the `keys` directory of this repo. We can add other keys there too.

PR #4394, opened 2023-10-11
[nginx] make playbook work

### LINKED ISSUE #4393: nginxplus playbook
new nginxplus playbook fails with this error

```bash
fatal: [lib-adc2.princeton.edu]: FAILED! => {"msg": "Invalid data passed to 'loop', it requires a list, got this instead: #ansible-alerts. Hint: If you passed a list/dict of just one element, try adding wantlist=True to your lookup invocation or use q/query instead of lookup."}
```

kayiwa:
- the curly braces need to be quoted when they are the first item
- remove unneeded ssl_certificates in location block

closes #4393

### DIFF
--- a/playbooks/nginxplus.yml
+++ b/playbooks/nginxplus.yml
@@ -36,6 +36,8 @@
   remote_user: pulsys
   strategy: linear
   become: true
+  vars_files:
+    - ../group_vars/nginxplus/main.yml
 
   tasks:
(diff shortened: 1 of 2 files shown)


maxkadel (approved):

acozine (approved):
Looks good, thanks @kayiwa!

Issue #4349, opened 2023-10-02
Create nginx config using a template, not files


acozine:
Right now we copy static config files from `/roles/nginxplus/files/http/conf/` to the load balancers. This has some drawbacks:

- deploying config is slow
- when we delete a site, we have to delete the entire directory to get rid of its config file
- separate files make it hard to see where our configuration diverges
- small fixes like the one for #4090 require editing dozens of files

Replacing static files with a template should include:
- [ ] add vars for things that do diverge, including health_check_path

We may need to retain a few static config files for a while at least; figure out how the two types of config can co-exist.

acozine:
#4231 is a start on this work.

carolyncole:
After this work is complete  “evaluate if adding a capistrano-like deployment method would make nginx+ more stable.”

acozine:
Per @maxkadel the allsearch configs would be a good first test. Per @bess we will follow a phased approach, moving one or two configs to the template at a time.

PR #4397, opened 2023-10-11
First steps to nginx config template

acozine:
Partial fix for #4349.

This PR:
- creates a template for nginx config files
- adds four sites to the vars consumed by the template (plus more, commented out)
- adds tasks to create and upload site configs for those four sites using the template, to the /tmp directory so we can test/diff
- removes tasks for Stream config, which we are not using
- uses `|` instead of `:` in task names in the `upload-config.yml` file so we don't have to quote them
- uses FQCN for tasks in the `upload-config.yml` file


bess (approved):
😍

roles/nginxplus/tasks/conf/upload-config.yml
```diff
@@ -44,35 +53,27 @@
...
-
-- name: "Setup: Upload NGINX Stream Configuration Files"
-  copy:
-    src: "{{ item }}"
-    dest: "{{ nginx_stream_upload_dest }}"
+- name: Setup | Upload NGINX HTTP Templated Site Config Files
+  ansible.builtin.template:
+    src: http/library.conf.j2
+    dest: /tmp/{{ item.name }}.conf
+    # dest: /etc/nginx/conf.d/{{ item.name }}.conf
```
maxkadel, line 60: Should we put in a comment explaining that this is where they will eventually go, once testing is complete, or is that self-explanatory?

roles/nginxplus/templates/http/library.conf.j2
```diff
@@ -0,0 +1,97 @@
...
+        app_protect_security_log_enable on;
+      {% endif %}
+        proxy_pass http://{{ item.short_name }};
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_cache {{ item.short_name }}cache;
+        proxy_intercept_errors on;
+        # TODO don't do a health check if there's only one server
+      {% if item.health_check_URI is defined %}
+        health_check interval=10 fails=3 passes=2 uri={{ item.health_check_URI }};
```
maxkadel, line 66: Could `health_check_URI` have a default of `/` (which I'm guessing is the nginx default) and thus avoid a conditional? I think it's fine either way, just trying to think of ways to simplify.

roles/nginxplus/vars/main.yml
```diff
@@ -1,2 +1,152 @@
 ---
-# vars file for roles/nginxplus
+# the 'sites' dictionary is used by the library.conf.j2 template
+sites:
+  # - name: example-prod
```
maxkadel, line 4: Really appreciate this in-file documentation

roles/nginxplus/vars/main.yml
```diff
@@ -1,2 +1,152 @@
...
+    visibility: public
+    app_protect: disabled
+  - name: abid-staging
+    servers:
+      - abid-staging1
+    short_name: abid-staging
+    location: "/"
+    visibility: private
+    app_protect: enabled
+  - name: allsearch-api-prod
```
maxkadel, line 45: This might not have been in when you started this, but I think you could do allsearch and allsearch-staging as well (in addition to the api-only boxes)

maxkadel (approved):
Looks good! Left a couple questions, but I really like where this is going!

roles/nginxplus/templates/http/library.conf.j2
acozine: I'm kinda cheating here. We don't want to do health checks on all sites, because for sites with only one server a health check is all overhead and no gain. This way we can use "health_check_URI" tell which sites should have one at all. But yeah, your solution is a good one longer-term.

roles/nginxplus/vars/main.yml
acozine: Cool! I'll add them in.

acozine:
Try tigerdata staging and prod, orcid staging and prod.

acozine:
For bibdata, we need to have three additional proxy settings:

- `proxy_connect_timeout`
- `proxy_send_timeout`
- `proxy_read_timeout`

All are set to `2h` - is this a reasonable default for all our sites?

acozine:
Known remaining work to a first-round merge:
- [ ] add a directory for templated config
- [ ] include the templated config directory in the main nginx config file, so both static and templated config files are loaded by the load balancer
- [ ] update the task to send the templated files to that new directory
- [ ] remove the `template_test` tag
- [ ] test playbook again

acozine:
Gist for current blocker: https://gist.github.com/acozine/889141cde3f713d8c4b26b83fe93d334

acozine:
@sandbergja and I tested the tasks for identifying and removing obsolete dynamic config files today. I added tasks for doing the same for obsolete static config files. I think after the break we can start moving the easy configs over. Procedure:
- [ ] confirm that the SSL certs for the first sites to migrate have the correct dir names
- [ ] update the main config to include the new dynamic config directory
- [ ] remove the static config files for the first sites to migrate
- [ ] run the playbook and test

If that works as expected, then we can merge this PR and migrate one or two configs per PR for the next few months.

acozine:
Rebased ahead of the holidays.

acozine:
Next (final?) steps for current PR:
- [ ] backup config directories for easy rollback
- [ ] run nginx -t after main config tasks are done
- [ ] if nginx -t fails, restore prior config from the backed up directories

Future work before we move our most complicated sites to the config template:
- [ ] engineer this so we can, from a branch, test our configs without going live - so for a test run we could just put config in non-active directories for comparison and debugging
Workflow for migrating a site from static to templatized config would then be:

1. open a branch, add vars to make the templatized config
2. run playbook from branch with the flag or other solution to output all configs to a non-working directory
3. compare static and templatized configs and do any needed debugging
4. delete static config on branch once things look good
5. run the playbook from branch "for realz" to affect actual working config

acozine:
More progress today. When the `nginx -t` verification fails, instead of doing an automated recovery from the backed-up config directory, we decided it's better to post a message with the failure saying "time to get a human involved, this failed, please restore from <dir>."

acozine:
Closing in favor of #4654

PR #4403, opened 2023-10-12
[ORCID-PRINCETON] updating to more closely match pdc_describe to get cas login to work

carolyncole:


carolyncole:
This issue seems to be fixed by not utilizing the [latest rack and Puma](https://github.com/pulibrary/orcid_princeton/pull/40).

PR #4411, opened 2023-10-16
Added allsearch api url

rladdusaw:
Added a required config for pulibrary/allsearch_frontend#41

### DIFF
--- a/group_vars/allsearch_frontend/production.yml
+++ b/group_vars/allsearch_frontend/production.yml
@@ -1,3 +1,4 @@
 ---
+allsearch_api_url: https://allsearch-api-production.princeton.edu
 passenger_server_name: allsearch.princeton.edu
 allsearch_frontend_environment: production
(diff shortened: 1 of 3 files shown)


group_vars/allsearch_frontend/production.yml
```diff
@@ -1,3 +1,4 @@
 ---
+allsearch_api_url: https://allsearch-api-production.princeton.edu
```
sandbergja, line 2:
```suggestion
allsearch_api_url: https://allsearch-api.princeton.edu
```

The api url doesn't have `-production`

sandbergja (changes requested):
Thanks, @rladdusaw -- just a small correction to the prod url

group_vars/allsearch_frontend/production.yml
rladdusaw: Thanks!

sandbergja (approved):
🐈‍⬛

PR #4414, opened 2023-10-17
[allsearch_frontend] prefix envvar with VITE_

sandbergja:
This makes it available to the client

Helps with https://github.com/pulibrary/allsearch_frontend/issues/88

Sibling PR: https://github.com/pulibrary/allsearch_frontend/pull/89


christinach (approved):
Thanks @sandbergja 🍄 🍄

PR #4435, opened 2023-10-25
add pg_stat_statements to track SQL statistics

### LINKED ISSUE #4369: Fix postgresql Datadog integration
Logs are now transmitted to Datadog from our postgresql cluster. However, the integration still does not work correctly. On lib-postgres-prod1 if you run `sudo datadog-agent status` you see:
```
  Loading Errors
  ==============
    postgres
    --------
      Core Check Loader:
        Check postgres not found in Catalog

      JMX Check Loader:
        check is not a jmx check, or unable to determine if it's so

      Python Check Loader:
        unable to import module 'postgres': No module named postgres
```
and in the Datadog UI, the host entry for lib-postgres-prod1 still complains:
```
Instance #initialization[ERROR]: {"Core Check Loader":"Check postgres not found in Catalog","JMX Check Loader":"check is not a jmx check, or unable to determine if it's so","Python Check Loader":"unable to import module 'postgres': No module named postgres"}
```
Diagnose and fix this.

acozine:
Related to, and possibly closes #4369.

This change-set is necessary but may not be sufficient - it enables the statistics module and adds the related extension to the default database. However, the [docs](https://postgrespro.com/docs/postgresql/15/pgstatstatements) say that `query identifier calculation must be enabled`, and I'm not sure how or where to do that.

The `pg_stat_statements` module requires additional shared memory, so we should check the memory allocation on the postgres boxes before we implement this change.

I think we can use `pg_stat_statements` with or without Datadog, so implementing this in staging seems like a good next step.


acozine:
Relevant docs:
https://postgrespro.com/docs/postgresql/15/runtime-config-statistics#GUC-COMPUTE-QUERY-ID
https://postgrespro.com/docs/postgresql/15/pgstatstatements

acozine:
Hold until the new year, when we are all back and ready to triage as necessary.

Issue #4409, opened 2023-10-16
Review Drupal 9 and Research Data Roles


kevinreiss:
Review playbook. 

related to https://github.com/pulibrary/researchdata/issues/263

sandbergja:
Closing in favor of https://github.com/pulibrary/researchdata/issues/263

PR #4437, opened 2023-10-26
[researchdata] update to jammy

kayiwa:
- update nodejs
- use jammy vm

related to #4409

### DIFF
--- a/roles/nginxplus/files/conf/http/researchdata_staging.conf
+++ b/roles/nginxplus/files/conf/http/researchdata_staging.conf
@@ -4,6 +4,7 @@ proxy_cache_path /data/nginx/researchdata-staging/NGINX_cache/ keys_zone=researc
 upstream researchdata-staging {
     zone researchdata-staging 64k;
     least_conn;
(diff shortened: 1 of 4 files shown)


roles/nginxplus/files/conf/http/researchdata_staging.conf
```diff
@@ -4,6 +4,7 @@ proxy_cache_path /data/nginx/researchdata-staging/NGINX_cache/ keys_zone=researc
 upstream researchdata-staging {
     zone researchdata-staging 64k;
     least_conn;
+    # server prds-staging1.princeton.edu resolve;
     server prds-staging1.princeton.edu resolve;
```
sandbergja, line 8: We want the load balancer to talk to staging2, right?

```suggestion
    server prds-staging2.princeton.edu resolve;
```

sandbergja (changes requested):
Thanks @kayiwa !  Just one question about the nginx config.

sandbergja (approved):
Thanks, @kayiwa !

PR #4438, opened 2023-10-27
[researchdata] ensure that nginx user can access the settings.php symlink

sandbergja:


sandbergja:
Closing in favor of putting this commit into #4437

PR #4448, opened 2023-11-01
Introduce and document consistent scheme for central redis db usage

### LINKED ISSUE #4046: Redis DB numbering scheme
Many of our group vars files have an [app]_REDIS_DB environment variable. There are a couple of central redis instances and guidance on what number to use when spinning up new infrastructure would be helpful. Should we maintain or maybe be able to generate a list of numbers already in use?

Running `redis-cli INFO keyspace` will list info about databases that currently have keys, but it's not clear whether this would be sufficient.

### LINKED ISSUE #4294: Document how to use the central redis server
in this repo, in the redis role readme

- What variables should be used in the group_vars file
- How to find a redis database id that's not already in use (the databases are referred to by number, from 0 to 15).

hackartisan:
Note that orangelight QA and orangelight prod both are and have been pointing at db 7 on the central prod redis host.

closes #4294
closes #4046
advances pulibrary/pulfalight#1323

### DIFF
--- a/roles/redis/README.md
+++ b/roles/redis/README.md
@@ -1,7 +1,52 @@
-Role Name
+Redis Role
 =========
 
-This role install redis
+This role installs a redis server. If your application only needs to connect to
+the central redis instance, you don't need to run this role.
+
+### Connecting to the central redis
+
+To connect to the central redis instance, define a variable `central_redis_db` and
+give it a number between 1 and 15 that hasn't been used yet for the central host
+you're connecting to. Don't use database
+0, which is the default database in case one is not specified. To find an unused
+number, grep the group_vars directory for `central_redis_db` /
+`central_redis_host` pairs (example `ag` command below). Once you've
+identified a db number for your application you can set it to an environment
+variable that your application will access in its redis or sidekiq
+configuration.
+
+You also need to have your application install the redis-tools package.
+
+For good measure, add your app to the list below to help us track which
+applications are using the central redis hosts.
+
+Example configuration in a group_vars file:
+
+```
+rails_app_dependencies:
+  - redis-tools
+central_redis_host: 'lib-redis-prod1.princeton.edu'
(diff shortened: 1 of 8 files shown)


roles/redis/README.md
```diff
@@ -1,7 +1,50 @@
-Role Name
+Redis Role
 =========
 
-This role install redis
+This role install a redis server. If your application only needs to connect to
```
acozine, line 4:
```suggestion
This role installs a local redis server. If your application only needs to connect to
```

acozine (commented):
It's funny that we document using the central Redis server in the role for installing a local Redis server, but after an initial laugh I agree that it's a good place. If someone is looking for information on connecting an application to Redis, they will see both options here. 

I found one typo, but otherwise it looks good. Thanks @hackartisan!

christinach (approved):

christinach (approved):
Thanks @hackartisan !

roles/redis/README.md
hackartisan: Thanks! Fixed.

tpendragon:
I like it, but I do wish there was a way to put all the redis DB numbers next to one another so we can visually verify there aren't any apps using two.

hackartisan:
This has been run for pulfalight. @christinach do you all want to run it on orangelight-qa to get it looking at the new redis db? And if so do you want to do that before it's merged or after?

hackartisan:
> I like it, but I do wish there was a way to put all the redis DB numbers next to one another so we can visually verify there aren't any apps using the same one.

Yeah. We can always change it again later if we have a new idea!

hackartisan:
Okay I went ahead and reverted the orangelight redis db numbers, so that this PR isn't blocked and DACS can follow up on that in their own time. I created https://github.com/pulibrary/princeton_ansible/issues/4455. cc: @christinach

PR #4453, opened 2023-11-02
Skip rails services restart if it's not there or manually

### LINKED ISSUE #4315: Only restart workers on boxes that have workers
Changes introduced in https://github.com/pulibrary/princeton_ansible/pull/4221 make playbook runs fail for machines with no workers. Add in a check or change configuration such that we only try to restart services that are actually on a given box.

hackartisan:
- Only restart rails services that are defined on the box
- Allow skipping the rails services restart

closes #4315
helpful for pulibrary/pulfalight#1323

### DIFF
--- a/roles/rails_app/handlers/main.yml
+++ b/roles/rails_app/handlers/main.yml
@@ -4,4 +4,7 @@
     name: "{{ item }}"
     state: restarted
   loop: "{{ rails_app_site_config_services }}"
-  when: running_on_server
+  # the list of services includes `.service` on the end of each one, and you
+  # can't use template variables in conditionals, so we do this map stuff to
+  # remove the `.service` and then check for each service
+  when: running_on_server and (item in services.keys()|list|map('split', '.')| map('first'))
--- a/roles/rails_app/handlers/main.yml
+++ b/roles/rails_app/handlers/main.yml
@@ -1,3 +1,4 @@
+# To run the rails role without restarting services use --skip-tags restart_services
 ---
 - name: "restart rails app services"
   ansible.builtin.service:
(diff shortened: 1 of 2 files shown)


roles/rails_app/handlers/main.yml
```diff
@@ -1,7 +1,14 @@
...
     state: restarted
   loop: "{{ rails_app_site_config_services }}"
-  when: running_on_server
+  # the list of services includes `.service` on the end of each one, and you
+  # can't use template variables in conditionals, so we do this map stuff to
+  # remove the `.service` and then check for each service
+  when: >
+    running_on_server and
+    (item in services.keys()|list|map('split', '.')| map('first')) and
+    'restart_services' not in ansible_skip_tags
```
acozine, line 14: This is an odd use of tags. I would have to remind myself why I need to skip a `restart_services` tag, when nothing in the role has that tag. Did you try marking the handler with `tags:` and adding both `restart_services` and `site_config`? That would be a more standard way to incorporate the same logic and use of tags when running a playbook or role.

acozine (commented):
Overall this is great, and fixing this problem is hugely helpful. I had one question about the implementation - see below.

roles/rails_app/handlers/main.yml
hackartisan: I didn't try it because my understanding was that you can't skip tags with a handler. But maybe that has changed -- I'll try it now!

hackartisan: Yeah, it doesn't work :( 

Handlers have some kind of special logic so they kick in no matter what you do with tags. This `not in ansible_skip_tags` is a workaround for that.

acozine: Gotcha. I'll keep thinking about alternatives, but meanwhile this solves our immediate problem.

Issue #4443, opened 2023-10-30
Why the nginx playbook only puts the start in infrastructure slack


carolyncole:
I'm not saying we should change this, but I want to document that the reason the start is being put in infrastructure and not the end.

The reason is that the start utilizes a [token that can put a message into infrastructure](https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/nginxplus.yml#L26). The finish utilizes a [token that CAN NOT put a message into infrastructure](https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/nginxplus.yml#L54).

PR #4458, opened 2023-11-03
[ORCID] Moving slack alert to an included file

carolyncole:
The goal would be to eventually have a singe place to update slack alerts if desired

We are using an old slack token that can only alert to `ansible-alerts`: `vault_pul_slack_token`.  I think we should be moving towards utilizing  `vault_tower_slack_token`

This could just be a search and replace, but it seemed nice to move the slack update to a central location.  I only updated ORCID for now, but all playbooks could be updated to utilize the `include_tasks` once this is merged.

refs #4443 

Code bing run: https://pulibrary.slack.com/archives/C9Z7D6D4G/p1699025104773019


playbooks/slack_playbook_complete.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: tell everyone on slack you ran an ansible playbook
+  community.general.slack:
+    token: "{{ vault_tower_slack_token }}"
+    msg: "Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}"
+    channel: "{{ item }}"
+  loop: "{{ slack_alerts_channel is string | ternary([slack_alerts_channel], slack_alerts_channel) }}"
```
carolyncole, line 7: takes care of the fact that the channel could be a string or array

acozine (commented):
+1 for this - it would make future changes to the standard slack message much easier.

The `loop` that handles either a string or an array is unfamiliar to me - I'd love to see it in action.

I find the name of the new file misleading. I'd change it from `playbooks/slack_playbook_complete.yml` to something like `playbooks/slack_tasks.yml`, because with only tasks (no `hosts` entry, etc.) that file would not run as a playbook on its own.

carolyncole:
@acozine I meant it as the slack tasks for the completion of a playbook.  What would make that more clear? `slack_task_for_end_of_playbook` `slack_end_of_run` (something else)?

carolyncole:
@acozine it's the tunery that allows it to either be an array or a single item `ternary([slack_alerts_channel], slack_alerts_channel)`

VickieKarasic (approved):
Looks good - thanks @kayiwa!

PR #4464, opened 2023-11-10
[postgresql] add warm standby

kayiwa:
- add postgresql production cluster
- add postgresql warm standby server


kayiwa:
run successfully on production cluster

jrgriffiniii (approved):

PR #4466, opened 2023-11-10
[nginxplus/tls] add private key

### LINKED ISSUE #4463: add private keys
we need the private keys for pulmirror and tigris added

kayiwa:
pulmirror keys were already added

this closes #4463


VickieKarasic (approved):
Thanks @kayiwa

PR #4473, opened 2023-11-21
allow catalog to use custom 404s

acozine:
Closes https://github.com/pulibrary/orangelight/issues/3732.

Creates an error config page that handles most errors on the load balancer, but allows 404s to be handled by the underlying application.

Changes the catalog config to use this new error config page.

Other applications can now update their configs to use this as desired.


christinach (approved):
🍄 🐴 🐬

PR #4067, opened 2023-07-14
Update solr configs

acozine:
Following our most recent challenges with our Solr infrastructure, this PR updates our garbage collection settings and sets the stage for more granular control of the heap memory.

The new GC settings have already been implemented and tested.

The heap changes suggested in these comments have not been implemented or tested at all. If we do pursue optimizing heap memory management, we will have to update both scripts (solr.in.sh and jammy_solr.in.sh) as well as the group variables files.

PR #4484, opened 2023-11-28
partially reverts PR #4067

### LINKED ISSUE #4483: Revert configuration changes to solr garbage collection
We made some changes in response to an incident which turned out to have had a cause unrelated to GC / heap settings. Configuration changes were checked in to this repo and deployed to staging, but not to prod. We want to bring this repository and the staging machines back in line with the current state of the production machines.

A complicating factor is that the playbook currently fails.

Acceptance criteria:
- [ ] Revert changes made in https://github.com/pulibrary/princeton_ansible/pull/4067/files
- [ ] Ensure both staging and production have the right configuration

acozine:
The changes from #4067 were never deployed to production. We are now seeing issues in Solr, and wondering if we should use the older configuration.

Closes #4483.

### DIFF
--- a/roles/solrcloud/defaults/main.yml
+++ b/roles/solrcloud/defaults/main.yml
@@ -42,8 +42,13 @@ solr_gc_tune: "-XX:NewRatio=3 \
 -XX:+ParallelRefProcEnabled"
 
 # Enable verbose GC logging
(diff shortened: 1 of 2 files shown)


acozine:
Merging to keep production stable. We should open a new PR with GC changes and test again in the new year.

PR #4497, opened 2023-12-04
[byzantine, ealapps, fpul] migrate databases

kayiwa:
- change database name to comply with stricter SQL standard
- point ealapps to new jammy database
- point byzantine applications to the new database


christinach (approved):
Thanks @kayiwa

Issue #4499, opened 2023-12-04
Jammy upgrades needed


acozine:
We have a spreadsheet of the last VMs and boxes still running Bionic. Upgrade/replace these!

https://docs.google.com/spreadsheets/d/1p96OiWVL_fR6rzNE8oeC744o1DJvsGPUV1K-7QeNXlk/edit#gid=0

### PHP Drupal Based Apps
1. get the mac address of the vm you are about to replace
2. create a jammy vm on the same host you are replacing
3. edit the `group_vars/project/main.yml` (sometimes `group_vars/project/common.yml` or `roles/project/vars/main.yml` to have `php_version: "8.1"`
4. run the `playbooks/project.yml`
5. Run a cap to deploy the application
6.

acozine:
We upgraded the Solr and Zookeeper shared boxes with #4944 and #4969.

acozine:
The last item on the original spreadsheet is libimages2 - final work is documented in #5020.

PR #4508, opened 2023-12-06
[byzantine] upgrade vm to jammy

kayiwa:
- jammy uses php8.1
- use the new jammy VM for byzatine-staging

related to #4499


sandbergja (approved):
Thanks, @kayiwa !

PR #4534, opened 2023-12-12
adds second box to Tower config on nginxplus

acozine:
We now have two VMs for our Tower/Controller.

This PR adds the second one to the config for the load balancer.


kayiwa (approved):
OK! Thanks @acozine

PR #4535, opened 2023-12-12
[locator] move staging to jammy vm

VickieKarasic:
upgrade locator-staging1 to Jammy VM


acozine (approved):
Already run in both prod and staging. New version of php is now defined in the common vars.

PR #4539, opened 2023-12-14
[libwww] move to jammy redis

### LINKED ISSUE #4504: point libwww to lib-redis-prod1
It looks like we have moved `libwww-staging` to point to the redis-staging server.

Can we do this for the production

kayiwa:
point to the correct special collection db server

will close #4504


acozine:
@kayiwa has run this from the branch, merging so we don't run from main and undo this!

PR #4562, opened 2024-01-04
[os-updates] clean up space

kayiwa:
some of our servers are filling up disk space because of retained files. This task fixes that


VickieKarasic (approved):
Thanks @kayiwa

PR #4563, opened 2024-01-04
[pulfalight] provision jammy prod and staging machines

hackartisan:
refs pulibrary/pulfalight#1355


hackartisan:
This is all set, just needs to get swapped in on the load balancer.

hackartisan:
These are live.

kayiwa (approved):
OK! Thanks @hackartisan

PR #4568, opened 2024-01-08
[os_updates] i4567 comment out apt clean task in maintenance playbook

VickieKarasic:
comment out task to clean the local repository related to i4567


kayiwa (approved):
OK! Thanks @VickieKarasic

PR #4498, opened 2023-12-04
Configure passenger to load bundler before loading app

hackartisan:
This helps avoid gem version conflicts like the one encountered in
https://github.com/pulibrary/pulfalight/pull/1353

PR #4570, opened 2024-01-09
Remove bundler version management

hackartisan:
Modern versions of bundler comply with gemfile configuration by checking
the "bundled with" line and downloading the desired bundler version
before executing bundler. (if it's already installed it'll just use the
correct installed version). This means we don't need to manage bundler
versions carefully like we did before.

Also, this specific bundler version management causes problems in the
context of the changes introduced in
https://github.com/pulibrary/princeton_ansible/pull/4498 -- when
passenger preloads bundler, ruby looks for a version of bundler that
appears to be locked in at installation time when installed from source.
I couldn't find a way to configure ruby to use a specific bundler
version during installation from source. So we have to ensure that the
version that ruby will look for is installed.


acozine:
PR currently fails the `capistrano` test suite with:
```
  TASK [ruby_s : ruby_s | output existing bundler version] ***********************
  fatal: [instance]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'installed_bundler' is undefined\n\nThe error appears to be in '/home/runner/work/princeton_ansible/princeton_ansible/roles/ruby_s/tasks/main.yml': line 32, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: ruby_s | output existing bundler version\n  ^ here\n"}
```

hackartisan:
I realized I can achieve this same end by simply not defining `bundler_version` for my apps. So I'm fine to close it. If at some point all the teams stop defining bundler_version for all their apps, maybe we'll want to delete the code.

PR #4580, opened 2024-01-11
[prancible] upgrade the version of ansible

### LINKED ISSUE #4567: Maintenance playbook should run apt clean
We've had a couple of machines run out of space because of too many old apt package files.

PR #4562 has the correct fix, but the task currently fails. The `clean` option was added in Ansible Core 2.13, and we are still running and older version of Ansible with Ansible Core 2.12. So this issue is related to #3782.

kayiwa:
this updates our version of ansible to 9.1.0 

closes #4567


VickieKarasic (approved):
This looks good to me - thanks @kayiwa

acozine (changes requested):
I'm eager to get Ansible upgraded, but I would like to see at least some Molecule tests run and succeed before merging this. It looks like our CI suite just skipped everything - they all seem to say "successful" but in under 60 seconds each, which doesn't sound like any tests actually ran. I'm going to review this as "Request changes" but the PR itself looks fine - the review is only intended to hold up merging until we can at least run the tests locally.

roles/tomcat9/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+# defaults file for roles/tomcat8
+tomcat8_enabled: true
+tomcat8_hostname: localhost
+tomcat8_server_port: 8005
+tomcat8_catalina_port: 8983
+tomcat8_catalina_redirect_port: 8443
```
acozine, line 7: These vars are all named `tomcat8_` something - should we update them to be `tomcat9_`? I'm not clear if we're changing the existing role to install tomcat 9 instead of tomcat 8, or if we need to maintain both for a while, or what. But I find this confusing now, and I bet Future Me will find it even more confusing in future.

roles/tomcat9/handlers/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+# handlers file for roles/tomcat8
```
acozine, line 2: Here too, and in the service name below . . .

roles/tomcat9/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+galaxy_info:
+  role_name: tomcat8
```
acozine, line 3: And here.

roles/tomcat9/meta/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  company: Princeton University Library
+  description: Tomcat8 Role
```
acozine, line 5: And here.

roles/tomcat9/molecule/default/converge.yml
```diff
@@ -0,0 +1,15 @@
...
+  become: true
+  pre_tasks:
+    - name: update cache
+      ansible.builtin.apt:
+        update_cache: true
+        cache_valid_time: 600
+  tasks:
+    - name: "Include tomcat8"
+      ansible.builtin.include_role:
+        name: tomcat8
```
acozine, line 15: This one would be particularly challenging if we continue to have a role called `tomcat8`, since then the tomcat9 tests would actually be running the tomcat8 tasks . . .

roles/tomcat9/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# vars file for roles/tomcat8
```
acozine, line 2: 9

acozine (commented):
This PR is doing a lot of great stuff! The Tomcat changes are a little confusing, and there's still a CI failure, but we're getting there.

roles/tomcat9/defaults/main.yml
kayiwa: we are replacing tomcat8. I was uncomfortable removing it because of the geoserver but will remove it with the next commit.

acozine (approved):
CI success, huzzah!

PR #4583, opened 2024-01-11
[nfs/bibdata] add qa nfs 

kayiwa:
- add lib-fs-qa server
- point bibdata to qa nfsserver

connect bibdata to nfs-qa fileshare


sandbergja (approved):
Thanks, @kayiwa !

PR #4586, opened 2024-01-12
[bibdata] passenger: preload bundler to avoid error with psych gem dependency

sandbergja:


sandbergja:
This was fixed in #4814, closing

PR #4587, opened 2024-01-12
[solr8cloud] install zookeeper and solr8cloud

kayiwa:
- add new jammy zookeeper and solr hardware
- use zookeeper role
- point to solr8 group vars
- comment out solr9 configuration
- simplify the checks


hectorcorrea (approved):
Thank you!

PR #4593, opened 2024-01-17
[solr8cloud] add a qa environment

kayiwa:
- add none-d hosts
- add a qa environment


kayiwa:
we decided to make this the new staging cluster.

PR #4594, opened 2024-01-17
[orangelight] add new qa machine

kayiwa:
add a new indexer-1 machine and bring back the web app vm
deploy application to use new db host


christinach (approved):
Thanks @kayiwa

PR #4601, opened 2024-01-22
[common] this adds plocate to jammy boxes

### LINKED ISSUE #4449: Could we install plocate onto all the jammy boxes?
I love using `locate`, but not all the boxes have it.  [plocate](https://plocate.sesse.net/) is the preferred implementation at the moment, and it's available in the jammy apt repo.

kayiwa:
adds plocate to all our jammy boxes
adds fdfind which is available on the ubuntu repos for jammy

closes #4449


sandbergja (approved):
Awesome!  Thanks, @kayiwa !

PR #4612, opened 2024-01-24
Configure nginx to allow tower to call /websocket/

### LINKED ISSUE #4043: Tower reports 404s on /websocket
When we deploy code from Ansible Tower, we see a lot of requests to the `/websocket` path that result in 404s. We see these in the nginxplus dashboard (in the entry for Ansible Tower) and in the browser (when looking at the network tab in devtools). 

These errors may reflect attempts to update the playbook in progress. If we can fix these, therefore, we might see real-time updates to our Tower jobs.

Investigation today with @ishasinha1 @lizgarciao @sandbergja @christinach suggests that the problem is in the nginx config. 

Resources:
https://github.com/ansible/awx/issues/1861
https://github.com/ansible/awx/blob/devel/docs/websockets.md
https://stackoverflow.com/questions/76441962/authentication-to-awx-websocket
https://docs.ansible.com/ansible-tower/3.1.0/html/administration/troubleshooting.html#websockets-port-for-live-events-not-working (though we do see requests reaching the server, so this is less likely to be a firewall issue)

Nginx-specific resources:
https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/
https://datatracker.ietf.org/doc/html/rfc7239
https://stackoverflow.com/questions/72557636/difference-between-x-forwarded-for-and-x-real-ip-headers

Possibly related:
https://github.com/ansible/awx/issues/1850

acozine:
Closes #4043.

Attempt to fix issue 4043 with nginx config. Very much WIP.

### DIFF
--- a/roles/nginxplus/files/conf/http/ansible-tower_prod.conf
+++ b/roles/nginxplus/files/conf/http/ansible-tower_prod.conf
@@ -38,9 +38,15 @@ server {
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
+        #attempt to enable websocket
+        #proxy_set_header Host $host; not sure if we can use this
+        proxy_http_version 1.1;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "Upgrade";
         proxy_cache ansible-towercache;
         proxy_intercept_errors on;
         health_check interval=10 fails=3 passes=2;
     }
+
     include /etc/nginx/conf.d/templates/errors.conf;
 }
--- a/roles/nginxplus/files/conf/http/ansible-tower_prod.conf
+++ b/roles/nginxplus/files/conf/http/ansible-tower_prod.conf
@@ -38,14 +38,18 @@ server {
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
-        #attempt to enable websocket
-        #proxy_set_header Host $host; not sure if we can use this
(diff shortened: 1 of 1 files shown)


acozine:
We ran this today, and while it didn't break anything, we're still seeing the errors in the logs about `django.request Not Found: /websocket/`. That said, it looked like live job updates might be happening - so maybe the log messages are not tightly coupled to the problem . . . I'm going to hold this PR until we have some EEs and can gather more data.

acozine:
Job updates are still not happening. If you kick off a job, you see the `Waiting for updates` message as long as you stay on the Jobs page, long after the job has failed (or succeeded). To see the actual output, you have to go back to `Jobs` and then select the individual job.

acozine:
In the Nginx access logs, I see:

```
73.xxx.xxx.xxx - - [26/Jan/2024:14:36:58 +0000] "GET /websocket/ HTTP/1.1" 404 2502 "-" 128.112.IP-of-ansible-tower2:443, 404, 0.091, 0.091"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36" "-"
```

acozine:
@VickieKarasic and @sandbergja and I fixed this today - in addition to this nginx config update, we had to add a config file to Tower. The file is called `/etc/tower/conf.d/websocket_config.py` and the content is:

```
# support headers for real-time updates via websocket
BROADCAST_WEBSOCKET_PROTOCOL = 'https'
BROADCAST_WEBSOCKET_PORT = 443
BROADCAST_WEBSOCKET_VERIFY_CERT = True
```

With those two changes, the WebSockets connections works and live updates come through. If you inspect the web browser, there are still some errors listed, but they don't seem to affect performance, so we are ignoring them for now.

kayiwa (approved):
NICE! thanks @acozine

Issue #4491, opened 2023-11-30
Solr8cloud playbook fails on conditional check


acozine:
The `solr8cloud.yml` playbook currently fails on the first task of the `openjdk` role. The error message is:
```
fatal: [lib-solr-staging6.princeton.edu]: FAILED! => {"msg": "The conditional check 'install_ruby_from_source is not defined or install_zookeeper_from_source|bool == false' failed. The error was: error while evaluating conditional (install_ruby_from_source is not defined or install_zookeeper_from_source|bool == false): 'install_zookeeper_from_source' is undefined\n\nThe error appears to be in '/Users/alicia/projects/princeton_ansible/roles/openjdk/tasks/main.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# tasks file for roles/openjdk\n- name: openjdk | install java (package)\n  ^ here\n"}
```

The source of the failure is not in the `openjdk` role, it's [here](https://github.com/pulibrary/princeton_ansible/blob/a3de12f00fe670d42a25dbd48bb96f964b0cb170/roles/solrcloud/meta/main.yml#L19-L20) in the `meta/main.yml` file for the `solrcloud` role. 

The conditional check was added as part of supporting Solr9 in #3819. I'm guessing it works for the Solr9 use case but breaks for Solr8.

PR #4616, opened 2024-01-25
[pulmap] point to new solr8d cluster

kayiwa:
the solr8 staging cluster is in an unreliable state

we are moving collections to the new solr8d cluster

related to #4491


VickieKarasic:
Superseded by #4771

PR #4625, opened 2024-01-29
adds bastion info for GCP hosts

acozine:
Partial fix for #4018.

This provides connection details for our Google cloud resources. However, we still need to provide connection details for our AWS cloud resources.


kayiwa (approved):
OK! Thanks @acozine

PR #4628, opened 2024-01-30
Adds GCP groups to staging and prod supergroups

acozine:
Partial fix for #4018.

Not sure this is the best approach; we can discuss here.


acozine:
We need to use `doas` for the `become_method` on all openBSD machines. All bastion hosts on GCP and on AWS are running openBSD.

kayiwa (approved):
OK! @acozine

Issue #4622, opened 2024-01-29
ensure www-data is part of the conan group


kayiwa:
we have the cdh webserver having problems reading media files.

Make sure the www-data are part of the conan group

kayiwa:
Also make sure the group user can write to `/var/nfs/cdh`

kayiwa:
futureproof solution is get idmapping to work

PR #4637, opened 2024-02-06
[nfsserver] add fqdn

kayiwa:
while idmap is enabled on the NFS servers we need to add the FQDN or it will export this are princeton.edu

related to #4622

### DIFF
--- a/roles/nfsserver/tasks/main.yml
+++ b/roles/nfsserver/tasks/main.yml
@@ -38,6 +38,14 @@
   loop:
     - "{{ cdh_fileshare_mount }}"
 
+- name: nfsserver | add fqdn for idmapping
+  ansible.builtin.lineinfile:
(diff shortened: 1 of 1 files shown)


roles/nfsserver/tasks/main.yml
```diff
@@ -38,6 +38,14 @@
   loop:
     - "{{ cdh_fileshare_mount }}"
 
+- name: nfsserver | add fqdn for idmapping
+  ansible.builtin.lineinfile:
+    path: /etc/idmapd.conf
+    insertbefore: '# Domain = localdomain'
+    line: "Domain = {{ inventory_hostname }}"
```
rlskoeser, line 45: IDK what this means exactly, but IDK if that matters. is it worth a comment?

rlskoeser (approved):
OK

PR #4645, opened 2024-02-08
[lae] Add rate limiting to staging and prod

hackartisan:
following the example in #4072


kayiwa (approved):
OK! thanks @hackartisan

hackartisan:
Note that to speed the site back up we also had to restart after applying the rate limit. @maxkadel had seen this before because the passenger queue was so clogged up. A restart clears it out to be able to efficiently accept new requests.

PR #4646, opened 2024-02-08
[prancible] Adding a mise config for those who use the mise software

VickieKarasic:


kayiwa (approved):
OK! Thanks @VickieKarasic

PR #4647, opened 2024-02-08
Add tags for datadog to LAE.

tpendragon:


group_vars/lae/production.yml
```diff
@@ -80,6 +80,7 @@ sneakers_worker_name: lae-sneakers
 datadog_api_key: "{{ vault_datadog_key }}"
 
 datadog_config:
+  tags: "application:lae, environment:production, type:webserverj"
```
hackartisan, line 83: is this right? `webserverj`?

hackartisan (changes requested):

PR #4654, opened 2024-02-09
Offer a template for nginx config files

### LINKED ISSUE #4349: Create nginx config using a template, not files

acozine:
Closes #4349.

Supersedes #4397.

- Adds a template for our "bog standard" nginx config, making it easier to propagate changes across multiple sites
- Makes a backup of the entire config directory so we can roll back quickly when changes fail
- Puts nginx config files into two subdirectories of `/etc/nginx/conf.d`
   - `/static` for static config files (stored as .conf in the nginx role) - we will likely always have static config files for complicated sites
   - `/dynamic` for dynamic config files (built from a template and variables in the nginx role)
- Checks for obsolete files in both directories and removes them - this makes it easy to migrate sites from static to dynamic config and avoids the step of deleting the entire config directory
- Loads the two new subdirectories in the main nginx config
- Adds a custom failure message when `nginx -t` fails
- Moves at least one site from static to dynamic config


acozine:
I don't think we've sanitized the SSL certificate directory names - some of them don't match the config file names. Before we merge this PR, we may want to:
- [ ] add a check that the SSL certificate directory matches the variable name used for the config (this should make `nginx -t` fail if it's wrong, but . . . belt and suspenders)
- [ ] update the cert-generation playbook to use the vars from the nginx role
- [ ] pre-check existing certs and regenerate any that don't match

acozine:
Current failure is:
```
   ERROR! Unable to retrieve file contents
  Could not find or access '/home/runner/work/princeton_ansible/princeton_ansible/roles/byzantine/molecule/default/conf/cleanup-config.yml' on the Ansible Controller.
```

carolyncole:
During Ansible open house we tried to run this PR in it's current state.  It did not create the static directory and skipped the steps for some reason.  @acozine will add in the command we ran after she returns tomorrow.

VickieKarasic:
Discussed in Ansible Open Hours: This seems like it's ripe for testing on the dev loadbalancer now that we have it.

PR #4656, opened 2024-02-09
Includes git branch in slack notification for nginx playbook

acozine:
When somebody runs the nginx playbook, it notifies slack. The message already included the play name, the tags (showing `all` if the playbook was run with no tags), and the server it ran against.

This change discovers the git branch and adds it to the slack message. This is useful in the moment when someone is running updates to the nginx config. It's also useful after the fact if we need to figure out which branch was used the last time our config was updated.

### DIFF
--- a/playbooks/nginxplus.yml
+++ b/playbooks/nginxplus.yml
@@ -15,19 +15,27 @@
     - ../group_vars/nginxplus/main.yml
 
   pre_tasks:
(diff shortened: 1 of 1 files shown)


acozine:
Screenshot of the output from when I ran the branch:
Screenshot 2024-02-09 at 2 10 49 PM

hackartisan (approved):
yessss thank you

PR #4659, opened 2024-02-12
[common] update dust

kayiwa:
the container uses a different architecture platform than our VMs
install these applications on the vm only


sandbergja (approved):
Thanks, @kayiwa

PR #4669, opened 2024-02-15
Replace pulmap blacklight_app with rails_app role

inieinie:
- Also allows running playbooks on Bionic servers.


acozine:
Looks okay, thanks @eliotjordan.

PR #4671, opened 2024-02-16
Upgrade node

maxkadel:
Since all environments are now on Jammy, move nodejs version to common.yml

Run and deployed against staging, looks good.

Connected to https://github.com/pulibrary/bibdata/pull/2314


christinach (approved):
Thanks @maxkadel 🦄

Issue #4663, opened 2024-02-14
Bundler miss match with passenger


carolyncole:
When you install a new ruby and a new passenger gem gets installed.  That gem likes a certain version of bundler to be available.

I have outlined a manual process for reconciling the bundler version here https://github.com/pulibrary/pul-it-handbook/pull/248

What should really be done with passenger.  @hackartisan mentioned that the latest bundler will just download and install the correct version.  Maybe we should stop installing different versions of bundler and just use what gets installed with ruby?

hackartisan:
Here's documentation of how that works in bundler: https://bundler.io/blog/2022/01/23/bundler-v2-3.html

hackartisan:
The passenger error you get looks like `can't find gem bundler (= 2.5.6) with executable bundle (Gem::GemNotFoundException)`

hackartisan:
Can we avoid a world where we have to have a million PRs that look like https://github.com/pulibrary/princeton_ansible/pull/4661/files ?

carolyncole:
I just noticed... `- name: ruby_s | remove installed bundler if default` should happen before `- name: ruby_s | uninstall wrong version of bundler` for the uninstall to work correctly

Maybe passenger would be happy if there was not two versions of bundler installed??

carolyncole:
We just decided in cross team that we should start removing the version from projects as we run the playbooks.  Leave the code for now, with the knowledge that is doesn't work without manual intervention
https://docs.google.com/document/d/1AllskD7tY3mGIc7qVqqs46zSVDGHXckFSl_G1BFJiUg/edit#heading=h.3ejetkruiwxz

PR #4698, opened 2024-02-22
Do not define a version of bundler

acozine:
Partial fix for #4663.

Removes all definitions of `bundler_version` from group_vars and from roles.

With this gone, we can probably remove the `install_bundler` tasks file from the ruby_s role. 

Do we need a task in the main ruby_s tasks file for installing bundler, or can we rely on deployments to do that for us?


VickieKarasic:
From Ansible Open Hours: We think this is merge-able as-is but would want to make sure that we don't take out the step that installs any version.

VickieKarasic (approved):
We tested this on dss-staging and the system stayed up. Thank you @acozine and @kayiwa!

PR #4702, opened 2024-02-22
moves collection list to EE file, builds basic EE for Tower

### LINKED ISSUE #4688: Build execution environment
Build execution environment with Ansible core 2.15.7 and collections versions copied from https://github.com/ansible-community/ansible-build-data/blob/main/8/ansible-8.7.0.yaml

acozine:
Closes #4688.

Up to now we have installed the collections we need to run our playbooks on Tower at runtime. Using custom EEs, with the collections pre-loaded on them, should make Tower runs faster.

This PR:

- deletes the `collections/requirements.yml` file
- adds a directory called `tower_ees`
- adds an EE definition file with the collections list

I don't think this will affect anything in local environments, but I'll make this PR a draft just in case.


acozine:
To build or rebuild an EE:
- change into the `tower_ees` directory: `cd tower_ees`
- edit the EE definition file as needed: `vim execution-environment.yml`
  - alternatively, add a new YAML file in the directory to define a new EE: `vim NewExecEnv.yml`
- run `ansible-builder`: `ansible-builder build -v3 -t <TagOrNameOfEE> --squash all`
  - by default this builds based on `execution-environment.yml`
  - if you added a new file to define a new EE, point `ansible-builder` at your new file: `ansible-builder build -v3 -f NewExecEnv.yml -t <TagOrNameOfEE> --squash all`
- note the hash of the built image in the output of the `ansible-builder` command
- authenticate to `quay.io`
  - For easy authN, create a ~/.config/containers/auth.json file with the top level items ‘auths’ and an entry for each container registry you want to use. Each entry contains a key/value pair: the key is “auth” and the value is the output of “echo -n ‘username:password’ | openssl base64”. See [this PR](https://github.com/containers/image/pull/821/files) and [this superuser post](https://superuser.com/questions/120796/how-to-encode-base64-via-command-line) for more details.
  - authenticate from the command line: `podman login quay.io`
- push the new image to quay.io: `podman push <hash-of-image-ID> quay.io/pulibrary/<name-of-image>`

carolyncole (approved):
We tested this new version of the Execution Environment on Tower for projects, inventory, and rails app.
We also tested running a playbook locally and that did not cause any issues
The tests were done with @acozine @regineheberlein and @sandbergja

PR #4712, opened 2024-02-28
move veridian from load balancer

kayiwa:


acozine:
I'd like to understand why we are making this change and what it means. There doesn't seem to be a ticket describing this work. Where did content from The Prince move to? Not a blocker, necessarily, but PRs with either a description or a related issue (or both) are easier and more comfortable to merge.

acozine:
This PR removes The Prince from our config because it does not run on the load balancers today. There is a VM in our production vSphere with the name `theprince-staging` with DNS of `theprince.princeton.edu`. It is not behind the load balancer. The name redirects to https://papersofprinceton.princeton.edu/, which actually is on our load balancers. We have minimal other information.

acozine:
See #5995 for follow-up.

PR #4713, opened 2024-02-28
Update scsb path for submitCollection

Beck-Davis:


christinach (approved):
Thank you @Beck-Davis 🐻‍❄️  , @sandbergja 🍄 , @christinach 🐬

PR #4715, opened 2024-02-28
Add a DACS group to the inventory

sandbergja:
This way, we can run playbooks and limit to just DACS applications.


christinach (approved):
Thanks 🍄 🐬

PR #4716, opened 2024-02-29
install and/or remove crowdstrike

kayiwa:
two playbooks that install crowdstrike on a VM
currently need to determine how to make it work with OpenSSL 1.1.1 on a Ubuntu jammy server


playbooks/utils/security_theater.yml
```diff
@@ -0,0 +1,95 @@
...
+    ansible.builtin.debug:
+      var: ansible_facts.services
+
+  - name: Print falcon-sensor.service name
+    ansible.builtin.debug:
+      var: ansible_facts.services['falcon-sensor.service'].name
+
+  - name: Print falcon-sensor status name
+    ansible.builtin.debug:
+      var: ansible_facts.services['falcon-sensor.service'].status
```
acozine, line 29: We can remove these tasks, or at least comment them out. We only needed them to see what the variable we needed looked like.

playbooks/utils/security_theater.yml
```diff
@@ -0,0 +1,95 @@
...
+    ansible.builtin.debug:
+      var: ansible_facts.services
+
+  - name: Print ir_agents.service name
+    ansible.builtin.debug:
+      var: ansible_facts.services['ir_agent.service'].name
+
+  - name: Print ir_agent status name
+    ansible.builtin.debug:
+      var: ansible_facts.services['ir_agent.service'].status
```
acozine, line 70: We can delete all of this too - we already checked if the service is running, and we don't need the debugging information now that our condition works.

acozine (commented):
Huzzah for getting our mandated security tools installed!

We can streamline the playbook now that it works well.

Since this playbook's hosts line works a little differently than most, I might also add a comment at the top - something like: 
"# By default this playbook runs on all hosts in the three environment groups. To run against a single host or group, use '--limit <group_or_host_name>'. For example '--limit qa' or '--limit figgy-web-staging1.princeton.edu'."

PR #4717, opened 2024-03-01
[ORCID] Upgrade ruby to 3.2.3

bess:
Ref https://github.com/pulibrary/orcid_princeton/issues/124

### DIFF
--- a/group_vars/orcid/main.yml
+++ b/group_vars/orcid/main.yml
@@ -29,6 +29,8 @@ project_db_host: '{{postgres_host}}'
 # Use Ruby 3.2.3 and install from source
 install_ruby_from_source: true
 ruby_version_override: "ruby-3.2.3"
+bundler_version: "2.4.20"
+passenger_ruby: "/usr/local/bin/ruby"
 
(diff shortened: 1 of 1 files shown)


group_vars/orcid/main.yml
```diff
@@ -26,9 +26,11 @@ application_host_protocol: 'https'
 application_port: '443'
 project_db_host: '{{postgres_host}}'
 
-# Use Ruby 3.2.0 and install from source
+# Use Ruby 3.2.3 and install from source
 install_ruby_from_source: true
-ruby_version_override: "ruby-3.2.0"
+ruby_version_override: "ruby-3.2.3"
+bundler_version: "2.5.6"
```
carolyncole, line 32: @bess I do not think we want to be in the business of saying which bundler.  This is what broke the machine. 
```suggestion
```

carolyncole (changes requested):
Nice except I think the new way is to just let bundler do what version it wants...

carolyncole (approved):

PR #4711, opened 2024-02-28
Adding a playbook to enable redis

carolyncole:
fixes #4707

See enabled on the third line down... Prior to running the playbook it stated disabled
```
pulsys@tigerdata-staging2:~$ sudo systemctl status redis-server
● redis-server.service - Advanced key-value store
     Loaded: loaded (/lib/systemd/system/redis-server.service; enabled; vendor preset: enabled)
     Active: active (running) since Wed 2024-02-28 13:33:16 UTC; 1min 1s ago
       Docs: http://redis.io/documentation,
             man:redis-server(1)
   Main PID: 456492 (redis-server)
     Status: "Ready to accept connections"
      Tasks: 6 (limit: 9347)
     Memory: 3.1M
        CPU: 218ms
     CGroup: /system.slice/redis-server.service
             └─456492 "/usr/bin/redis-server 0.0.0.0:6379" "" "" "" "" "" "" "" "" ""

Feb 28 13:33:16 tigerdata-staging2 systemd[1]: Starting Advanced key-value store...
Feb 28 13:33:16 tigerdata-staging2 systemd[1]: Started Advanced key-value store.
```

PR #4720, opened 2024-03-04
we want redis to restart on reboot

acozine:
Follow-up to #4629.

Enable the redis service so it will start up when a machine is rebooted.


hackartisan (approved):
ty!

acozine:
Note: The playbook added by #4711 sets `enabled` to `true`, but if you then ran the full role from any commit before this PR got merged, it would set `enabled` back to `false`. With this change, `redis-server` will always start on reboot.

PR #4727, opened 2024-03-04
[Common] Fix Vector.

tpendragon:
Datadog has acquired Vector, so all the packages moved to them.

Pulled from https://vector.dev/highlights/2023-11-07-new-linux-repos/

### DIFF
--- a/roles/common/tasks/vector.yml
+++ b/roles/common/tasks/vector.yml
@@ -1,14 +1,20 @@
 ---
-- name: common | add vector repository configuration
-  ansible.builtin.get_url:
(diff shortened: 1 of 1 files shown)


acozine (commented):
Could we put this in the Datadog role instead of in the Common role? We only need these keys on machines where we're installing Datadog, right?

tpendragon:
> Could we put this in the Datadog role instead of in the Common role? We only need these keys on machines where we're installing Datadog, right?

I dunno. I'm not sure what we were using Vector for before datadog acquired it.

acozine:
Okay, makes sense. We can address this question when we pull the `common` role out into a maintenance playbook instead of the build playbooks.

PR #4735, opened 2024-03-06
[Figgy] Upgrade to Jammy

tpendragon:
For now this leaves Figgy connected to figgy1.princeton.edu just for Redis until tonight, when the workers finish processing their jobs and I can migrate redis data to figgy-web-prod1.

### DIFF
--- a/group_vars/figgy/staging.yml
+++ b/group_vars/figgy/staging.yml
@@ -13,7 +13,7 @@ postgresql_is_local: false
 postgres_admin_user: '{{vault_postgres_staging_admin_user}}'
 nodejs_release_url: "https://github.com/pulibrary/nodejs-binaries/releases/download/"
 desired_nodejs_version: "v18.18.0"
-passenger_server_name: "figgy-staging2.princeton.edu"
+passenger_server_name: "figgy-staging.princeton.edu"
 passenger_app_root: "/opt/figgy/current/public"
 passenger_app_env: "staging"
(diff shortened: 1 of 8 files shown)


group_vars/figgy/staging.yml
```diff
@@ -31,6 +31,7 @@ figgy_rabbit_password: '{{vault_rabbit_staging_password}}'
 figgy_rabbit_server: 'amqp://{{figgy_rabbit_user}}:{{figgy_rabbit_password}}@{{figgy_staging_rabbit_host}}:5672'
 figgy_geoserver_user: '{{ vault_figgy_geoserver_user }}'
 figgy_geoserver_password: '{{ vault_figgy_geoserver_password }}'
+figgy_staging_rabbit_host: 'figgy-web-staging1.princeton.edu'
```
acozine, line 34: I did a search for where this var is getting used, and found [this reference](https://github.com/pulibrary/princeton_ansible/blob/c368eed1c2c486b02e07189f024c7ed0b4a21292/group_vars/figgy/staging.yml#L31) in the figgy staging group vars. Does port 5672 need to be open on figgy-web-staging1?

acozine (commented):
It's not a blocker to merging this PR, but I'm wondering if we need to open the RabbitMQ port in staging (or in production) . . .

group_vars/figgy/staging.yml
tpendragon: Yeah probably.

PR #4736, opened 2024-03-06
new role for postfix on pony express

kayiwa:


inventory/all_projects/sendmail
```diff
@@ -0,0 +1,3 @@
+
+[libpony_staging]
```
acozine, line 2: What is the difference between the `libpony_staging` group and the `postfix_staging` group? Do we need both for the same server? Or did you mean to put the old server here, maybe?

playbooks/sendmail.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: build the ponyexpress
+  hosts: libpony_{{ runtime_env | default('staging') }}
+  remote_user: pulsys
+  become: true
+  vars_files:
+    - ../group_vars/sendmail/vars.yml
+    # - ../group_vars/sendmail/{{ runtime_env | default('staging') }}.yml
+    - ../group_vars/sendmail/vault.yml
+  roles:
+    - role: roles/sendmail
```
acozine, line 13: The PR doesn't add a `sendmail` role. Is this outdated?

roles/postfix/README.md
```diff
@@ -0,0 +1,35 @@
+Role Name
+=========
+
+Creates a relay to proofpoint and creates a relay for our applications
```
acozine, line 4:
```suggestion
Builds an email server with a relay to proofpoint and creates a relay for our applications to send email.
```

roles/postfix/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  when: running_on_server
+
+- name: Postfix | Generate Postfix hash map for authentication
+  ansible.builtin.command: postmap /etc/postfix/sasl_passwd
+  changed_when: false
+  when: running_on_server
+  notify: restart postfix
+
+- name: Postfix | Configure Postfix relay restrictions
+  ansible.builtin.lineinfile:
```
acozine, line 49: Using `lineinfile` means we never clean out old IP addresses. This could come back to bite us if we re-use an old IP for something that should not have postfix functionality.

roles/postfix/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+Role Variables
+--------------
+
+Most use for this role will involve adding your IP address to the allow list with the following variable:
+
+```bash
+relay_new_host: 1.2.3.4
```
acozine, line 13: This var name might get confusing. At the time that we add a new host, it would make sense. A year later, it might not feel "new". Not sure how to handle this . . . could we have a central list of "IPs that should have access to our email relay"?

roles/postfix/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+---
+# tasks file for roles/postfix
+- name: Postfix | Install Postfix
+  ansible.builtin.apt:
+    name: postfix
+    state: present
+    update_cache: true
+
+- name: Postfix | Configure Postfix as relay host
+  ansible.builtin.lineinfile:
```
acozine, line 10: With `lineinfile` I think the task would just add the line, so if the file already had `smtp_use_tls = no`, we might end up with both lines in the config file. Something to test.

acozine (commented):
Awesome work!

I had a couple of questions.

roles/postfix/README.md
kayiwa: heh. var names are ALWAYS hard for me! 😬

playbooks/sendmail.yml
kayiwa: removed those. PR started out as sendmail and OIT said we didn't need to use sendmail.

roles/postfix/tasks/main.yml
kayiwa: opted to use templates instead. Thanks

.mise.toml
```diff
@@ -1,3 +1,29 @@
...
 [tools]
-python = {version='3.11', virtualenv='.venv'}
-ruby = '3.2'
+python = ["3.11"]
+ruby = ["3.2.0"]
+node = ["20"]
+
+[settings]
+python_venv_auto_create = true
+experimental = true
```
acozine, line 25: This change affects everything in the repo, right? I have no idea what this file does, but setting a global variable to `experimental` gives me the willies a bit.

Could we add a sentence at the top of this file that explains what it's for?

Also, what does `experimental = true` do for us?

group_vars/sendmail/vars.yml
```diff
@@ -0,0 +1,5 @@
+---
+sendmail_is_local: true
+sendmail_user: "{{ vault_sendmail_user }}"
+sendmail_password: "{{ vault_sendmail_password }}"
+proofpoint_host: "{{ vault_proofpoint_host }}"
```
acozine, line 5: Do we still need these? I thought we were not using sendmail?

roles/postfix/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+Role Variables
+--------------
+
+Most use for this role will involve adding your IP address to the allow list with the following variable:
+
+```bash
+relay_add_host: 1.2.3.4
+relay_remove_host: 1.2.3.4
```
acozine, line 14: are we still using lineinfile for this? I'd prefer to see a template used, so you just update the variable with the hosts in it - any host you add to the var gets added to the config; any host you remove from the var gets removed from the config, like we do with SSH keys.

roles/postfix/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+    owner: root
+    group: root
+    mode: "0644"
+
+- name: Postfix | Add Proofpoint authentication credentials
+  ansible.builtin.copy:
+    dest: /etc/postfix/sasl_passwd
+    content: |
+      [{{ postfix_smtp_relay_host }}]:{{ postfix_smtp_relay_port }} {{ postfix_smtp_relay_username }}:{{ postfix_smtp_relay_password }}
+    mode: '0644'
```
acozine, line 41: I'm guessing there's a reason, but I don't know what it is, so . . . why are we setting this file to `0644` in one task and then immediately resetting it to `0600` in the next task?

roles/postfix/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+  when: running_on_server
+
+- name: Postfix | Generate Postfix hash map for authentication
+  ansible.builtin.command: postmap /etc/postfix/sasl_passwd
+  changed_when: false
+  when: running_on_server
+  notify: restart postfix
+
+- name: Postfix | Add Postfix hosts
+  ansible.builtin.lineinfile:
```
acozine, line 60: This still relies on the right IPs being in the right variable at the right time, and it doesn't allow for tracking those IPs over time under source control. I would much rather see a template here, with a variable that shows changes to the relay hosts over time. That way, for example, when we add a new VM to the inventory, we can add its IP to the relay hosts list in the same PR. And same goes for removing VMs.

acozine (commented):
This is a great step forward, thanks @kayiwa. I see a few places where it could be even better.

PR #4741, opened 2024-03-08
[Figgy] FQDN for Figgy load balancer.

tpendragon:


kayiwa (approved):
OK! Thanks @tpendragon

Issue #4740, opened 2024-03-08
[nginxplus] add x-frame config for library website


VickieKarasic:
- [x] add x-frame option for https://princeton.libwizard.com/ on library-staging config file (see [library-staging.conf](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/library-staging.conf#L43) for where to put header option)
- [x] add x-frame option for https://princeton.libwizard.com/ on library prod config file
- [x] add x-frame option for https://princeton.libwizard.com/ on catalog-staging config file (see [catalog-staging.conf](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/catalog-staging.conf)]
- [x] add x-frame option for https://princeton.libwizard.com/ on catalog-prod config file (see [catalog-prod.conf](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/catalog-prod.conf)]
- [x] add x-frame option for https://princeton.libwizard.com/ on dss-staging config file (see [dss-staging.conf](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/dss-staging.conf)]
- [x] add x-frame option for https://princeton.libwizard.com/ on dss-prod config file (see [dss-prod.conf](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/dss-prod.conf)]
- [ ] put in ticket with Ex Libris to see if we can add an exception for LibWizard domain to Articles+ platform

Some [guidance](https://stackoverflow.com/questions/47405597/x-frame-options-in-nginx-to-allow-all-domains) on how to do this

More [examples](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1183)

VickieKarasic:
Tried changing the `ALLOW-FROM` on nginx http config but this did not work so far (see #4743)

VickieKarasic:
The `ALLOW-FROM` protocol is outdated, so adding `add_header Content-Security-Policy "frame-ancestors 'self' https://princeton.libwizard.com;"` this worked instead.

VickieKarasic:
There wasn't further interest in exploring the Articles+ integration, so will close this issue.

PR #4743, opened 2024-03-08
[nginxplus] Add X-Frame config to library-staging and library-prod

VickieKarasic:
See #[i4740 ](https://github.com/pulibrary/princeton_ansible/issues/4740)


kayiwa (approved):
OK! Thanks @VickieKarasic

PR #4744, opened 2024-03-08
Build a custom EE for faster code deploys

acozine:
The EE now builds and a test template runs on it. However, `cap deploy` still fails, and the problem is with SSH keys. Not sure of the details.

We also need to move the installation of the Capistrano dependencies out of the playbook and into the EE build - this will speed things up too.


acozine:
See https://ansible-tower.princeton.edu/#/jobs/playbook/579/output for more details. If you click on the error message of the failing task, you can see in the `Output` tab that the connection is not using SSH keys. It's prompting for a password instead: `deploy@abid-staging2.princeton.edu's password:`.
Note that the pre-existing deploy template still exists and still succeeds - we are only using this custom EE and the playbook changes in a new template.

PR #4762, opened 2024-03-14
[lib-jobs] add lib-jobs-prod1

kayiwa:
The lib-jobs-prod1 VM was still bionic. This change updates it to jammy and puts them in production


christinach (approved):
Thanks @kayiwa and @VickieKarasic

PR #4754, opened 2024-03-13
[nginxplus] i4624 checkmk role

VickieKarasic:

PR #4782, opened 2024-03-22
Do not list collections in the repo

acozine:
Ansible Tower uses the `collections/requirements.yml` file to load collections into EEs at runtime. This is slow and may be interfering with using custom EEs.

This file was removed in #4702 and was mistakenly reintroduced in #4754.


VickieKarasic (approved):
Thanks @acozine

acozine:
This change definitely fixed the problem we had where the `Prancible` project would not update. See [this job](https://ansible-tower.princeton.edu/#/jobs/project/680/output). All collections are built into the Tower EEs now, so we cannot have a `collections/requirements.yml` file, especially one that calls for different versions of collections than we are installing on Tower.

Issue #4745, opened 2024-03-11
Shut down annotations servers


kelynch:
Please shut down https://annotations.princeton.edu/ and https://annotations-staging.princeton.edu/ according to standard Ops procedures. Let me know if I can add any additional information to this ticket.

VickieKarasic:
Hi @kelynch, I'll be working on this one. We will decommission the servers according to [standard procedures](https://github.com/pulibrary/pul-it-handbook/blob/main/services/decommissioning.md). Usually, we wait 30 days after turning off a server to start the deletion process. Would you like 30 days or would shorter (say, 14 days/2 weeks) suffice in this case? Happy to proceed however you think is best. Thanks!

VickieKarasic:
We will wait the 30 days per Kate's request and will begin decommissioning the week of April 8, 2024.

VickieKarasic:
Decommissioning complete and closed by #4839

PR #4784, opened 2024-03-22
comments out annotations boxes, adds python interpreter for GCP bastion

acozine:
Related to #4745.

The annotations VMs have been powered off but not fully decommissioned. However, they are causing failures in Ansible. This PR comments them out; when we decommission them, we can remove the entire file for the project, along with any references to the group.

This PR also adds a python interpreter definition for the GCP bastion. This should take care of this warning: `Platform openbsd on host bastion-staging.pulcloud.io is using the discovered Python interpreter at /usr/local/bin/python3.9, but future installation of another Python interpreter could change the meaning of that path.`

### DIFF
--- a/inventory/all_projects/annotations
+++ b/inventory/all_projects/annotations
@@ -1,4 +1,4 @@
 [annotations_staging]
-annotations-staging1.princeton.edu
+# annotations-staging1.princeton.edu
 [annotations_production]
-annotations-prod1.princeton.edu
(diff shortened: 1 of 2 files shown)


sandbergja (approved):
Thanks, @acozine !

VickieKarasic (approved):
This looks good - thanks @acozine!

PR #4785, opened 2024-03-22
Run the pulsys user playboook without become/sudo

acozine:
The pulsys user playbook is failing on some hosts because sudo is not available. I think we can run this playbook without using `become`.


acozine:
I ran the pulsys user playbook against one of the failing boxes from this branch, with `privilege escalation` unchecked on the template, and it succeeded: https://ansible-tower.princeton.edu/#/jobs/playbook/696/output. When we merge this PR, be sure to uncheck `privilege escalation` on the template permanently.

sandbergja (approved):
Thanks, @acozine !

Issue #4787, opened 2024-03-25
Offboard Favour Nwogbo


acozine:
Favour has been accepted to an international internship program and will no longer be working in the library.

- [x] Remove her keys from our systems.
- [ ] What else?

acozine:
Related to https://github.com/pulibrary/pul-it-handbook/pull/261

acozine:
Also needed to remove her from the [Students team.](https://github.com/orgs/pulibrary/teams/students/members)

acozine:
Key-update playbook has been run across our infrastructure. I think this is done.

PR #4793, opened 2024-03-26
remove keys for Favour

acozine:
Partial fix for #4787.

Removes SSH keys for Favour. We also need to:
- [x] remove her name from the DevOps groups.


kayiwa (approved):
OK! Thanks @acozine

Issue #4790, opened 2024-03-25
add oawaiver back to full capacity


kayiwa:
with the increased capacity we can now add oawaiver-staging1 and oawaiver-prod1 
Get old [MAC addresses](https://networkregistration.princeton.edu)

- [x] oawaiver-staging1
- [x] oawaiver-prod1
- [x] Uncomment names on inventory oawaiver-staging1
- [x] run playbook (for staging)
- [x] Uncomment names on inventory oawaiver-prod1
- [x] run playbook (for production)

acozine:
The current oawaiver-prod1 is a Bionic box. Decommission that one, and build a new one on Jammy.

VickieKarasic:
Closed by #4802 and #4803. See #3897 for decommissioning the bionic oawaiver-prod1 VM.

PR #4805, opened 2024-03-28
[nginxplus] add oawaiver VMs back into nginxplus config

VickieKarasic:
related to #4790


acozine:
All current machines are running Jammy - we have deleted the old Bionic oawaiver boxes.

PR #4810, opened 2024-03-28
Redirects the CheckMK sites

### LINKED ISSUE #4809: [CheckMK] Traffic to checkMK site should go to the full site URL
Right now, if you open a browser to https://pulcheck-staging.princeton.edu (from on the VPN), you get the standard Apache default page. This is because CheckMK defines a "site" after the main URL. The correct URL for our staging site is https://pulcheck-staging.princeton.edu/pulmonitor/.

Modify the nginx configuration file for pulcheck-staging to send all traffic to the correct URL.

leefaisonr:
closes #4809

### DIFF
--- a/roles/nginxplus/files/conf/http/checkmk_prod.conf
+++ b/roles/nginxplus/files/conf/http/checkmk_prod.conf
@@ -41,7 +41,6 @@ server {
     ssl_prefer_server_ciphers  on;
 
     # Redirect top level traffic to /pulmonitor/
-    # until the top level has content.
     location / {
         return 302 https://$server_name/pulmonitor/;
     }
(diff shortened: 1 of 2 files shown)


acozine:
Per @kayiwa we should copy the pattern from the [datacommons config](https://github.com/pulibrary/princeton_ansible/blob/9613013ddf9d2e0c62ba216286370e0af7b236e4/roles/nginxplus/files/conf/http/datacommons.conf).

acozine:
Tested today - the redirection works, but it seems to break our authentication settings - I get a popup window asking for a Username and Password instead of the SSO auth.

acozine:
Updated the config and deployed to both LBs today. Worked on my machine; awaiting confirmation from others.

acozine:
I did not deploy the production changes today, but they *should* work the same as staging.

acozine:
Tested today, and both staging and production redirect correctly with this config.

jrgriffiniii (approved):

carolyncole (approved):

PR #4811, opened 2024-03-28
[friends] correct state in ansible.builtin.service call

sandbergja:
The documentation says we should use restarted, rather than restart, as the state: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/service_module.html#parameter-state

This caused an intermittent failure in the friends_of_pul playbook (it only failed in cases where the handler was notified by a change made in the id_mapping lineinfile).


christinach (approved):
Thanks @sandbergja !

PR #4818, opened 2024-04-01
remove mailcatcher version from PAS group vars

acozine:
Mailcatcher runs fine without a version. If we upgrade the gem without re-running the whole role, the mailcatcher service fails when we've specified a version.

There is a role default that sets the mailcatcher location to `/usr/local/bin/mailcatcher`, which works just fine. So, don't specify anything in group vars.


kayiwa (approved):
OK! Thanks @acozine

PR #4819, opened 2024-04-02
[postgresql] Configure more frequent log rotation

maxkadel:
Logrotate takes precedence over postgres config for log rotation

Not tested yet


hackartisan (approved):

acozine:
Testing today in staging - this affects the figgy-db-staging boxes as well as lib-postgres-staging.

acozine:
Today we deployed this change to the four staging postgresql servers. On the shared postgresql servers, we deleted some logs from 2023 (`rm /var/lib/postgresql/15/main/pg_log/postgresql-2023-*`) - these logs used the old naming convention and were not getting rotated at all. We noticed that lib-postgres-staging1 is getting a lot more traffic than lib-postgres-staging2. Log rotation does not seem to be happening correctly yet, we will continue to investigate.

acozine:
I was able to run logrotate manually with `sudo service logrotate start` and the files in `/var/lib/postgresql/15/main/pg_log` rotated. This took longer on lib-postgres-staging1, but it did eventually finish. I'll look tomorrow or the next day to see if the cron job make this happen correctly.

acozine:
I checked today, and it looks like this config is doing the right thing - at least all the servers have new logs datestamped today - for example:
```
figgy-db-staging2:
$ ls -la /var/lib/postgresql/15/main/pg_log
total 648
drwx------  2 postgres postgres  20480 May 23 21:26 .
drwx------ 20 postgres postgres   4096 May 23 00:00 ..
-rw-r--r--  1 postgres postgres   5328 May 23 21:10 pg.json
-rw-r--r--  1 postgres postgres 619903 May 23 00:00 pg.json.1
-rw-r--r--  1 postgres postgres   2436 May 21 16:19 pg.log
```
We won't know for sure until tomorrow - if it rotates again tomorrow, then I think we can call this good.

PR #4821, opened 2024-04-03
updates zookeeper inventory, round 1

acozine:
I think the zookeeper inventory hasn't kept pace with the Solr cloud changes.

When we [updated the Solr8 cloud](https://github.com/pulibrary/princeton_ansible/pull/4587/files#diff-a27de0ecaa34c807588fafc5c43509e72fb3261580dbcaba32f1633da341b07c), we changed the staging solr config to look at new zookeeper boxes, but we didn't update the zookeeper inventory.

This might need more cleanup. I cannot SSH to some of the zookeeper prod boxes in the inventory - we may have obsolete boxes listed there too.


kayiwa (approved):
OK! Thanks @acozine

PR #4652, opened 2024-02-08
Create playbook for checkmk agent

leefaisonr:
To run this, you need to install the CheckMk collection: https://galaxy.ansible.com/ui/repo/published/checkmk/general/

PR #4826, opened 2024-04-04
[CHECKMK] Move vars to a file so they can eventually be overridden

carolyncole:
was suggested in https://github.com/pulibrary/princeton_ansible/pull/4652#issuecomment-1965498582


acozine:
Looks good. We probably need to load `group_vars/all/checkmk.yml` as a vars file for the [CheckMK server-build playbook](https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/checkmk.yml) too.

PR #4831, opened 2024-04-05
[bibdata] Move dependencies into playbook to reduce repetition

maxkadel:
- Move role dependencies for testing to converge.yml
- List all roles that are actually needed, comment out roles that are currently duplicates

~As a nice side effect, test runs appear to be [much faster on average](https://github.com/pulibrary/princeton_ansible/actions/workflows/molecule_tests.yml?query=branch%3Abibdata_move_roles_to_playbook+is%3Asuccess)~

To verify, switch to fresh staging VMs, run the playbook, deploy bibdata to the fresh staging VMs, and trigger an incremental update on the Alma sandbox.

Performed the above on April 9-10. 2024, and successfully built the boxes and ran jobs on Sidekiq.

### DIFF
--- a/playbooks/bibdata.yml
+++ b/playbooks/bibdata.yml
@@ -12,12 +12,31 @@
     - ../group_vars/bibdata/vault.yml
     - ../group_vars/bibdata/{{ runtime_env | default('staging') }}.yml
   roles:
+    # re-add when removed from: ruby_s, deploy_user, rails_app, and samba
+    # - role: common
     - role: deploy_user
-    - { role: roles/sidekiq_worker, when: "'worker' in inventory_hostname" }
-    - role: roles/bibdata
-    - role: roles/bibdata_sqs_poller
-    - role: 'hr_share'
-    - role: roles/datadog
+    - role: redis
+    # re-add when removed from passenger
+    # - role: ruby_s
+    # re-add when removed from rails_app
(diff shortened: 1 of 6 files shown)


playbooks/bibdata.yml
```diff
@@ -12,12 +12,31 @@
...
-    - role: roles/bibdata
-    - role: roles/bibdata_sqs_poller
-    - role: 'hr_share'
-    - role: roles/datadog
+    - role: redis
+    # re-add when removed from passenger
+    # - role: ruby_s
+    # re-add when removed from rails_app
+    # - role: bind9
+    # re-add when removed from rails_app
```
sandbergja, line 23: This is nitpicky, but I found it hard to know at a glance which role each comment applied to.  Would it be possible to add them to the comment?
```suggestion
    # re-add when passenger role is removed from rails_app
```

sandbergja (commented):
Thanks, @maxkadel !  I think it's great, just one suggestion for comment readability

playbooks/bibdata.yml
maxkadel: Thanks @sandbergja! Good idea, I've tried to do that

sandbergja (approved):
This is awesome, thanks so much @maxkadel 🦄

PR #4837, opened 2024-04-09
[panos] add to denylist

kayiwa:
- **new roles to add ips to the panos denylist**
- **provide documentation to add new IPs**
- **playbook to add new IPs**


acozine:
We didn't write a ticket for this work . . . is this an accurate definition?

Problem: 
- to avert DDOS attacks and other challenges, we want to be able to add IPs and IP ranges to the library firewall's deny list quickly and reliably

Requirements:
- solution must work even during a major outage - for example, if the load balancers go down completely
- solutions should be accessible to people who do not have direct access to the firewall

Suggested solution:
- create a file that lists IPs and IP ranges to deny
- set the firewall to consume this file and update our deny list on a schedule
- host this file somewhere that is not on our load balancers so we can update it and the firewall can access it even during a major outage

Based on those requirements, could we achieve this goal by creating a deny-list file in the `local_files` directory of this repo and pushing it to a public S3 bucket? This would be lower-maintenance than an openBSD cloud VM with a role that installs and configures nginx on it. 

See the [docs for hosting static websites on S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteHosting.html).

acozine:
Could we design the task to concatenate multiple files? Then we could have one file per incident in a directory in this repo. That way it would be much easier to keep the IPs and ranges in order, and it would be easier to un-ban specific sources if we ever wanted to.

acozine:
Closing in favor of #4872.

PR #4839, opened 2024-04-09
[annotations] i4745 remove annotations vms

VickieKarasic:


acozine:
The annotations project has been retired entirely. This PR removes all references to it on our load balancers, in our inventory, and in our variables. 

Looks good, thanks @VickieKarasic!

PR #4843, opened 2024-04-11
[samba] Remove dependency on common

maxkadel:
- Reduces running the "common" role by 1 in approvals, bibdata, and lib_jobs playbooks

### DIFF
--- a/roles/approvals/meta/main.yml
+++ b/roles/approvals/meta/main.yml
@@ -7,14 +7,14 @@ galaxy_info:
 
   license: MIT
 
-  min_ansible_version: 2.2
+  min_ansible_version: '2.2'
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - 'bionic'
+        - 'jammy'
 dependencies:
   - role: 'rails_app'
   - role: 'mailcatcher'
   - role: 'postgresql'
-  - role: 'hr_share'
--- a/roles/approvals/meta/main.yml
+++ b/roles/approvals/meta/main.yml
@@ -12,7 +12,6 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - 'bionic'
         - 'jammy'
 dependencies:
   - role: 'rails_app'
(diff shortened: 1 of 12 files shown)


roles/approvals/meta/main.yml
```diff
@@ -7,14 +7,14 @@ galaxy_info:
...
 
-  min_ansible_version: 2.2
+  min_ansible_version: '2.2'
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - 'bionic'
+        - 'jammy'
```
christinach, line 16: why do we keep version jammy?

roles/hr_share/meta/main.yml
```diff
@@ -2,16 +2,16 @@
...
+  min_ansible_version: '2.2'
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
-dependencies:
-  - role: 'samba'
+        - 'bionic'
+        - 'jammy'
```
christinach, line 16: why do we keep version jammy?

roles/samba/meta/main.yml
```diff
@@ -7,12 +7,12 @@ galaxy_info:
...
   platforms:
     - name: Ubuntu
       versions:
-        - 16.04
-        - 18.04
-dependencies:
-  - role: 'common'
+        - 'xenial'
+        - 'bionic'
+        - 'jammy'
```
christinach, line 17: why do we keep version: jammy?

roles/lib_jobs/meta/main.yml
```diff
@@ -7,17 +7,17 @@ galaxy_info:
...
 
-  min_ansible_version: 2.2
+  min_ansible_version: '2.2'
 
   platforms:
     - name: Ubuntu
       versions:
-        - 18.04
+        - 'bionic'
+        - 'jammy'
```
christinach, line 16: why do we keep version: jammy?

christinach (commented):
Thanks @maxkadel . One question that I added as a comment.

roles/approvals/meta/main.yml
maxkadel: Updated to only include Jammy, and not Bionic, since Jammy is the version of Ubuntu we know these roles run on successfully.

christinach (approved):
Thanks @maxkadel 🦄

PR #4848, opened 2024-04-12
Annotation inventory

kayiwa:
these refer to servers no longer available


VickieKarasic (approved):
Thanks for catching the annotations references @kayiwa! I'll let Alicia weigh in on the security_theater additions.

Issue #3897, opened 2023-05-12
Audit inventory


acozine:
Now that we have consolidated automation in this repo, audit the inventory. 
- [x] run a report in vSphere to confirm that all existing VMs appear on the list (eventually, can we automate this part and avoid having to maintain a static inventory of our vSphere VMs?)
- [ ] confirm the status of the legacy boxes and add comments to the inventory list for Future Us (for example, what does libserv53 do?)
- [ ] test the new inventory with a simple playbook that gathers facts and returns a list of VMs by OS flavor & version
We should be able to answer questions like:
- [x] is everything listed in inventory active/being used?
- [x] is everything we want to automate listed in inventory?
- [ ] do the groups and groups-of-groups accurately reflect our use cases?

acozine:
For instructions on exporting a list of all VMs, see https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vcenterhost.doc/GUID-C0E8DD52-677E-464F-A3EA-044EE20B7B92.html and/or https://masteringvmware.com/export-reports-from-vsphere-web-client/.

acozine:
VMs to decommission:
- [ ] oawaiver-prod1
- [ ] ojs-prod2
- [ ] prds-prod1
- [ ] prds-staging1
- [ ] pulfa3-staging1
- [ ] pulfalight-prod1
- [ ] pulfalight-prod2
- [ ] pulfalight-worker1
- [ ] repec-prod1

acozine:
We added the QA NFS servers to the QA group in #4583 but the Staging and Prod groups for NFS are not in the larger production and staging groups (so they are not getting touched by Patch Tuesday). Is that on purpose @kayiwa?

acozine:
Decommissioned VMs to remove from inventory:
- [ ] lib-monitor1
- [ ] libserv31
We could also look at the list of failed hosts in Ansible Tower to see if any of those are obsolete.

VickieKarasic:
Hosts still failing in Ansible Tower on weekly staging os_updates (as of today): 

- [ ] ansible-exec-node1.princeton.edu
- [ ] ansible-exec-node2.princeton.edu
- [ ] ansible-tower1.princeton.edu
- [ ] ansible-tower2.princeton.edu
- [ ] bastion-prod.pulcloud.io
- [ ] community.docnow.io  
- [ ] eal-apps-prod1.princeton.edu
- [ ] gcp-dataspace-prod1 
[...]

PR #4849, opened 2024-04-12
Inventory and OS-discovery improvements

acozine:
Partial fix for #3897.

Adds an `obsd` group that sets `doas` for the privilege escalation method for all GCP OpenBSD boxes.
Removes the `annotations` groups from the `staging` and `production` supergroups.
Sets `hosts: all` for the Check OS playbook, so we don't miss strays (not all boxes are included in the three `env` groups - `staging` and `qa` and `production`).
Shows the distro as well as the version in the playbook output.


kayiwa (approved):
OK! Thanks @acozine

PR #4850, opened 2024-04-12
[svn] add new user

kayiwa:
Add the password in the vault


VickieKarasic (approved):
Looks good @kayiwa - thanks!

PR #4854, opened 2024-04-15
[figgy] pin kernels

kayiwa:
this will revert and pin your kernel to a different one


tpendragon (approved):

tpendragon (approved):

acozine:
Because we were nervous about pinning the kernel, we closed this PR instead of merging it. However, we had pinned the kernel on figgy-worker-prod6. To un-pin the kernel, you need to do the following steps:
- run `uname -a` to see the current kernel version
- delete `/etc/apt/preferences.d/pin-kernel` (or other added filename; by default the directory only contains two files)
- run `sudo apt-get install --install-recommends linux-generic` to install the latest kernel
- reboot so the newly installed kernel is active
- run `uname -a` again to confirm

PR #4855, opened 2024-04-15
Add a few vim configurations to all the servers

hackartisan:
This file can also act as a first .vimrc for anyone new to vim to use on
their local machines


acozine:
I like this idea - I'm always scrambling to turn on line numbers, it would be great to have them on by default.

Could we move `templates/vimrc.local` to `files/vimrc.local` and use the file module instead of the template module? There are no variables in that file, so we don't need to template it.

acozine:
We looked at this in the Open House - general agreement that it's a good idea and the settings are at least a good start, and maybe a good ending point. Any reason this is still a draft @hackartisan ?

hackartisan:
Cool, thanks @acozine! It's still a draft because I didn't try it out yet. I can at least make the change from templates to files, that's a good call. If someone wants to try out this branch while doing other work that'd be cool, otherwise I will look to try it out before too long.

hackartisan:
@acozine I tried this out locally with molecule and it worked great. I was able to ssh into the container, open a file in vim, and see some code highlighting, line numbers, do case-insensitive search and see the search results highlighted. screenshot attached

Screenshot 2024-05-17 at 12 31 29 PM

christinach (approved):
Thanks @hackartisan

PR #4857, opened 2024-04-16
[prancible] delete temp files

### LINKED ISSUE #4552: provision all machines to periodically delete old files from /tmp
Proposal to configure all machines to auto-delete anything in /tmp older than a certain time frame. 6 months seems very conservative?

kayiwa:
- **delete unused temp files**
- **pass a variable for those who need deletion to happen more frequently than once a week**

refs #4552


hackartisan (approved):
This seems useful. I am not sure it closes that issue, which was more about setting a (probably bigger than a week) default cleanup, I guess I assumed as a role we'd include in every VM build?

How about I merge this, but I'll change the "closes" to a "refs" and we can discuss on that issue whether the original ticket is in line with y'all's vision in operations?

PR #4859, opened 2024-04-16
Adding playbook to add rule for server

carolyncole:
how do I get this to run not on the server, but locally run with `ansible-playbook playbooks/utils/checkmk_add_rule.yml --ask-vault-pass --limit=pdc-describe-staging2.princeton.edu`


carolyncole:
The configuration should move to group_vars/checkmk/rails_rule.yml and then in the playbook we should read in an environment variable like `runtime_env`.  Default is empty string to run generic checkmk rules.

carolyncole:
Configuration was moved to a file per group rails, ruby, drupal.  The only rule group we have so far is for rails.

kayiwa (approved):
OK! Thanks @carolyncole

bess (approved):

PR #4861, opened 2024-04-17
[redis] upgrade redis

kayiwa:
when we upgrade redis the installation adds a new configuration
this ensures that the new configuration will listen on all interfaces


acozine:
Updates to the latest available packaged Redis - we are not constraining the version of redis-server.

PR #4868, opened 2024-04-22
[nginxplus] setup the dev environment

### LINKED ISSUE #4674: refactor nginx plus role and include WAF
We will need to refactor nginxplus role to do the following

- [ ] install nginxplus on a clean VM
- [ ] install WAF on a cleanVM
- [ ] Deploy applications on dev environment

kayiwa:
enable us to build a nginxplus role from scratch
Final steps are documented by this https://github.com/pulibrary/pul-it-handbook/pull/279
closes #4674


playbooks/nginxplus_production_rebuild.yml
```diff
@@ -1,12 +1,15 @@
 ---
 - name: rebuild nginx server
-  hosts: nginxplus_production # default to staging when we get a staging environment going
+  hosts: nginxplus_staging # default to staging when we get a staging environment going
```
tpendragon, line 3: Should this run off a variable? Like `hosts_{{ environment }}` or something

roles/nginxplus/files/conf/http/001-default.conf
tpendragon, line 1: I think all these still need to be around.

playbooks/nginxplus_staging.yml
```diff
@@ -0,0 +1,66 @@
+---
+# you MUST run this playbook on a single host with '--limit' for example `ansible-playbook -v --limit lib-adc2.princeton.edu playbooks/nginxplus.yml`
+# to update the second load-balancer, switch which adc# host is currently active, then run the playbook on the second (formerly active, now inactive) host
+# to update configuration for existing sites, run with '-t update_conf' for example `ansible-playbook -v --limit lib-adc2.princeton.edu playbooks/nginxplus.yml -t update_conf`
+# if you want to remove obsolete config files, you must pass '-e nginx_cleanup_config=true'
+# note that this will delete the entire config directory
+# To find the active load-balancer ssh onto a machine and run `ip a |grep 146` the active one will display `inet 128.112.203.146/32 scope global eno1`
+
+- name: update loadbalancer configuration
```
tpendragon, line 9: I'm a little confused about why there are two playbook files if there's a staging and prod environment, and why they both point to the same hosts.

tpendragon (changes requested):

acozine (changes requested):
Great work getting the nginx role to build a pair of staging load balancers! The PR needs a bit more work before we can integrate these changes and manage production and staging from the same role.

playbooks/nginxplus_staging.yml
kayiwa: Until this is ready to go in, this is a safety precaution. While this is tested fairly well, there are still moving parts on our dev environment that I am not quite ready to apply. I can be convinced otherwise but it feels like a useful guard as we transition.

acozine: Is your concern that we might run the `staging` playbook against production LBs and get unexpected changes in production?

roles/nginxplus/templates/default_nginx.conf.j2
```diff
@@ -0,0 +1,69 @@
...
+{% endif %}
+{% if nginx_modules.image_filter %}
+load_module modules/ngx_http_image_filter_module.so;
+{% endif %}
+{% if nginx_modules.rtmp and nginx_type == "plus" %}
+load_module modules/ngx_rtmp_module.so;
+{% endif %}
+{% if nginx_modules.xslt %}
+load_module modules/ngx_http_xslt_filter_module.so;
+{% endif %}
```
acozine, line 30: I would remove the `if/endif` sections for modules we do not use and do not plan to use, just to clean up the template and make it easier to read (and also to reduce the number of variables we need).

roles/nginxplus/templates/default_nginx.conf.j2
```diff
@@ -0,0 +1,69 @@
+{{ ansible_managed | comment }}
```
acozine, line 1: I might rename this template (which would also require changing the name in the task that uses it). I don't think it's really a default config any more.

roles/nginxplus/tasks/modules/install-modules.yml
```diff
@@ -14,5 +14,8 @@
 - import_tasks: install-rtmp.yml
```
acozine, line 14: This might be for a future PR, but I would love to see us remove the `import_tasks` statements, conditions, etc., that we do not use and do not plan to use.

roles/nginxplus/tasks/keys/apt-key.yml
```diff
@@ -6,15 +6,25 @@
     update_cache: true
 
 - name: "Install: NGINX Signing Key URL"
-  set_fact:
+  ansible.builtin.set_fact:
     default_keysite: >-
-      {{ (ansible_distribution_release == "trusty")
+      {{ (ansible_distribution_release == "jammy")
```
acozine, line 11: Do we want to set the release to `jammy` everywhere? Isn't production running Focal?

acozine (commented):
Found a few other questions about this change-set.

roles/nginxplus/templates/default_nginx.conf.j2
kayiwa: 🤔 I am pretty sure it is used. (certainly on the dev environment)

tpendragon (changes requested):
Just some merge conflicts. This looks great otherwise!

tpendragon (approved):
This looks great to me, and I'd like to get it in before we start fussing with this branch much more. @acozine ?

PR #4869, opened 2024-04-22
[drupal9] unbreak CI on main

kayiwa:
- **we are using a different docker container**
- **we are using a different image**


kevinreiss (commented):
I don't know if it matters but we don't use Drupal 9 any longer. We use Drupal 10.

kayiwa:
made the change to rename role to drupal10

PR #4870, opened 2024-04-25
Updates read-only firewall discovery playbook

acozine:
Supersedes #3820.

Single commit of all this year's work on the read-only firewall discovery playbook.

This PR moves the firewall playbook into the `utils` directory and updates it to:
- accept an FQDN (for example, `myproject-staging1.princeton.edu`)
- discover the IP of that FQDN
- use the IP to discover and display the name of the related object in our firewall
- use the firewall object name to discover and display all rules for that destination

This playbook can only run in Ansible Tower.


kayiwa (approved):
OK! Only one commit? ;-) Thanks @acozine

PR #4872, opened 2024-04-26
Create our nginx deny list in s3 for minimal maintenance

acozine:
Partial replacement for #4837.

This PR refines the deny list workflow:
Create the deny list from a template
Publish the deny list to an S3 bucket

This means less maintenance for the deny list, because there is no OS to upgrade or update, just a file.


acozine:
We need to update the firewall to consume the file from S3.

acozine:
Created #4888 to track the migration of the DENY rules from the firewall to the deny list.

acozine:
Created #4890 to track the decommissioning work in GCP.

PR #4873, opened 2024-04-26
Create an openBSD/GCP/nginx role

acozine:
Follow up to #4837.

This preserves the work done to create an OpenBSD server on GCP running nginx and makes it generic.

The denylist work will be handled on S3.


acozine:
We agreed we don't need to create this role. We are phasing out GCP as a platform. The only info we might use in future is about using `doas` with `become` for OpenBSD machines, and we don't have a current use case for that. We can always look back at this PR for that info.

PR #4874, opened 2024-04-26
[solr] Use systemd to manage the solr process

sandbergja:
This will automatically start solr up again if the process is ended (for example, by the oom_killer script)

This has been run on all solr-prod8 and solr-staging8d boxes.


tpendragon (approved):

kayiwa (approved):
OK! Thanks @sandbergja @tpendragon

PR #4875, opened 2024-04-29
[bibdata] Only run the poller daemon on one server per environment

maxkadel:
I believe running the poller daemon on multiple servers per environment is causing Postgres database locks, because four servers are trying to create the same exact object at the same time.

This PR limits installing the daemon on one machine, but it does not remove it from machines where it is already installed.

See also https://github.com/pulibrary/bibdata/pull/2362


christinach (approved):
thanks @maxkadel

PR #4884, opened 2024-05-01
[bibdata/nfs] allow access to NFS servers

kayiwa:
- **add bibdata staging to allow list**
- **NFS servers require IPs to allow connections**


maxkadel (approved):
Thanks @kayiwa !

PR #4885, opened 2024-05-01
[bibdata] Cludge to create consistent working systems with consistent rubygems and bundler

maxkadel:
We've had continuing difficulties with getting Passenger, Rubygems, Ruby, and Bundler to play nicely with each other.

This is what worked to get the new bibdata boxes into a state where you can just deploy to them and they work, without having to run anything on the command line.

It is not pretty. But it works. For now.

### DIFF
--- a/roles/bibdata/tasks/main.yml
+++ b/roles/bibdata/tasks/main.yml
@@ -8,17 +8,21 @@
   ansible.builtin.include_tasks: mounts.yml
 
 - name: bibdata | update rubygems
-  ansible.builtin.command: sudo gem update --system 3.5.7
-  changed_when: false
+  ansible.builtin.command: gem update --system 3.5.7
+  become: true
+  register: update_rubygems
+  changed_when: '"Latest version already installed. Done." not in update_rubygems.stdout'
   tags: rubygems
 
 - name: bibdata | install default bundler
-  ansible.builtin.command: sudo gem install --default bundler:2.5.7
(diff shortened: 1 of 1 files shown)


roles/bibdata/tasks/main.yml
```diff
@@ -6,3 +6,19 @@
 
 - name: bibdata | include tasks to mount disks
   ansible.builtin.include_tasks: mounts.yml
+
+- name: bibdata | update rubygems
+  ansible.builtin.command: sudo gem update --system 3.5.7
```
sandbergja, line 11: Does it still work if we use `become:` instead of sudo?

maxkadel: I believe so! 

I also tried to make the "changed_when" more informative

roles/bibdata/tasks/main.yml
```diff
@@ -6,3 +6,23 @@
 
 - name: bibdata | include tasks to mount disks
   ansible.builtin.include_tasks: mounts.yml
+
+- name: bibdata | update rubygems
+  ansible.builtin.command: gem update --system 3.5.7
+  become: true
+  register: update_rubygems
+  changed_when: '"Latest version already installed. Done." not in update_rubygems.stdout'
```
sandbergja, line 14: Cool!

sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4889, opened 2024-05-02
[prancible] unban Francis

kayiwa:
we remove the ban on Francis and leave it as example documentation


VickieKarasic (approved):
Looks good - thanks @kayiwa

PR #4897, opened 2024-05-06
[bibdata] switch to new production boxes

maxkadel:
- Point load balancer to new boxes
- Make sure vars point to new boxes
- Remove deprecated boxes from inventory


christinach (approved):
Thanks @maxkadel

PR #4900, opened 2024-05-06
Fixup some tencent ranges

joecorall:
Broaden the CIDR range based on whois records of the IP ranges. These are just a handful I've seen actively disrespecting our robots.txt - there may be other ranges in this list for tencent that need broadened from `/32`.

I tried seeing about automating the list of ranges from tencent. Though Tencent Cloud does not provide a list of ranges like AWS or gcloud (i.e. https://ip-ranges.amazonaws.com/ip-ranges.json and https://www.gstatic.com/ipranges/cloud.json). I tried querying whois based on their `mnt-by: MAINT-ACEVILLEPTELTD-SG` but the number of records returned is larger than whois' default record limit so would need bulk access to get that.


kayiwa (approved):
OK! Thank you @joecorall

PR #4901, opened 2024-05-06
[prancible] move Deny001-Deny005 from panOS to AWS denylist

VickieKarasic:
Affiliated ServiceNow ticket to remove rules from PanOS: TKT0419458


kayiwa (approved):
OK. Thanks @VickieKarasic @Beck-Davis

PR #4902, opened 2024-05-06
Moved marquand-catalogs to static-tables

rladdusaw:
PR pending renaming the repo

### DIFF
--- a/roles/libstatic/defaults/main.yml
+++ b/roles/libstatic/defaults/main.yml
@@ -102,12 +102,12 @@ sites:
       options: 'Indexes FollowSymLinks MultiViews'
       version: 'main'
       alias: 'pathebaby'
-    - doc_root: '/var/local/marquand_catalogs/site'
-      repo_root: '/var/local/marquand_catalogs'
-      git_repo: 'https://github.com/pulibrary/marquand-catalogs.git'
+    - doc_root: '/var/local/static-tables/site'
+      repo_root: '/var/local/static-tables'
+      git_repo: 'https://github.com/pulibrary/static-tables.git'
       options: 'Indexes FollowSymLinks MultiViews'
       version: 'main'
-      alias: 'marquand_catalogs'
+      alias: 'static-tables'
 deploy_keys:
     - mssimages_ed25519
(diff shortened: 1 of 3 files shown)


christinach (approved):
Thanks @rladdusaw

christinach (approved):

rladdusaw:
I don't think we need both, I've changed some things to underscores to better match the files they are in. 

I think the /marquand_catalogs was just being used to make the rewrite rules shorter, and was only linked to in the repo. I'm fine with preserving it but I believe the two rewrite rules were what we were actually using for production sites.

maxkadel:
> I don't think we need both, I've changed some things to underscores to better match the files they are in.
> 
> I think the /marquand_catalogs was just being used to make the rewrite rules shorter, and was only linked to in the repo. I'm fine with preserving it but I believe the two rewrite rules were what we were actually using for production sites.

I would actually prefer for the url for the site be static-tables, since that's the convention for URLs.

Am I misunderstanding you, or do you think that going to https://library.princeton.edu/marquand_catalogs/ would take you to https://library.princeton.edu/static_tables without the added rewrite rules?

rladdusaw:
> I would actually prefer for the url for the site be static-tables, since that's the convention for URLs.
> 
> Am I misunderstanding you, or do you think that going to https://library.princeton.edu/marquand_catalogs/ would take you to https://library.princeton.edu/static_tables without the added rewrite rules?
> 
> edit: I am not the strongest on rewrite rules, so I could be wrong, but I would _think_ that it would be needed, and that that would be good to have in case people had bookmarked the site, or linked to it directly in a LibGuide or something like that.

I agree that marquand-catalogs would be better for the url, but I don't like adding a bunch of rewrite rules that are not going to get used.

The rewrite rule would be required to take you from /marquand_catalogs to /static_tables. I think the original /marquand_catalogs url was created as a side effect of making a reusable location to redirect the /marquand/resources/sales-catalogs and /libraries/marquand/resources/sale urls to. I don't think it's being used outside the dev team linking to it in our documentation.

maxkadel (approved):

PR #4903, opened 2024-05-06
[panos] add verbosity

kayiwa:
we want to expand the results and filter again

Co-authored-by: Beck Davis <beck-davis@users.noreply.github.com>
Co-authored-by: John Kazmierski <jkazmier-PUL@users.noreply.github.com>


playbooks/utils/firewall_update.yml
```diff
@@ -41,6 +41,10 @@
       ansible.builtin.set_fact:
         filter_input: "{{ filter_condition }}{{filter_ip }}"
 
+    - name: gives us all the results
+      ansible.builtin.debug:
+        msg: filter_input
```
acozine, line 46: This var will only have the combination of the condition and the IP - right now that would look like `value contains 128.112.000.00`. If you move this task down under the next task, and use `test_object` as the variable, then you'll see all the results from the firewall.

acozine (commented):
Good idea to add visibility into the full set of results. See below for details on how to get the best results.

PR #4905, opened 2024-05-07
[security theater] OIT apps

### LINKED ISSUE #4894: Can we uninstall and reinstall Security Theater

The ir_agent from Rapid7 and falcon sensor are not registering as installed on BigFix. 

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

- We did a report on BigFix which suggested that rapid7 and crowdstrike are not properly installed. The remove software works successfully but leaves zombie-fied systemctl processes which our playbook depends. Our playbook will need to be modified to either search for another criteria to ensure that the software full installs. (it currently skips all the tasks with

```bash
skipping: [sandbox-acozine.lib.princeton.edu] => {"changed": false, "false_condition": "not rapid7_home.stat.exists", "skip_reason": "Conditional result was False"}
```
with a false condition

### Acceptance criteria
 - [x] Can see Rapid7 and Crowdstrike on BigFix
 - [x] Playbook successfully unistalls and reinstalls software

kayiwa:
the Rapid7 uninstaller fails to clean up after itself. Any subsequent
attempts to reinstall rapid7 on a VM leaves the application systemd
process registered.
Removing the directory allows us to have these conditional step
Crowdstrike is left registered so it is easier to reinstall

closes #4894

### DIFF
--- a/playbooks/utils/remove_rapid7.yml
+++ b/playbooks/utils/remove_rapid7.yml
@@ -15,6 +15,10 @@
       args:
         chdir: /opt/rapid7/
 
+    - name: completely remove rapid7 directory
+      ansible.builtin.file:
+        path: /opt/rapid7
+        state: absent
   # post_tasks:
     # - name: send information to slack
       # ansible.builtin.include_tasks:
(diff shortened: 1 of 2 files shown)


kayiwa:
successfully removed and installed on all sandboxes (which showed up on big fix)

bess (approved):

PR #4907, opened 2024-05-07
Move rules Deny011 through Deny015 from Panorama to AWS template

Beck-Davis:
# Co-authored-by: Vickie Karasic <vickiekarasic@users.noreply.github.com>


VickieKarasic (approved):
Looks good @Beck-Davis - thanks!

PR #4912, opened 2024-05-07
[prancible] create a new role

### LINKED ISSUE #4910: Create a new role documentation does not work
### Introduction

molecule no longer provides `init role`, so the [documented role creation](https://github.com/pulibrary/princeton_ansible?tab=readme-ov-file#create-a-new-role) does not work. 

Here is the current molecule documentation: https://ansible.readthedocs.io/projects/molecule/getting-started/

Figure out the new right way to create a new role and set it up to use molecule. Update the README with these steps.

### Acceptance criteria

- [ ] The README has a step by step guide: Provides the exact steps to guide the user on how to do a task.
- [ ] It includes references: References any related resources that have assisted creating this documentation (links to other tickets, online articles etc.).

kayiwa:
use the `ansible-galaxy` tool now that molecule can no longer initialize a new role

closes #4910


hackartisan (changes requested):
Thank you, this seems like a good start! I think this does not create the `molecule/default` folder with converge, destroy, etc. Here is the one we worked on today, for example: https://github.com/pulibrary/princeton_ansible/commit/328ddefc275b929bd1c4a8f35ea2ba043672fe01

Should we add steps to generate the default molecule files?

Also this doesn't seem to add `.ansible-lint` or `.yamllint` so if those are important we Should add steps for those, also.

kayiwa:
If you CTRL + f for  `copy over the root molecule.yml` from main you will see the steps for `.ansible-lint` and `.yamllint`

hackartisan (approved):

PR #4915, opened 2024-05-08
Move rules Deny016 through Deny025 from Panorama to AWS template

Beck-Davis:


kayiwa (approved):
OK! Thanks @Beck-Davis

PR #4917, opened 2024-05-08
[denyhosts] add a description field

kayiwa:
- **correct YAML markup on a previous PR**
- **place holder for adding descriptions in templates**
- **we may need to YAML'ize it but this unbreaks the FW**


Beck-Davis:
Thank you @kayiwa !

PR #4918, opened 2024-05-08
[postfix] replacement ponyexpr

### LINKED ISSUE #4891: Make postfix relay allow list easily automatable and trackable
Follows up on #4736. 

### User story

As an operations person:
- I want to know at all times which VMs are authorized to send mail using our postfix relay.
- When I build a new postfix server, I want to be sure that all currently authorized VMs are included in the configuration of the new postfix server.
- When I create and delete VMs, I want to update the list of authorized VMs easily with Ansible.
- The list of VMs that can use our postfix relay to send email should be tracked under source control, and the configuration should be easy to automate.

### Acceptance criteria

- [x] I can build a brand-new postfix relay server and all authorized VMs will be in the postfix config.
- [x] I can remove a VM from the list of authorized VMs and run a playbook to remove that VM from the postfix config.
- [x] I can add a VM to the list of authorized VMs and run a playbook to add that VM to the postfix allow list.
- [x] Later I can look in GitHub and see when any particular VM was added to the configuration list or removed from it.

- [x] Move the `lineinfile` tasks that update the postfix config from `tasks/main.yml` to a new file called something like `tasks/update_config.yml`.
- [x] Include/import the new tasks file into the main tasks file
- [x] Create an ansible template for the postfix configuration file that lists authorized VMs
- [x] Create a list variable in `vars/main.yml` that includes at least one VM
- [x] Replace the `lineinfile` tasks with a `template` task that uses the template you created and the list variable to generate the allow-list config file
- [x] Add a tag to the include/import (check the docs on tag inheritance for which one we want to use here) so we can run the playbook with `all` tags (the default) to build a new postfix relay server or with `-t update_config` to update the allow list for the postfix relay (make the tag name match what we've used in other playbooks/roles for this purpose)

kayiwa:
- **template a file instead of copy a file**
- **add empty relay host address**
- **use a template instead of lineinfile module**
- **add pointers on how to add a VM to the list of allowed relays**
closes #4891

### DIFF
--- a/playbooks/postfix.yml
+++ b/playbooks/postfix.yml
@@ -1,6 +1,8 @@
 ---
 # by default this playbook runs in the staging environment
 # to run in production, pass '-e runtime_env=production'
+# most common use case is `ansible-playbook playbooks/postfix.yml -t update_config`
+# after adding the IP address of your vm in `vars/main.yml`
 - name: build mailrelay to proofpoint
   hosts: postfix_{{ runtime_env | default('staging') }}
   remote_user: pulsys
--- a/roles/postfix/README.md
+++ b/roles/postfix/README.md
@@ -7,14 +7,9 @@ Creates a relay to proofpoint and creates a relay for our applications
 Role Variables
 --------------
 
-Most use for this role will involve adding your IP address to the allow list with the following variable:
+Most use for this role will involve adding your IP address to the allow list of our relay. Edit the [vars/main.yml](vars/main.yml)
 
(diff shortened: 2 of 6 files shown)


maxkadel (approved):
I like it!

Definitely not-blocking suggestion - "mynetworks" is a bit generic - is there something more descriptive we could use?

kayiwa:
Keeps it consistent with postfix configuration language

PR #4919, opened 2024-05-09
Try copying over the whole example role

hackartisan:
deleted `ansible-lint` because there was already an `.ansible-lint` deleted `example/playbook.yml` because it's not referenced anywhere in the readme and i wonder if it's been maintained in line with our changing conventions.

@kayiwa I dug in a little more and made this PR against your PR. I ran through it and it does work. What do you think? Thanks!


kayiwa (approved):
OK! Thanks @hackartisan

PR #4927, opened 2024-05-10
Make use of yarn for pkg management optional

maxkadel:
We may sometimes want to use npm as our package manager (as for the new static_tables app), and we don't want to install yarn if we don't need it.


rladdusaw (approved):

kayiwa (approved):
OK! Thanks @maxkadel

PR #4939, opened 2024-05-13
add an easy-to-find message for nginx restarts

acozine:
We've been seeing connection failures for both Solr and PostgreSQL, and we're trying to eliminate the ones that have known causes. One known cause is Patch [Mon/Tues]day. The other is load balancer restarts. 

This PR adds a special message for the nginx restart play, which would let us track that known cause more easily.

### DIFF
--- a/playbooks/nginxplus.yml
+++ b/playbooks/nginxplus.yml
@@ -41,7 +41,7 @@
   roles:
     - role: ../roles/nginxplus
 
-- name: restart nginx with updated loadbalancer configuration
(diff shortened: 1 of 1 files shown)


acozine:
Thanks to @VickieKarasic for the idea!

VickieKarasic (approved):
This looks great - thanks, @acozine!

sandbergja (approved):
Thanks, @VickieKarasic and @acozine ☕ 🐴

PR #4946, opened 2024-05-15
[CHECKMK] Adding a bash script for monitoring rails health pages

carolyncole:
Screenshot 2024-05-15 at 2 25 32 PM


carolyncole:
Show failure when I deployed a health check that always fails https://github.com/pulibrary/pdc_describe/blob/health-monitor-fail/config/initializers/health_monitor.rb#L19
Screenshot 2024-05-15 at 3 00 50 PM

PR #4947, opened 2024-05-15
[pulibrary/static-tables#26] Add tags in static-tables main task

christinach:


sandbergja (approved):
🐬 Thanks @christinach

PR #4948, opened 2024-05-15
[passenger] Add a variable to set number of passenger processes

hackartisan:
I ran the figgy staging role with and without this variable defined. When not defined, it continues to use the passenger defaults. when defined, it sets the number of processes and warms them.

closes pulibrary/figgy#6398


hackartisan:
this was run on figgy prod

acozine:
Looks good - not a blocker, but if we're going to keep the number of processes the same across envs, could we define it in a shared vars file?

hackartisan:
@acozine could we talk about doing that beyond just this var? There isn't currently a shared vars file for figgy servers.

PR #4949, opened 2024-05-15
Add tree as a default package

hackartisan:
because it is the best and I am constantly installing it


kayiwa (approved):
OK! Thanks @hackartisan

PR #4960, opened 2024-05-20
[static_tables] Use marquand-catalogs for catalog path

maxkadel:
Connected to https://github.com/pulibrary/static-tables/issues/130

Run successfully on both load balancers.


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4962, opened 2024-05-20
[static_tables] Add more new sites

maxkadel:
Update previous sites to use /special-collections/databases/ path


christinach (approved):
Thanks @maxkadel

PR #4964, opened 2024-05-21
[static-tables] Add alumni index part 2 to libwww proxy pass rules

sandbergja:
Part of pulibrary/static-tables#44


christinach (approved):
Thanks @sandbergja

Issue #4858, opened 2024-04-16
Upgrade production Solr and Zookeeper boxes


acozine:
We want to end up with:

- Jammy for the OS
- Solr 8

Other requirements?

acozine:
We are taking a "rolling update" approach, adding a Jammy box to the cluster, creating replicas for everything on it, then deleting the replicas from an old box and decommissioning that box. Repeat two more times until all boxes in the cluster are running Solr 8 on Jammy. 

We also need to update [this config](https://github.com/pulibrary/pul_solr/blob/1eb72f24b161e4f0b55c277ae9999fbb923c6dc9/config/deploy/production.rb#L1) in the pul-solr repo so that changes to the configsets are deployed to a Jammy box.

acozine:
As of today, all Solr8 boxes are running on Jammy. See #4944 for the necessary changes.

Zookeeper next.

acozine:
Now that Solr has all been upgraded, we can decommission the Bionic boxes:

- [ ] lib-solr-prod4
- [ ] lib-solr-prod5
- [ ] lib-solr-prod6

See our [decommissioning documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/decommissioning.md) for more details.

acozine:
Procedure for swapping out zookeeper VMs:
- create 3 new VMs with the latest OS
- update the zookeeper inventory to remove one old VM and add one new VM
- update the `vault_lib_zk#_host` variable in the solr8 group vars vault file - these vars are used by both Solr and Zookeeper
- run the zookeeper_solr8 playbook to install and configure zookeeper on the new VM and update the zookeeper config on the two old VMs that are still in the zookeeper trio
- run the solr8cloud playbook to update the solr config to point to the new zookeeper trio
- stop the zookeeper service on the VM you removed from inventory & vars
- restart Solr on all 3 Solr servers
- tunnel into Solr and confirm that the Solr cloud is using the correct zookeeper instances
- repeat for the 2nd and 3rd new zookeeper VMs (remove an old one & add a new one in the inventory and the group vars each time)
- Troubleshooting: each zookeeper VM should have a unique ID in `/etc/zookeeper/conf/myid`; each zookeeper VM should have three correct zookeepers in `/etc/zookeeper/conf/zoo.cfg`, each Solr VM should have three correct zookeepers in `/etc/default/solr.in.sh`

acozine:
[...]

PR #4969, opened 2024-05-22
updates inventory and solr8 vars for jammy zk boxes

acozine:
Partially addresses #4858.

We have built new Jammy-based zookeeper boxes for production Solr8. Once these are in use, the entire Solr8 cluster will be running on Jammy-based boxes, and we can decommission the old Bionic VMs.

Do not merge until we have run the Solr8 playbook and confirmed that everything is running well.


maxkadel (approved):
Thanks @acozine! 🏇🏼

PR #4970, opened 2024-05-22
[prancible] moving Deny070 through Deny079 from PanOS to AWS DenyList

VickieKarasic:


kayiwa (approved):
OK! Thanks @VickieKarasic

PR #4972, opened 2024-05-22
[prancible] adding Deny0065 through Deny069 from PanOS to AWS DenyList

VickieKarasic:


kayiwa (approved):
OK! Thanks @VickieKarasic

VickieKarasic (commented):
Thanks @kayiwa

PR #4975, opened 2024-05-23
[static_tables] Add two more proxy-pass rules

maxkadel:
Re-order so they're in alphabetical order. Add last two sites.

Run successfully on the load balancers.


christinach (approved):
Thanks @maxkadel

Issue #4982, opened 2024-05-24
create a new unattended replacement vm


kayiwa:
### User story

As a dev I would like to be able to delete the virtual machine and create a new one in its place without calling on the operations team to assist me

### Acceptance criteria

- [x] I can run a playbook from Ansible Tower to delete a my bad vm
- [x] I can run a playbook from Ansible Tower to create its replacement vm

kayiwa:
the [community vmware_guest module](https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_guest_module.html) will create the new VM

kayiwa:
Getting [VM information](https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_vm_info_module.html)

kayiwa:
We will need to create a EE that has pyvnomi and vmware collection

kayiwa:
https://github.com/pulibrary/princeton_ansible/pull/4702 (for the EE)

kayiwa:
Is it possible we have encountered [this bug](https://github.com/ansible/ansible/issues/64774)

kayiwa:
this is largely done or certainly the requirements have been met.

PR #4984, opened 2024-05-25
I4982 replacement vm

kayiwa:
This is a playbook to be run from Tower to be used to replace an existing VM


hackartisan:
refs #4982

playbooks/utils/replace_vm_host.yml
```diff
@@ -0,0 +1,75 @@
...
+      community.vmware.vmware_vm_info:
+        hostname: '{{ vcenter_hostname }}'
+        username: '{{ vcenter_username }}'
+        password: '{{ vcenter_password }}'
+        vm_name: "{{ vm_to_replace }}"
+      facts:
+        - mac_address
+        - uuid
+        - datastore_url
+      register: vm_info
```
hackartisan, line 24: Can we print this info? If something errors after deletion such that the new VM isn't created, I think we would lose these values.

hackartisan (changes requested):

playbooks/utils/replace_vm_host.yml
```diff
@@ -0,0 +1,75 @@
+---
+#  This playbook will delete a VM and replace it. This is the only warning you will get. Running this playbook has destructive consequences
```
maxkadel, line 2: Can we put more warnings in Ansible Tower? Maybe even one of those annoying things where you have to type something into a box in order for it to really run?

acozine (approved):
Good first steps - we will revise as we learn more.

playbooks/utils/replace_vm_host.yml
acozine: Addressed in b8753ad8d70698d4e8ce093d9ccbc4c4fe330bce

PR #4991, opened 2024-05-28
Move Deny037 through Deny046 from PanOS to AWS DenyList

Beck-Davis:


VickieKarasic (approved):
Looks good, @Beck-Davis - thanks!

PR #4995, opened 2024-05-29
[prancible] unbreak CI

kayiwa:
the firewall rule only needs to be blocked on an actual server
clean up some linting


VickieKarasic (approved):
Thanks @kayiwa!

PR #5006, opened 2024-05-30
Updating location of cicognara-static

escowles:
We reorganized the Cicognara repos into a single repo, so the static website is now in the new repo.


kayiwa (approved):
OK! Thanks @escowles

kevinreiss (approved):

PR #5010, opened 2024-05-30
[Nomad] Disable app protect for nomad.

tpendragon:
It submits JSON sometimes, which app protect doesn't like.


kayiwa:
Heya @tpendragon if we add logging (from the WAF) will this be worth fixing on the nomad configuration to make it behave properly?

tpendragon:
@kayiwa Uhh, I don't know anything about app_protect to the extent I don't know what you mean, but I'm willing to learn!

PR #5018, opened 2024-06-04
Fix Solr backups

acozine:
Partial fix for https://github.com/pulibrary/pul_solr/issues/407.

The diglibdata mount for Solr backups was owned by `root:root` even though we were correctly setting the owner and group on the directories, because the mount process did not specify a GID or UID.


kayiwa (approved):
OK! Thanks @acozine

PR #5024, opened 2024-06-07
I4982 replacement vm two

kayiwa:


kayiwa:
This work is blocked upstream and can be revisited. We have been unable to get the playbook to pick up the MAC address of the VM and replace it.

Issue #5013, opened 2024-06-03
[allsearch_api] Create shared Solr variables for the Allsearch API


maxkadel:
### User story
The Allsearch API reads from the Solr collection for a number of our applications. We have these variables in a config file in the application right now; however, it is easy for these to drift from the correct values (see, e.g., there were almost three months between when the [staging solr url was changed for DPUL](https://github.com/pulibrary/princeton_ansible/pull/4768/files) and when we [updated it in the Allsearch API](https://github.com/pulibrary/allsearch_api/pull/215).

We need a way to keep this information aligned and avoid drift. The source of truth for these values for the applications whose information we're consuming is Prancible, so this is the most logical place to create the solution.

### Acceptance criteria
Applications include:
- [x] Catalog / Orangelight
- [x] DPUL
- [x] FindingAids
- [x] pulmap

- [x] The solution is acceptable to the teams maintaining the applications above

### Implementation notes, if any
Would it be possible to create a separate solr config file in each application's group_vars directory, and each of those could be loaded both for the individual application and the allsearch api?

May want to either add a comment to let folks know that the allsearch variables need to be updated, or possibly add a handler to update them.

acozine:
Discussed at Ansible Open House today. We wondered if it would make sense to create a consolidated `group_vars/solr/all_solr.yml` file (name up for debate) that would include all settings for Solr across all projects and all Solr versions. Then we could use ENV_VARS in the allsearch repo.

acozine:
Probably the next step is a draft PR so we can figure out what we have forgotten . . .

maxkadel:
Closing in favor of #5045

PR #5025, opened 2024-06-07
[allsearch_api] Use environment variables for Solr

maxkadel:
Run successfully on staging and production

This is a first step to ensuring we keep the allsearch api in sync with the applications it consumes

Connected to #5013


christinach (approved):
Thanks @maxkadel

PR #5031, opened 2024-06-12
cdh-test-prodigy updates: proxy port 8080 on upstream; remove circular redirect

rlskoeser:
I added port 8080 to the upstream declaration, because the service is running on port 8080 - this is an ASGI application, so we can't run it through nginx/passenger. 

I'm getting a circular redirect when I try to access the public url for the test site. The block I removed looks like the culprit to me, but I did test this on the active load balancer and reloaded nginx but it didn't solve the problem.


kayiwa (approved):
OK! Thanks @rlskoeser

acozine:
Deployed today, the circular redirect is fixed. Huzzah!

PR #5040, opened 2024-06-17
[approvals] install specific rubygems and default bundler versions

sandbergja:
I ran this on staging and prod today


christinach (approved):
🍄

PR #5044, opened 2024-06-18
I5013 share solr vars

maxkadel:
Allsearch-api uses the Solr of certain apps directly (catalog, dpul, pulfalight, maps), so we need to find a way to keep these in sync. 

Since they are all rails apps, we don't want allsearch-api to use their entire config files, because there are conflicting variables, so splitting solr into its own config files by environment.

Connected to #5013


maxkadel:
@tpendragon  - would you mind reviewing this?

tpendragon (approved):

sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #5053, opened 2024-06-20
Lockers fix bundler

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #5056, opened 2024-06-21
utility playbook that will install podman engine

### LINKED ISSUE #5054: Manually Install Podman
### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- RHEL does not update podman in Ubuntu registries (even for security patches 😞 ) Can we have a playbook that will install the latest podman engine?

### Acceptance criteria
 - [ ] Playbook will install the most current version of podman on an endpoint

kayiwa:
this playbook will install podman to an endpoint and create a symbolic link to docker

Closes #5054

### DIFF
--- a/playbooks/utils/install_podman.yml
+++ b/playbooks/utils/install_podman.yml
@@ -1,0 +1,82 @@
+---
+#  This playbook will install the podman container engine
+#     
+- name: Install podman engine
+  hosts: staging:qa:production
+  remote_user: pulsys
+  become: true
+  vars:
+    - podman_version: "v5.1.1-1"
+    - podman_static_url: "https://github.com/mgoltzsche/podman-static/releases"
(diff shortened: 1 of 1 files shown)


tpendragon:
For Nomad, podman needs a systemd that runs a "system socket activation unit": https://developer.hashicorp.com/nomad/plugins/drivers/podman#client-requirements. For a package install this means that post-package there needs to be a `podman` service installed via systemd, which presumably (?) runs that API service.

https://github.com/mgoltzsche/podman-static/issues/45 seems to imply to me that the static binaries won't work for that use case. :(

tpendragon:
I don't think we can use this, should I close it @kayiwa ?

kayiwa:
we have elected to have container applications should be deployed via rocky linux

PR #5057, opened 2024-06-21
[lib-jobs] Add airtable personal access token for our service account

sandbergja:
This token has "data.records:read" permission for the "PUL Staff Directory for Website" base.

Instructions for how to rotate the token can be found in a comment in the vault.

Related to pulibrary/lib_jobs#697


christinach (approved):
Thanks @sandbergja 🍄

PR #5062, opened 2024-06-24
[globus] add a globus-check

### LINKED ISSUE #5061: [Globus] Restart globus-gridftp-server when Monit restarts apache2
### User story

When apache2 goes down on the globus services it is likely because the `globus-gridftp-server` has a problem.  We should retsart both `apache2` and `globus-gridftp-server` when apache2 goes down

### Acceptance criteria

- [ ] [This line](https://github.com/pulibrary/princeton_ansible/blob/main/roles/monit/templates/apache2_monit.conf.j2#L9) restarts `apache2` and `globus-gridftp-server`

kayiwa:
create a check for the globus-gridftp service. This service is dependent
on the apache2 service. If we restart apache2 we will also restart the
globus service

closes #5061


carolyncole (approved):
Thank you @kayiwa

PR #5070, opened 2024-06-25
[loadbalancer + allsearch-api] set the X-Real-IP header to prevent logs from being filtered out

### LINKED ISSUE #5069: [allsearch-api] Load-balanced traffic does not show up in nginx access.log

Set the `X-Real-IP` header, so that the passenger nginx configuration does not filter out all the logs.

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

Currently, we aren't getting any logs when users go to allsearch.princeton.edu.  If you `curl localhost` from within the boxes, it does appear in access.log, but if it comes via the load balancer, it gets filtered out.

### Acceptance criteria
 - [ ] Traffic to allsearch.princeton.edu shows up in /var/log/nginx/access.log
 - [ ] Traffic to allsearch.princeton.edu shows up in datadog

sandbergja:
Closes #5069 

I ran this on both load balancers today (25 June)

### DIFF
--- a/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
+++ b/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
@@ -46,6 +46,7 @@ server {
         # handle errors using errors.conf
         proxy_intercept_errors on;
         proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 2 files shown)


christinach (approved):
🍄

PR #5072, opened 2024-06-27
[Figgy] Add bitcur-archive as an ingest source.

tpendragon:
Closes pulibrary/figgy#6421

This also removes the cantaloupe mount - we don't use it anymore.

This has been run in staging and production.


kayiwa (approved):
OK! Thanks @tpendragon

PR #5079, opened 2024-06-28
[allsearch-api] Error statuses in JSON format

### LINKED ISSUE #4396: Adjust Error Handling for the All Search API
The allsearch api returns 400s with useful JSON data when an empty query is passed, and 500s if it encounters an error, but in both cases, it returns the generic PUL error page (see, e.g., https://allsearch-api.princeton.edu/search/catalog?query=). This may be an issue of needing to change behavior on the load balancer (in the Princeton Ansible NGINX config, probably something to do with the [errors.conf template](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/allsearch-api_prod.conf#L52)). We'd like to return a meaningful error message to client applications.

sandbergja:
This allows 400 and 500 JSON responses through from the application, and returns other errors in JSON format, rather than the standard "Something went wrong" html page.

Closes #4396

For background information on why we'd like to offer JSON errors, see [this ADR in allsearch_api](https://github.com/pulibrary/allsearch_api/blob/main/architecture-decisions/001_error_output.md)

I ran this on both load balancers today.

### DIFF
--- a/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
+++ b/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
@@ -51,5 +51,5 @@ server {
         proxy_cache allsearch-apicache;
     }
 
-    include /etc/nginx/conf.d/templates/errors.conf;
+    include /etc/nginx/conf.d/templates/errors-in-json-format.conf;
 }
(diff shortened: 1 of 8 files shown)


christinach (approved):
Thanks @sandbergja 🍄 !

PR #5080, opened 2024-07-01
Pull exact IPs in the firewall playbook

### LINKED ISSUE #4892: Firewall playbook sometimes returns the wrong object
### Expected behavior

When I run the firewall playbook and pass any FQDN, it retrieves the IP for that FQDN and finds the matching object in the firewall.

### Actual behavior

With certain IP addresses, the playbook displays the wrong object. See https://ansible-tower.princeton.edu/#/jobs/playbook/1493/output.

This happens when the last segment of the IP address is short. The task that queries the firewall is retrieving a list of matches (for example, if it's searching for `111.222.33.4` it matches `111.222.33.40` and `111.222.33.487` as well as the exact match) and our display is only presenting the first item in the list. 

This happens because we are using `value contains` plus the IP as our filter [here](https://github.com/pulibrary/princeton_ansible/blob/833abec2f664b314ce08372ad0bbb846c26a6345/playbooks/utils/firewall_update.yml#L34). In the example above, any IP with `.4x` or `.4xx` as the last segment will match, because they all contain `.4`. We used `value contains` because not all objects in the firewall have `/32` at the end of the IP definition.

### Steps to replicate

Run the firewall playbook and pass `pulcheck-staging1.princeton.edu` as the VM you're curious about. 

### Impact of this bug

We can't trust the output of the firewall playbook in all cases.

We can change the playbook to display all matches. Or we can clean up our firewall data so all entries have `/32` and we can filter more exactly. If we want to do the latter, we can write a playbook that pulls all object data out of the firewall, then go through it to generate a list of IPs that need `/32` added to the end. We might even try automating that update as our first non-read-only firewall task.

acozine:
Closes #4892.

We have cleaned up the data in the firewall enough now that we can pull exact IP addresses (with `/32` at the end). This change will find the exact IP and return rules for it.

Still fails if the FQDN is not represented in the firewall at all. We can make a separate ticket for that if we decide to handle it.

### DIFF
--- a/playbooks/utils/firewall_update.yml
+++ b/playbooks/utils/firewall_update.yml
@@ -31,7 +31,7 @@
 
     - name: set filter condition
       ansible.builtin.set_fact:
-        filter_condition: 'value contains '
+        filter_condition: "value == "
 
     - name: set filter IP
       ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK! This is great @acozine

PR #5083, opened 2024-07-01
[cdh] add cdh-web3 and 4

### LINKED ISSUE #5081: VMs and load balancer configuration for pre-production instance of CDH website

We need a pre-production environment for the CDH website that can be run in tandem with the current production site for a period of one month or so while we do content entry and cleanup before switching the new version live.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

We expect to start the content entry in early to mid-July. Go-live is scheduled for August 20 (new version of the site needs to be up before the fall semester).

We are working on a major update to the CDH website that requires significant content editing and organization. We need time to enter and manage the content updates in a production environment before we switch the new version live.

### Acceptance criteria
 - [ ] production VM(s) available for deploy of new CDH web code
 - [ ] pre-production site accessible at cdh-beta.princeton.edu on VPN only
 - [ ] current production site remains functional and accessible at cdh.princeton.edu 
 - [ ] we can easily switch pre-production site to the live site 

I propose that we either temporarily split cdh-web1 and cdh-web2 to make one production and the other pre-production, or we temporarily add cdh-web3 and cdh-web4 for the duration of the time we need to run both versions of the site in production.  We can revert back to the previous setup (two production VMs only) after the new version is live and stable. (Archive/remove the old database, shutdown any extra VMs, etc.)

We already have the existing hostname/alias cdh-beta.princeton.edu from the previous migration. I would like that hostname used for the pre-production site; should be VPN accessible only.

I plan to handle the changes needed in cdh-ansible to handle deploy for the pre-production instance, including provisioning it with a new database separate from the current production site.

My thought in requesting production VMs was that we could easily make the new version live either by updating the load balancer configuration and/or by deploying the new code and application configuration to production VMs (e.g., push new code and config to look at new version of the db).

kayiwa:
create a new server block for the beta website that will point to the
new VMs. The cdh-beta name is already part of the SAN cert

closes #5081

### DIFF
--- a/roles/nginxplus/files/conf/http/cdh_prod_web.conf
+++ b/roles/nginxplus/files/conf/http/cdh_prod_web.conf
@@ -139,6 +139,10 @@ server {
         proxy_ignore_client_abort on;
         proxy_intercept_errors on;
         health_check interval=10 fails=3 passes=2;
+        # allow princeton network
+        include /etc/nginx/conf.d/templates/restrict.conf;
+        # block all
+        deny all;
     }
     include /etc/nginx/conf.d/templates/cdh-errors.conf;
 }
--- a/roles/nginxplus/files/conf/http/cdh_prod_web.conf
+++ b/roles/nginxplus/files/conf/http/cdh_prod_web.conf
@@ -1,6 +1,17 @@
 # Ansible managed
 proxy_cache_path /data/nginx/prod_cdhweb/NGINX_cache/ keys_zone=prod_cdhwebcache:10m;
 
+upstream beta_prod_cdhweb {
+    least_time header 'inflight';
+    zone prod_cdhweb 64k;
+    server cdh-web3.princeton.edu;
+    server cdh-web4.princeton.edu;
+    sticky learn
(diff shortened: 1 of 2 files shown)


roles/nginxplus/files/conf/http/cdh_prod_web.conf
```diff
@@ -80,6 +110,39 @@ server {
...
+        proxy_connect_timeout      2h;
+        proxy_send_timeout         2h;
+        proxy_read_timeout         2h;
+        proxy_max_temp_file_size 0;
+        proxy_buffer_size          128k;
+        proxy_buffers              4 256k;
+        proxy_busy_buffers_size    256k;
+        proxy_ignore_client_abort on;
+        proxy_intercept_errors on;
+        health_check interval=10 fails=3 passes=2;
```
rlskoeser, line 141: is it missing the config to restrict to campus IPs / VPN or am I just not seeing it?

rlskoeser (approved):

PR #5085, opened 2024-07-01
first attempt to redirect PUDL viewer URLs

### LINKED ISSUE #4566: Redirect PUDL viewer URLs
We redirected most of the PUDL URLs that contain ARK IDs so the old URLs will get users to the new location of the content. But we just had a report from a user that this URL wasn't working:

`https://pudl.princeton.edu/viewer.php?obj=2801pg40q#page/143/mode/2up`

we should redirect that pattern:

`pudl.princeton.edu/viwer?obj=[X]` -> `arks.princeton.edu/ark:/88435/[X]`

acozine:
Closes #4566.

Currently PUDL URLs with `viewer.php` return errors. We should be able to redirect them. Doesn't work yet, but hoping to figure this out soon.

### DIFF
--- a/roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
+++ b/roles/nginxplus/files/conf/http/templates/pudl_proxy_pass.conf
@@ -1,6 +1,8 @@
 # arks redirects
     rewrite ^/objects/(.*)$ https://arks.princeton.edu/ark://88435/$1 redirect;
 
+    rewrite ^/viewer.php?obj=(.*)$ https://arks.princeton.edu/ark://88435/$1 redirect;
+
     rewrite ^/sheetreader.php?obj=(.*)$ https://arks.princeton.edu/ark://88435/$1 redirect;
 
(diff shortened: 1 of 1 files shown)


acozine:
This did not work, it looks like we need to add a section for `args`. See this [ServerFault question](https://serverfault.com/questions/160790/nginx-rewrite-for-an-url-with-parameters) and this [blog post](https://ubiq.co/tech-blog/rewrite-url-parameters-nginx/).

sandbergja (approved):
Thanks, @acozine, @dphillips-39 , @Beck-Davis, @kayiwa , @tpendragon !

acozine:
Turns out we did not need to use `args` or a complicated regex in this redirect rule. Nginx always accepts `$arg_obj` to represent the first argument in a URL, and in this case there was only one argument.

The working rule redirects `^/viewer(.*)$` ("any URL that begins with pudl.princeton.edu/viewer") to the ARK URL and adds the object ID at the end.

PR #5090, opened 2024-07-02
Override the default org-wide issue templates

acozine:
Closes #5089.

Many of the org-wide issues templates make no sense in this repo. This PR includes the most relevant templates for issues in prancible.

Issue #5089, opened 2024-07-02
Override the org-wide issue template settings in this repo


acozine:
### Expected behavior
As a user, when I open an issue in princeton_ansible, I would like to see templates that help me create automation-related tickets.

### Actual behavior
When I open an issue in princeton_ansible, I see a lot of templates that do not apply to this repo. For example, `Feature` or `Accessibility` and so on. 

### Steps to replicate
Click on `Issues . . . add an issue` in this repository. 

### Impact of this bug
Users have to work harder to open relevant issues in this repo. First they have to decide what type of issue to open, which sometimes discourages them from continuing.

### Implementation notes, if any
Override the pulibrary org default issue templates in this repo.
We think the repo is inheriting the default settings from the `pulibrary` organization, possibly from [this repo](https://github.com/pulibrary/.github/tree/main/.github/ISSUE_TEMPLATE). Can we override those settings just for this repo?
See the GitHub docs [on configuring templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) and [on templates generally](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates).

PR #5091, opened 2024-07-02
Adds a maintenance template to the issue templates for this repo

acozine:
Follow up to #5090. See also #5089.

We decided the maintenance template is also useful.


dphillips-39 (approved):
Thanks, looks good

PR #5096, opened 2024-07-02
Proxy prodigy servers on port 80 instead of 8080

rlskoeser:
This switches the nginxplus config for cdh-prodigy and cdh-test-prodigy from proxying port 8080 to port 80

I discovered that I need to serve out content alongside the application, so I'm setting up nginx on the VMs to proxy the 8080 port and serve out static content.


kayiwa (approved):
OK! Thanks @rlskoeser

PR #5100, opened 2024-07-05
[globus] another restart for globus

kayiwa:
we will monitor the apache2 error logs. When monit matches on apache2
was restarted it will also restart globus

related to #5061

### DIFF
--- a/roles/monit/templates/monit.conf.j2
+++ b/roles/monit/templates/monit.conf.j2
@@ -15,4 +15,4 @@ check process {{ globus_ftp_service_name }} with pidfile /run/{{ globus_ftp_serv
 
 # Trigger globus-gridftp-server restart on apache2 restart and log the restart
 check file apache2_error_log with path {{ apache_error_log_path }}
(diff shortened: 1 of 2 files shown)


roles/monit/templates/monit.conf.j2
```diff
@@ -1,16 +1,15 @@
 {{ ansible_managed | comment }}
-check process {{ apache_service_name }}
-  with pidfile /var/run/{{ apache_service_name }}/{{ apache_service_name }}.pid
+check process {{ apache_service_name }} with pidfile /var/run/{{ apache_service_name }}/{{ apache_service_name }}.pid
   start program = "/usr/sbin/service {{ apache_service_name }} start"
   stop program  = "/usr/sbin/service {{ apache_service_name }} stop"
-  if failed port 80 protocol http
```
carolyncole, line 6: @kayiwa why did you remove the timeout information here.  I think that might still be needed to notice apache is having an issue.

carolyncole (approved):

PR #5129, opened 2024-07-08
[tigerdata] migrate vms

kayiwa:
migrate the names off the publicly routable IPS
move the loadbalancing configs to the staging loadbalancer

under separate cover run the incommon playbook to update TLS certs

Co-authored-by: Carolyn Cole <carolyncole@users.noreply.github.com>


carolyncole (approved):
Thanks @kayiwa

PR #5134, opened 2024-07-10
Add new cdh web pre-production VMs to nfs production access list

rlskoeser:
Adds cdh-web3 and cdh-web4 to production nfs access list

We'll have to remember to remove them when from this list when we remove the VMs


kayiwa (approved):
OK! Thanks @rlskoeser

Issue #5131, opened 2024-07-08
Researchdata and static-tables playbooks are broken


acozine:
### Expected behavior
These two playbooks (researchdata and static-tables) should install open-source nginx as a local webserver.

### Actual behavior
Both playbooks fail with the error:

```
TASK [nginxplus : Install: Set NGINX App-Protect Signing Key URL] ********************************************************
fatal: [prds-staging2.princeton.edu]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'nginx_app_protect_signing_key' is undefined. 'nginx_app_protect_signing_key' is undefined\n\nThe error appears to be in '/Users/cc62/apps_team/princeton_ansible/roles/nginxplus/tasks/keys/apt-key.yml': line 32, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Install: Set NGINX App-Protect Signing Key URL\"\n  ^ here\nThis one looks easy to fix. It seems that there is a value started\nwith a quote, and the YAML parser is expecting to see the line ended\nwith the same kind of quote. For instance:\n\n    when: \"ok\" in result.stdout\n\nCould be written as:\n\n   when: '\"ok\" in result.stdout'\n\nOr equivalently:\n\n   when: \"'ok' in result.stdout\"\n"}
```

If you add a line to include `/group_vars/nginxplus/main.yml`, providing access to the variable from the error above, the playbooks build an nginxplus load balancer on the prds-staging and static-tables VMs.

### Steps to replicate
Run the playbooks on staging.

### Impact of this bug
We cannot update any of the VMs for these project.

### Relevant links and code snippets, if applicable

```
```

### Implementation notes, if any
We think the playbooks were working correctly as of July 1st. They are broken as of July 8. The `nginx_type` variable should default to `opensource`, which should not run the task that errored out above.

kayiwa:
Should we need to use multiple CORS this [will be the way to do that](https://www.juannicolas.eu/how-to-set-up-nginx-cors-multiple-origins/)

maxkadel:
I believe allsearch-frontend works with tunneling because the Access-Control-Allow-Origin header is turned off entirely - see [vite.config.ts:8](https://github.com/pulibrary/allsearch_frontend/blob/a7a449cc9a9915d005dd94c7cc6aa96931a0bd61/vite.config.ts#L8).

[...]

PR #5141, opened 2024-07-11
[researchdata] use nginx role

kayiwa:
- **add nginx role**
- **make the research data role use nginx**

related to #5131


kevinreiss (approved):

roles/nfsserver/tasks/main.yml
```diff
@@ -16,19 +16,18 @@
...
     - "{{ tigerdata_fileshare_mount }}"
 
-- name: nfsserver | add conan group
-
+- name: Nfsserver | add conan group
   ansible.builtin.group:
     name: "conan"
 
-- name: deploy_user | create system user
+- name: Deploy_user | create system user
```
maxkadel, line 23:
```suggestion
- name: Nfsserver | create system user
```

maxkadel (approved):
Looks good! Thank you!

Made one small suggestion which does not need to block.

Question - do we need to add the new nginx role to any list to make sure it runs in CI? Or is that decided automagically in `run_molecule.py`

PR #5142, opened 2024-07-11
I5131 researchdata static tables

kayiwa:
Static Tables


kayiwa:
this work is superseded by #5151

PR #5151, opened 2024-07-12
decouple researchdata and static table

kayiwa:
we have moved research data into the main branch. This work focuses on static tables.


roles/static_tables/templates/default.conf.j2
```diff
@@ -1,12 +1,20 @@
...
 server {
     listen       80;
     server_name  localhost;
 
     #access_log  /var/log/nginx/host.access.log  main;
 
     location / {
-        add_header Access-Control-Allow-Origin {{ static_tables_cors_allow | default('https://library-staging.princeton.edu') }};
+        if ($cors != "") {
+          add_header 'Access-Control-Allow-Origin' "$cors" always;
```
acozine, line 16: Per discussion at the Open House, set this to *

maxkadel (approved):
Thanks so much for all your work on this @kayiwa! Looks good!

Issue #72, opened 2017-10-23
Add studio staging server to Plum/Figgy mounts


escowles:
See https://github.com/pulibrary/plum/issues/1424

escowles:
This is fixed in Figgy.

PR #5153, opened 2024-07-15
[ops-catchall #72] Remove nagios-staging.conf

Beck-Davis:
Completes https://github.com/pulibrary/ops-catchall/issues/72


kayiwa (approved):
OK! Thanks @Beck-Davis

PR #5170, opened 2024-07-17
fix whichiso deploy

Beck-Davis:


kayiwa (commented):
OK! Probably want a rails native to review before letting it go into main

group_vars/whichiso/main.yml
```diff
@@ -29,6 +29,8 @@ project_db_host: '{{postgres_host}}'
 install_ruby_from_source: true
 ruby_version_override: "ruby-3.2.4"
 passenger_ruby: "/usr/local/bin/ruby"
+passenger_extra_http_config:
```
VickieKarasic, line 32: Discussed in Ansible open hours: We believe this line may not be needed; this should be tested when @Beck-Davis gets back to looking at this PR.

PR #5175, opened 2024-07-18
[videoreserves] After CAS login on staging, redirect back to videoreserves domain

sandbergja:
Part of PrincetonUniversityLibrary/video_reserves#74

I ran this on staging today, July 18.


christinach (approved):
🍄

PR #5176, opened 2024-07-19
Adds a draft playbook to reboot a single VM

acozine:
As a developer, I want to be able to reboot a VM as needed. This PR will add that capability in Tower.

We can't test in Tower until a skeleton version of the playbook has been merged to the `main` branch, so we will merge this early and work out the details in another branch.


christinach (approved):
Thanks @acozine

PR #5179, opened 2024-07-19
Includes IP from Ansible facts when adding hosts to CheckMK

acozine:
For VMs in our private IP space, CheckMK needs an IP to fully monitor those VMs.

This PR uses a pre-existing variable in the CheckMK agent role to populate the IP from Ansible facts when adding hosts to CheckMK.

It also removes a debug task we used to confirm the regex that strips the `.princeton.edu` or `.lib.princeton.edu` from inventory names for CheckMK. We've tested that functionality and it works well. No need to debug forever.

### DIFF
--- a/playbooks/utils/checkmk_agent.yml
+++ b/playbooks/utils/checkmk_agent.yml
@@ -15,22 +15,13 @@
     - ../../group_vars/checkmk/{{ runtime_env | default('staging') }}.yml
     #- ../../group_vars/all/checkmk.yml  # this is the likely correct path when prod goes live
     - ../../group_vars/checkmk/vault.yml
+  vars:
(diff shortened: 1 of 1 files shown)


playbooks/utils/checkmk_agent.yml
```diff
@@ -15,15 +15,13 @@
     - ../../group_vars/checkmk/{{ runtime_env | default('staging') }}.yml
     #- ../../group_vars/all/checkmk.yml  # this is the likely correct path when prod goes live
     - ../../group_vars/checkmk/vault.yml
+  vars:
+    checkmk_agent_host_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
```
kayiwa, line 19: clever!

kayiwa (approved):
OK! Thanks @acozine

PR #5193, opened 2024-07-23
implements 'clean' step for yum repos

### LINKED ISSUE #5168: RHEL & Rocky boxes fail on update playbook
### Expected behavior
Our weekly OS package updates should succeed on all boxes.

### Actual behavior
The weekly update playbook fails on RHEL boxes. The package update itself succeeds, but a later task fails.

### Error messages
On the task `[RedHat | clear local repository]` we see:
```
fatal: [ansible-exec-node2.princeton.edu]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.legacy.dnf) module: clean. Supported parameters include: allow_downgrade, allowerasing, autoremove, bugfix, cacheonly, conf_file, disable_excludes, disable_gpg_check, disable_plugin, disablerepo, download_dir, download_only, enable_plugin, enablerepo, exclude, install_repoquery, install_weak_deps, installroot, list, lock_timeout, name, nobest, releasever, security, skip_broken, sslverify, state, update_cache, update_only, use_backend, validate_certs (expire-cache, pkg)."}
```

See also this [recent Tower job](https://ansible-tower.princeton.edu/#/jobs/playbook/3133/output).

### Steps to replicate
Run `playbooks/os_updates.yml` against the `ansible_tower` group.

### Impact of this bug
Not clear. The package updates succeed on RHEL machines, but the RHEL VMs never get rebooted. Also, over time the package cache may use a lot of disk space.

Here's a permalink to the [task that's failing](https://github.com/pulibrary/princeton_ansible/blob/ba891f1e0fa8c5d3920c13190a7050f54b741fd4/playbooks/os_updates.yml#L35).

The [old yum module docs](https://docs.ansible.com/ansible/3/collections/ansible/builtin/yum_module.html) say: 
`The yum module does not support clearing yum cache in an idempotent way, so it was decided not to implement it, the only method is to use command and call the yum command directly, namely “command: yum clean all”` and link to [this issue comment from 2017](https://github.com/ansible/ansible/pull/31450#issuecomment-352889579).
In the latest Ansible release, there is no `yum` module, there is only a `dnf` module. It does not have a `clean all` option either, see the [dnf module docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#ansible-collections-ansible-builtin-dnf-module).
We should discuss the best solution for this - two known options are: 
- change the task to run `sudo yum clean all` as a `command` task
- remove that task from the playbook and use some other approach to cleaning the cache for RHEL boxes

acozine:
Closes #5168.

Since the `yum` module does not have a `clean` option, uses a `command` task instead.

We ran into a problem trying to run this on Tower when the RHEL boxes did get updated - because the Tower job rebooted the VM on which the Tower job was running . . . So this PR now reboots only RH-family VMs that are NOT Tower-related, and sends a slack alert to the ops team to reboot the Ansible machines by hand.

### DIFF
--- a/playbooks/os_updates.yml
+++ b/playbooks/os_updates.yml
@@ -33,8 +33,8 @@
       when: ansible_os_family == "RedHat"
 
     - name: RedHat | clear local repository
-      ansible.builtin.yum:
-        clean: true
+      ansible.builtin.command:
+        cmd: sudo yum clean all
       when: ansible_os_family == "RedHat"
 
     - name: Ubuntu | remove dependencies that are no longer required
@@ -58,6 +58,12 @@
       register: reboot_required_file
       when: ansible_os_family == "Debian"
 
+    # - name: RedHat | Note RH machines that need reboots
+    #   ansible.builtin.set_fact: needs_reboot=true
+    #   when:
+    #     - ansible_os_family == "RedHat"
+    #     - yum_update_status.changed
+
     - name: RedHat | Reboot if required
       ansible.builtin.reboot:
         msg: "Reboot initiated by Ansible because of yum updates."
@@ -69,6 +75,7 @@
       when: 
         - ansible_os_family == "RedHat"
         - yum_update_status.changed
+        - "'ansible' not in inventory_hostname"
 
     - name: Ubuntu | Reboot if required
       ansible.builtin.reboot:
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK. Thanks @acozine

playbooks/os_updates.yml
```diff
@@ -87,3 +94,15 @@
         token: "{{ vault_pul_slack_token }}"
         msg: "Ansible ran `{{ ansible_play_name }}` on {{ inventory_hostname }}"
         channel: "{{ slack_alerts_channel }}"
+
+  post_tasks:
+    - name: notify ops to reboot the ansible servers if needed
```
kayiwa, line 99: we will soon be getting Rocky VMs and this will get them also.

kayiwa (commented):
May need to create an issue on this for Rocky Linux VMS

playbooks/os_updates.yml
acozine: as long as the Rocky VMs return `RedHat` as their `os_family` they should be updated by the `clear local repository` task and rebooted by the `Reboot if required` task

acozine: Confirmed today that the Rocky VMs do report their `os_family` as `RedHat`, so they are updated correctly.

kayiwa (approved):
OK! Thanks @acozine

PR #5196, opened 2024-07-23
5114 allsearch health checks

### LINKED ISSUE #5114: [allsearch-api] Use health.json for health checks

When allsearch-api's health checks are failing (health.json), it should not serve that application to the public.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria
 - [x] When one node is broken it does not get traffic diverted to it.
 - [x] It checks `/health.json?providers[]=database&providers[]=solr`

Existing health check configuration: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/imagecat_staging.conf#L35

### First Step

rladdusaw:
Resolves #5114

### DIFF
--- a/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
+++ b/roles/nginxplus/files/conf/http/allsearch-api_prod.conf
@@ -43,6 +43,7 @@ server {
 
     location / {
         proxy_pass http://allsearch-api;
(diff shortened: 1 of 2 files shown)


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛ 🐴 🐬

PR #5201, opened 2024-07-24
Supports rebooting a VM from Tower

acozine:
Follows up on #5176.

Once this is merged, Tower users can view VM details (host/datacenter and storage locations) for any VM in our inventory. By default the Tower template just views details, but a survey gives you the option of rebooting the VM.

So far it only works in the staging environment, but the steps to enable it in production are on Tower, not in this repo.

### DIFF
--- a/playbooks/utils/vm_reboot.yml
+++ b/playbooks/utils/vm_reboot.yml
@@ -1,18 +1,10 @@
 ---
-- name: List all VMs in the dev vSphere
+- name: Get info on VM, with option to reboot
   hosts: localhost
-  # remote_user: pulsys
-  # become: true
 
   vars_files:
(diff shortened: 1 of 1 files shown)


playbooks/utils/vm_reboot.yml
```diff
@@ -1,47 +1,35 @@
 ---
```
maxkadel, line 1: On some playbooks we give examples on how to run it - would it be worthwhile to do that here, or is it a moot point since we'll be running it on Tower?

maxkadel: I think it's a moot point, ignore this.

maxkadel (approved):
Looks good, thanks @acozine!

PR #5203, opened 2024-07-26
Minimalist playbook to view info from vSphere

acozine:
This playbook returns information about:
- templates in vSphere
- hosts in vSphere
- datastores in vSphere
Eventually, this will support creating VMs from Tower.

For now, the playbook only gathers information. We can add more active tasks to it in a branch once this one is merged to main.

### DIFF
--- a/playbooks/utils/vSphere_audit.yml
+++ b/playbooks/utils/vSphere_audit.yml
@@ -1,0 +1,57 @@
+---
+- name: Get info from vSphere
+  hosts: localhost
+
+  vars_files:
+    - ../../group_vars/vsphere/vault.yml
+    - ../../group_vars/vsphere/{{ runtime_env | default('staging') }}.yml
+
+  tasks:
+    - name: Gather info on templates
+      community.vmware.vmware_vm_info:
+        hostname: "{{ vcenter_hostname }}"
+        username: "{{ vcenter_username }}"
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        show_datacenter: true
+        show_cluster: true
+        show_esxi_hostname: true
+        vm_type: template
+      register: template_info
+
+    - name: View template info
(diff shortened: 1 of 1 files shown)


playbooks/utils/vSphere_audit.yml
```diff
@@ -0,0 +1,57 @@
...
+        show_cluster: true
+        show_esxi_hostname: true
+        vm_type: template
+      register: template_info
+
+    - name: View template info
+      ansible.builtin.debug:
+        var: template_info
+
+    - name: Gather info on hosts
```
kayiwa, line 26: can we distinguish between ESXi host and host. I think we are asking for the ESXi host but not sure.

playbooks/utils/vSphere_audit.yml
```diff
@@ -0,0 +1,57 @@
...
+        password: "{{ vcenter_password }}"
+        validate_certs: false
+        show_datacenter: true
+        # schema: vsphere
+        # properties:
+        #   - hardware.memorySize
+        #   - something like numberOfVMs?
+      register: host_info
+
+    - name: View host info
```
kayiwa, line 39: Same here. I am unsure but can guess this is ESXi

kayiwa (approved):

playbooks/utils/vSphere_audit.yml
acozine: Good idea, thanks. Addressed in c1e12912b6ff1ba3f55d0bfb0f37a5e98ce8f3b5.

playbooks/utils/vSphere_audit.yml
acozine: Thanks again. Addressed in c1e12912b6ff1ba3f55d0bfb0f37a5e98ce8f3b5.

PR #5205, opened 2024-07-26
Vsphere audit

acozine:
Eventually this playbook should pull the parameters we care about from vSphere:

- how many VMs are running on each host?
- how much memory is free on each host?
- how much space is available on the various storage nodes?

Right now the playbook is not pulling the right values. Baby steps.

### DIFF
--- a/playbooks/utils/vSphere_audit.yml
+++ b/playbooks/utils/vSphere_audit.yml
@@ -52,7 +52,7 @@
 
     - name: View ESXi memory stats in GB
       ansible.builtin.debug:
-        msg: The {{ esxi_host_memory.ansible_facts.summary.config.name }} server is using {{ esxi_host_memory.ansible_facts.quickstats.overallMemoryUsage | map('int') | map('human_readable','unit','G') }}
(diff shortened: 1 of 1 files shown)


acozine:
As of 7cb0b7134e293fc50f49cd4a0c004fc6e739b111 the playbook succeeds and we get most of the data we want, see [this Tower run](https://ansible-tower.princeton.edu/#/jobs/playbook/4688/output). Clearly the 'quickStats.overallMemoryUsage` numbers are not the ones we want, but overall this is in decent shape.

acozine:
From the Ansible community - try converting MB to bytes and then using `human_readable` like this:
`int * 1024 * 1024 | human_readable`.

acozine:
Okay, a few math struggles later and this works well. There's a rounding-error difference between the memory usage stats in vSphere itself and the ones shown by the playbook, but it is certainly close enough. 

Do we want to order the storage info from most-available to least-available? Or is alphabetical by storage node okay?

Also, how do we add a reminder not to use the 15TB storage nodes?

acozine:
As of dd4de74ac19c15306134fcd0fabf0af860e03927 the playbook gets the data we need. Remaining work: present space available only for general purpose storage nodes, not for all storage nodes.

acozine:
Closing. We may be moving away from VMware anyway.

PR #5206, opened 2024-07-26
[orangelight] Add libanswers credentials as environment variables

sandbergja:
Helps with pulibrary/orangelight#3909

I ran this on staging today (26 July) and confirmed that the variables show up on the box.


maxkadel (approved):
Thanks @sandbergja! 🍄

Issue #5150, opened 2024-07-11
Remove library.princeton.edu proxy pass rule for repec role


kevinreiss:
- [x] Remove proxy pass rule [in staging](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-staging.conf#L115-L117)
- [x] Remove proxy pass rule [in production](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/templates/libwww-proxy-pass-prod.conf#L76-L78)
- [x] Redirect old staging proxy pass path to https://repec-staging.princeton.edu/
- [x] Redirect old production proxy pass path to https://repec-prod.princeton.edu/

kevinreiss:
Vendor has confirmed they are usiing repec-prod.princeton.edu to pull data and that this has been successful.

PR #5213, opened 2024-07-30
[libwww] Don't proxy pass library-staging to repec-staging

sandbergja:
Instead, do a 301 Moved Permanently redirect

Helps with #5150, I ran it on both load balancers today (30 July)


christinach (approved):
🍄

PR #5215, opened 2024-07-31
Builds a replacement VM with the same MAC address

acozine:
This PR is for testing, discussion and historical purposes only. I will squash the commits and open a clean PR once we're all happy with the functionality.

This branch gets our replace-a-VM playbook working with these steps:
- look at the existing VM and gather information about it
- move the existing VM to a folder called ToBeDeleted
- power off the existing VM so we can re-use the MAC address
- create a new VM from a template (we need to refine this task so the new VM gets the same resources as the old VM by default)
- add a NIC to the new VM with the same MAC address as the old VM (there's a bug in the collection that means we can't do this when we first create the new VM)
- power the new VM down
- delete the NIC that has the automatically-assigned MAC address
- power the new VM back up again
- remove the old VM
- alert slack

We use `uuid` or `moid` to select the correct VM for each task, so we can use a duplicate name.


acozine:
Known remaining work before the replacement VM will be action-ready:
- [ ] remove unnecessary parameters (if any) to make tasks easier to read
- [ ] remove, comment out, or consolidate debug tasks to streamline playbook
- [ ] find a solution for the Folder to move to-be-deleted VMs into - either create the `ToBeDeleted` folder in production as well as in staging, or add a variable for the folder to use and a matching Survey on the Tower template to request that information
- [ ] add Survey questions for the number of CPUs, amount of memory, and size of additional storage to add
- [ ] if possible, automate the [disk LVM expansion](https://github.com/pulibrary/pul-it-handbook/blob/main/services/disk_lvm_expansion.md), or consider mounting the extra disk at a known mount point and having build playbook install things there

acozine:
This playbook was succeeding in our staging environment; today it is failing.
I ran it in production, and the first error was `Failed to clone virtual machine acozine-test to folder 'vim.Folder:group-v9' due to permission issue: Permission to perform this operation was denied.` - fixed by updating the permissions in vSphere on the AnsibleControlledVMs folder. 
Then I went back to running it in the staging env, and the current error is `"Failed to create a virtual machine : Customization of the guest operating system is not supported due to the given reason: "` Which is not as informative as one could wish. It happens for both new templates - Rocky and the Ubuntu. Maybe we missed a step in creating the templates?

acozine:
Note for Future Us: as part of this work, we also needed to rebuild the templates on vSphere to expect the correct type of network device. With a mismatch, DHCP failed and the new VMs were unreachable via SSH. Thanks to @kayiwa for doing that!

carolyncole:
I have run this quite a few times on pdc-discovery staging and production machines.  It is definately working better than the main branch!

PR #5216, opened 2024-07-31
[Orangelight] update libanswers queue id

sandbergja:
We ran this on staging today, and confirmed that feedback went to the new queue.

We will remove the old queue after running this on production, because everybody keeps spamming the old queue.


christinach (approved):
Thanks @VickieKarasic and @sandbergja

PR #5219, opened 2024-08-05
[orangelight] send compressed assets if available and requested by the user's browser: qa and prod

sandbergja:
This extends #5218 to the production and qa environments.

The catalog application servers are already able to handle requests for gzipped assets, which is great, because that can make things much faster for users.

However, before this commit, the load balancer was causing us to lose this feature. Fortunately, telling the load balancer to use HTTP 1.1 (instead of the default 1.0) allows the content negotiation to still happen.

Helps with pulibrary/orangelight#4054

I ran this on the prod load balancer today (5 August, 2024)


christinach (approved):
Thanks @sandbergja

PR #5230, opened 2024-08-08
[NGINX] Allowing production ORCID to be accessed outside the VPN

carolyncole:
Also make inventory group for orcid

refs https://github.com/pulibrary/orcid_princeton/issues/181


leefaisonr (approved):
Thanks @carolyncole !

PR #5235, opened 2024-08-09
Add redirect for clip.php file.

kevinreiss:
I've learned that web content for course in canvas outside of ARES exists that points at the library.princeton.edu to view reserves content. Putting in a redirect just for the clip.php file as opposed to bring all the proxy rules back. Canvas folks tell me it's not practical for them to update URLs in this part of the system.


sandbergja (approved):
Thanks, @kevinreiss

christinach (approved):

PR #5236, opened 2024-08-09
Remove openbooks from load balancer configuration

sandbergja:
These VMs have been deleted, so nginx -t fails because it can't contact them.


kayiwa (approved):
OK! Thanks @sandbergja

PR #5238, opened 2024-08-09
Rewrite rule for videoreserves

maxkadel:
Tried on staging, ran on load balancers.


maxkadel (commented):
Changes look good! It worked!

kevinreiss (approved):

PR #5239, opened 2024-08-09
[ealapps] Serve ealapps via load balancer to non-princeton IPs

sandbergja:
Previously, users could access ealapps from outside Princeton via a library.princeton.edu proxy_pass rule.

With the website migration, this proxy_pass rule will no longer exist, so we need to make eal.lib.princeton.edu itself serve traffic from non-princeton IPs.

Thanks to Max for finding this.

I ran this on both prod load balancers today, and confirmed that I can access https://eal.lib.princeton.edu/EALJ/ without the VPN.


christinach (approved):
🦄 🍄

PR #5242, opened 2024-08-12
Add eal_production to inventory by environment

maxkadel:
- I cannot run the pulsys playbook on this server without adding it to the production inventory.


christinach (approved):
Thanks @maxkadel

PR #5245, opened 2024-08-12
remove openbooks

dphillips-39:


kayiwa (approved):
OK! thanks @dphillips-39

PR #5248, opened 2024-08-12
[DPUL] Install libidn-dev

tpendragon:
Necessary for upgrading to a new version of Spotlight.


kayiwa (approved):
OK! Thanks @tpendragon

PR #5251, opened 2024-08-14
[approvals] set SECRET_KEY_BASE envvar

sandbergja:
Previously, we were setting only an environment variable named APP_SECRET_KEY_BASE, which is used in the application but not by Rails in the `assets:precompile` step

This allows us to deploy approvals without getting an error `ArgumentError: Missing `secret_key_base` for 'production' environment`

@rladdusaw and I ran this on staging and it got the application deployable again.


maxkadel (approved):
thanks @sandbergja!

christinach (approved):
Thanks @sandbergja

PR #5254, opened 2024-08-14
[friends] update $base_url for new domain names

sandbergja:
Prior to this change, the friends sites displayed no CSS, since the CSS links included the old domain names.

I ran this on staging and prod.


christinach (approved):
Thanks @sandbergja

PR #5259, opened 2024-08-15
[mediaflux] add world reachable host

kayiwa:
modify the hostname of mflux-ci1 and create a NGINX config on the
loadbalancer


JaymeeH (approved):
Thanks @kayiwa !

PR #5261, opened 2024-08-15
Adds skeleton playbook for VEEAM backup query

acozine:
We currently manually maintain a spreadsheet of which VMs are backed up on which days. However, the actual backups are done based on tags in vSphere. So we should be able to pull this information from vSphere at any time.

As usual for automation with vSphere or the firewall, we need a minimal version of the playbook merged to main before we can test more functionality in Ansible Tower.


sandbergja (approved):
THanks, @acozine !

PR #5265, opened 2024-08-16
[CHECK_MK] update the local and role checks to use the new shorter ansible host names

carolyncole:
also added the checkmk health url for ORCID

refs https://github.com/pulibrary/orcid_princeton/issues/182


acozine:
Looks good, thanks @carolyncole!

PR #5270, opened 2024-08-19
[libwww] block external access

### LINKED ISSUE #5255: migrated drupal sites should not be accessible from outside the VPN
### Expected behavior

The following sites are not accessible from outside the VPN:
* fpul-migrated.lib.princeton.edu
* library-migrated.lib.princeton.edu
* library-migrated.lib.princeton.edu/special-collections

### Actual behavior

I can still access them.

### Steps to replicate

1. Leave the VPN
2. Go to one of the above URLs in your browser
3. Note that you can access the sites, rather than getting the expected "Something went wrong"

### Impact of this bug

Low: we don't want people off the VPN to access these, since it will be confusing and potentially out-of-date.

kayiwa:
currently this website is accessible outside of the princeton network.

closes #5255

### DIFF
--- a/roles/nginxplus/files/conf/http/fpul-prod.conf
+++ b/roles/nginxplus/files/conf/http/fpul-prod.conf
@@ -38,6 +38,10 @@ server {
         health_check interval=10 fails=3 passes=2;
         # handle errors using errors.conf
         proxy_intercept_errors on;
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @kayiwa

PR #5271, opened 2024-08-19
Reduce log level back to default for psql.

### LINKED ISSUE #5257: Reduce the log level on lib-postgres-prod1
We set the log level to `all` when pdc-describe was launched. We can reduce the size of the logs by reverting the log level to a more modest setting.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Relevant deadlines
N/A

Logs on lib-postgres-prod1 were filling up the disk. This was partly due to a flaw in our log rotation, but also due to the chattiness of the logs.

### Acceptance criteria
 - [ ] postgres logs grow more slowly
 - [ ] we still see errors but we do not log every transaction

[Docs](https://postgresqlco.nf/doc/en/param/log_statement/) on `log_statement`.

tpendragon:
Closes #5257


acozine:
Adding the words postgres and postgresql in a comment so this PR will show up when searching GitHub for those terms.

PR #5277, opened 2024-08-20
Refines playbook to view vSphere data on Veeam backups

acozine:
This playbook returns a summary of the tags we have in vSphere, how many VMs are being backed up for each tag, and the total disk size being backed up for each tag.

May need to be updated for use in production, where we have at least one tag we may want to ignore.

This playbook contains some good data-wrangling task examples.

### DIFF
--- a/playbooks/utils/veeam_backup_list.yml
+++ b/playbooks/utils/veeam_backup_list.yml
@@ -7,23 +7,40 @@
     - ../../group_vars/vsphere/{{ runtime_env | default('staging') }}.yml
 
   tasks:
-    - name: Gather info on tags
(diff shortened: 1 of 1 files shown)


acozine:
Huge thanks to @sandbergja who found https://play.jmespath.org/ for testing JSON queries and who suggested both queries and approaches that worked!

acozine:
See also [this gist](https://gist.github.com/acozine/5e1d14a8971213c6f9da7f64680c3843) for the last change that fixed the playbook. It has some good examples of queries we ended up not using, but that might be helpful for Future Us.

sandbergja (approved):
Thanks, @acozine

PR #5285, opened 2024-08-22
Updates the "new kicker" playbook

acozine:
This playbook can now be run in Tower in staging or in production.

This PR clarifies a task name (you may or may not be rebooting the VM) and adds a reminder about the format of our vCenter host names.


sandbergja (approved):
Thanks, @acozine !

PR #5290, opened 2024-08-23
adds template and task to override the default systemd logrotate timer

### LINKED ISSUE #5256: Systemd should check frequently whether log rotation needs to happen
### Expected behavior
When we configure log rotation, the system should rotate logs when they reach the size or age we define.

### Actual behavior
When using systemd for log rotation, a timer defines how often the system checks the log files to see if log rotation is needed. If this timer is set to a long interval, then logs can reach a much larger size without rotating.

### Steps to replicate
Set postgresql logging to `all`, set log rotation to happen when the log file reaches 1GB, and watch the disk fill up.
SSH to the machine and run `sudo systemctl list-timers` to see how long the machine will wait before even checking to see if maybe log rotation would be a good idea.

### Impact of this bug
Log rotation does not happen as expected, disks fill up, systems go down.

Incident [notes](https://docs.google.com/document/d/1lSDs0JxeGLzP8dW8z9WTv-BLEfDJeP_YsVDb1qf_ty4/edit)

Confirm that we are using systemd for logrotate across our fleet of VMs. Some VMs may be using cron instead.
Create a template for `/etc/systemd/system/logrotate.timer.d/override.conf`.
The next task after creating or updating `override.conf` should run `sudo systemctl daemon-reload` to pick up the new timer settings.
The template and tasks should go into the `common` role.

acozine:
Closes #5256.

Adds an override to all our servers for the systemd logrotate timer. With the default settings in this template, systemd will check the logs to see if they are large enough to warrant rotation (as defined by files in `/etc/logrotate.d`):
- 20 seconds after a service is restarted
- every 10 minutes thereafter

With this setting, even with large amounts of data going to a log, we should retain enough disk space for the service to continue uninterrupted.

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -1,2 +1,5 @@
 ---
 # handlers file for roles/common
+- name: reload logrotate timer settings
+  ansible.builtin.command: systemctl daemon-reload
+
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -68,10 +68,8 @@
     dest: "/etc/systemd/system/logrotate.timer.d/override.conf"
     mode: 0644
   when: ansible_service_mgr=="systemd"
-
-- name: common | reload to pick up logrotate timer settings
-  ansible.builtin.command: systemctl daemon-reload
-  when: ansible_service_mgr=="systemd"
+  notify:
+    - reload logrotate timer settings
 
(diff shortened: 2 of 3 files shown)


acozine:
Tests for the apache role fail with `Destination directory /etc/systemd/system/logrotate.timer.d does not exist`.

carolyncole (approved):
Thanks @acozine

PR #5291, opened 2024-08-23
Preload bundler in passenger by default.

tpendragon:
Nearly every site has this, and the ones that don't have this break as soon as we update to 7.1.


tpendragon:
I'm closing this. Deploying it to imagecat made me update rubygems, so this would break a bunch of deploys.

PR #5297, opened 2024-08-23
Add new archives for SC and Cams.

kevinreiss:
Adding two new repos for the SC libserv55 archive and CaMS docs archive.


christinach (approved):
Thanks @kevinreiss

PR #5300, opened 2024-08-27
Swap cdh web production and pre-production servers

rlskoeser:
This change makes the new version of the CDH website live and keeps the old version running at cdh-beta in case we need to reference the old site during the transition period.


acozine (approved):
Looks fine to me.

Issue #5299, opened 2024-08-26
Request new staging VMs for CDH HTR2HPC project


rlskoeser:
### Describe the VM(s) you need

Two staging VMs for testing out deploy and installation of the eScriptorium application.

If you do not know the answers to some of these questions, that's okay. Please add the information you do have, and we will work together to gather any other information we need. You can also look at the [documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/new_vms_for_devs.md) in pul-it-handbook.

#### Which project are these VMs for?
- [x] Project:  CDH HTR2HPC - testing out eScriptorium software 

#### How many VMs do you need?
- [x] Number of VMs: 2

#### What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [x] FQDN:  test-htr.princeton.edu (🤔❓ )

This is a temporary installation to test out the deploy and determine what would be needed for this application to be supported at Princeton in an ongoing basis. As such, I think it's simpler to leave the `.cdh.` out of the hostname in this case (since you can't use the automated form to request them).

#### What virtual resources does each VM need?
- [x] Number of CPUs per VM: (2 CPUS)
- [x] Amount of memory per VM (in 16 GB):
- [x] Amount of storage/disk space per VM (in 30 GB):

Not sure about this but inclined to start small. Our goal is to split out the web application from the high performance computing needs, and this instance will only be used by a small set of beta testers.  Whatever you typically deploy for CDH web applications should be plenty to start with.

#### What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [x] allow access via ssh port 22 from princeton VPN 
- [x] port 80 accessible to load balancer
- [x] access to postgres server
- [x] access to redis 
- [ ] access to mail server ? (need to be able to send mail)

I think the standard setup should work for now; are any firewall changes needed for internal library services? We will want to configure the application to access postgres, redis, and be able to send email.  We will need to figure out how to connect the task runner(s) for this application to Research Computing HPC cluster, but we will follow up when we have specifics about that.

#### What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [x] NFS

We can use the existing CDH section of NFS for this pilot phase. Please add the new VMs to the list of machines that should have access to the CDH staging mount point in PUL NFS.

The application needs to serve out and work on the images to be transcribed, and they will need to be shared by the two VMs.

### Relevant deadlines
[...]

PR #5310, opened 2024-09-05
[cdh] two new VMs

kayiwa:
add two new vms for HTR
add them to the nfs exports

related to #5299


rlskoeser (approved):

roles/nginxplus/files/conf/http/dev/cdh_test_htr.conf
```diff
@@ -0,0 +1,58 @@
...
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_cache cdh-test-htrcache;
+        # handle errors using errors.conf
+        proxy_intercept_errors on;
+        health_check interval=10 fails=3 passes=2;
+        # allow princeton network
+        # include /etc/nginx/conf.d/templates/restrict.conf;
+        # block all
+        # deny all;
```
rlskoeser, line 54: @kayiwa should these be commented out?

kayiwa: by default anything on the dev loadbalancer is blocked. So it would be redundant.

PR #5316, opened 2024-09-09
Polishing up the build-a-vm playbook

acozine:
The playbook now runs from Tower in either production or staging (there is a Tower Template for each environment for now, because the network names are not consistent across environments). It shows VM disk usage correctly, and it puts the replacement VM into the same folder in vSphere that the old VM was in.

I've removed the tags for now. They were useful for testing, but I think they might be tricky to manage in the long term. We could reinstate them and have multiple Tower Templates build off the same playbook with different tags . . . but let's talk about the use cases. 

When this gets merged, we should update the Tower templates so they no longer ask for a source control branch name.

### DIFF
--- a/playbooks/utils/replace_vm_host.yml
+++ b/playbooks/utils/replace_vm_host.yml
@@ -2,16 +2,19 @@
 #  This playbook will delete a VM and replace it. This is the only warning you will get. Running this playbook has destructive consequences
 #  This playbook should only be run from ansible tower.  No need for detailed instructions of examples here, because you should not run it here.
 #
-- name: replace the {{ replacement_vm }} VM on vSphere
+- name: replace the {{ replacement_vm }} VM on the {{ runtime_env | default('staging') }} vSphere
   hosts: localhost
   remote_user: pulsys
   become: true
+
   vars:
     slack_alerts_channel: "#infrastructure"
+
   vars_files:
     - ../../group_vars/all/vault.yml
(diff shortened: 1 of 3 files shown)


playbooks/utils/replace_vm_host.yml
```diff
@@ -25,26 +28,18 @@
         # show_cluster: true
         vm_name: "{{ replacement_vm }}"
       register: old_vm_info
-      tags:
```
carolyncole, line 28: @acozine I would prefer to leave a way to get the information about a host in the playbook.  Removing it means if I want to double check if the machine looks the same before and after the replacement I have to have access to vsphere.  Maybe the tag could be something like sys_info instead of disk

acozine: My concern about tags is really about the Tower way of managing them. There's no good way to see which tags are available from the Tower side. You can add them to a template, but then the user has to delete them to get the full playbook to run. But maybe we can meet this use case with multiple Templates for the same playbook. I will investigate.

acozine: We will add more diagnostics. Use case: we need to confirm that the new machine has the same CPU, memory, and amount of storage space as the old machine, for confidence that the replacement has been done correctly.

acozine: Can we also hook this up to the pulsys user playbook?

carolyncole (approved):
Thank you @acozine

PR #5323, opened 2024-09-10
Give more information about VMs in playbook output

acozine:
Now that we know how to gather data from vSphere about CPU, memory, disk space, and more, we can enhance the view provided by our view-details-and-reboot-if-needed playbook. 

The task now returns the host (with note about the location), MAC address, CPU, memory, storage amount, and storage location.


kayiwa (approved):
OK! Thanks @acozine

PR #5331, opened 2024-09-12
Update the veeam backups to show a list of VMs with no tags

acozine:
We add tags to VMs we want to back up with VEEAM. Each tag reflects the location and OS of the VM plus which day of the week it gets backed up. Best practice is to add a tag for "don't back this VM up" to any VMs we do not want to back up with VEEAM. This way we know for sure that the lack of a backup is intentional, and not an oversight.

This PR expands the VEEAM backup playbook to report a list of VMs that have no tags, so we can audit vSphere, make decisions about which VMs to back up, and keep the tags up to date.

### DIFF
--- a/playbooks/utils/veeam_backup_list.yml
+++ b/playbooks/utils/veeam_backup_list.yml
@@ -32,9 +32,7 @@
 
     - name: Find VMs that do not have tags
       ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK! Thanks @acozine

PR #5336, opened 2024-09-12
[Imagecat] Run on Nomad

tpendragon:
Run ImageCat on Nomad.


roles/pul_nomad/files/pul_nomad_policy.hcl
```diff
@@ -3,9 +3,12 @@ agent_prefix "" {
 }
 
 node_prefix "" {
-  policy = "read"
```
hackartisan, line 6: what about imagecat necessitated this change? would we have needed it eventually for dpul-c but we just didn't bump up against it yet?

tpendragon: Yeah. The nomad logs had some errors in it I only noticed because I was in there poking around. It could be a separate PR.

hackartisan: A separate PR would be nice

hackartisan (approved):

PR #5338, opened 2024-09-13
Update inventory entries for Nomad VMs (dpul-c)

acozine:
Adds the staging nomad client group to the larger staging group.
Updates the nomad prod group name to match our usual pattern.
Updates the prod environment entry to match.


tpendragon (changes requested):
https://github.com/pulibrary/princeton_ansible/blob/imagecat_staging/group_vars/nomad_cluster.yml#L2

Whatever that group is (in this case `nomad_cluster`) needs to have the whole cluster in it.

We can name the whole thing `_production` if we want. The "staging" clients are still hooked to the prod cluster, just hosted on staging VMs and running staging containers, if that makes sense?

tpendragon (approved):

PR #5352, opened 2024-09-17
Rename bibdata staging

christinach:
Rename bibdata-staging.princeton.edu to bibdata-staging.lib.princeton.edu
Bibdata staging is on adc-dev and using the .lib


kayiwa (approved):
OK! Thanks @christinach

PR #5353, opened 2024-09-17
Update cdh_test_htr config : add configurations for proxying websockets

rlskoeser:
This application includes an ASGI component, which is proxied by the nginx on our cdh-test-htr VMs. I think these are the configuration changes that are needed to allow nginx to proxy the websockets / ASGI.


kayiwa (approved):
OK! Thanks @rlskoeser

PR #5356, opened 2024-09-18
[Bibdata-staging][Passenger] Set bibdata-staging.lib to use dev load balancer;templatize dev and prod load balancer IPs

christinach:
[Bibdata-staging]Set bibdata-staging.lib to use dev load balancer by providing real IP variable in group_vars/bibdata/staging.yml

[Passenger]templatize dev and prod load balancer IPs make default to be prod load balance IPs. see: roles/passenger/defaults/main.yml Update roles/passenger/templates/passenger.j2


kayiwa (approved):
OK! Thanks @christinach

Issue #5354, opened 2024-09-17
New PAS Staging Environment


kevinreiss:
### Describe the VM(s) you need
A vendor is upgrading the Princeton and Slavery project to a new major version of CraftCMS. Going from the current application on version 4 to version 5.3.x. These VMs will be the "new" staging environment for the version 5.3.x of this project. We'll eventually provision a new prod environment. 

Which project are these VMs for?
- [x] Project: Princeton and Slavery https://slavery.princeton.edu/

How many VMs do you need?
- [x] Number of VMs:2 

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [x] FQDN: slavery-staging-upgrade1.princeton.edu and slavery-staging-upgrade2.princeton.edu. 

What virtual resources does each VM need? - Specs form existing prod box
- [x] Number of CPUs per VM: 2
- [x] Amount of memory per VM (in GB): 8
- [x] Amount of storage/disk space per VM (in GB): 28

What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [x] Port number(s) 22, 443, 80 redirected to 443
- [x] Service(s) MariaDB, libpony-express. 

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [x] TLS Certificates
- [x] nginxplus configs
- [x] Additional resources: Isilon mount to the existing pas mount at /mnt/diglibdata1/pas.  

### Relevant deadlines
If the new VMs must be ready by a certain date, please enter it here.
- [x] We need these VMs by: 09/23/2024

### Implementation notes, if any
The minimum dependencies for this CraftCMS 5.3.x are listed below. We'll need to complete this ticket to upgrade the pas role to use a higher version of PHP to successfully deploy the project: https://github.com/PrincetonUniversityLibrary/pas-craft3/issues/115. 

    PHP 8.2+
    MySQL 8.0.17+ using InnoDB, MariaDB 10.4.6+
    256MB+ memory allocated to PHP
    200MB+ free disk space
    Composer 2.0+

Craft also supports postgres but moving to postgres is beyond the scope of this upgrade which is being done by a vendor.

VickieKarasic:
Kevin confirmed that we can put these VMs on the private network, so we will change the FQDNs to: 

- [ ] slavery-staging-upgrade1.lib.princeton.edu
- [ ] slavery-staging-upgrade2.lib.princeton.edu

dphillips-39:
[...]

PR #5358, opened 2024-09-18
[PAS] Adding new VM's to new inventory

dphillips-39:
related to #5354 

### DIFF
--- a/inventory/all_projects/pas
+++ b/inventory/all_projects/pas
@@ -2,4 +2,6 @@
 slavery-prod1.princeton.edu
 [pas_staging]
 slavery-staging1.princeton.edu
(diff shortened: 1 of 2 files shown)


inventory/all_projects/pas
```diff
@@ -3,3 +3,6 @@ slavery-prod1.princeton.edu
 [pas_staging]
 slavery-staging1.princeton.edu
 # slavery-staging2.princeton.edu
+[pas_staging_upgrade]
```
acozine, line 6: We also need to add this new group to the `staging: children` list so that these VMs get the OS updates and other changes that run against the `staging` group.

VickieKarasic: Good catch - thanks! Just made that change and committed.

PR #5363, opened 2024-09-19
Deleted lib-monitor1 from nginx config - i5118

dphillips-39:
We decommissioned lib-monitor1 so that loadbalancer does not refer to lib-monitor1


kayiwa (changes requested):
Heya. We will need to remove it from the inventory.

https://github.com/pulibrary/princeton_ansible/blob/dc7f24a442e66167c9b7c46110b3fdc924ba8755/inventory/all_projects/_orphans#L8

dphillips-39:
replace with https://github.com/pulibrary/princeton_ansible/pull/5381

Issue #5294, opened 2024-08-23
Move Libweb2 Content to libstatic role


kevinreiss:
### What maintenance needs to be done?
Libserv55 (where the alias libweb2.princeton.edu currently lives) is an EOL out of maintenance ubuntu server. 

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- A clear description of why this is needed.

### Acceptance criteria
 - [x] Directories from libweb2 marked for "migrate" are available through the lib-static role see https://github.com/pulibrary/princeton_ansible/pull/5297
 - [x] Obtain report of any links in finding aids source data containing libweb2.princeton.edu. See [report request](https://github.com/pulibrary/aspace_helpers/issues/584).
 - [x] Obtain report of any links in the catalog pointing at libweb2.princeton.edu. See requests for [Portfolios](https://github.com/PrincetonUniversityLibrary/alma-config/issues/1290) and [856](https://github.com/PrincetonUniversityLibrary/alma-config/issues/1289) links. [Portfolio report](https://docs.google.com/spreadsheets/d/1zEliBEfqj32lrq69upE_G4Ges6rFoCT0/edit?gid=366144202#gid=366144202) and [856 report](https://docs.google.com/spreadsheets/d/1EzQak2e9wYrmFttOwqRimxR5mUZZrXd1/edit?gid=238523515#gid=238523515). 
 - [x] create ticket for Alma team to update Catalog links to point to new locations on lib-static - [Alma Config Ticket](https://github.com/PrincetonUniversityLibrary/alma-config/issues/1339). 
 - [x] Move libweb2 alias to lib-adc
 - [x] Create a redirect for the libweb2 top level path http://libweb2.princeton.edu/rbsc2/ to new static-prod path at https://static-prod.lib.princeton.edu/scsites/. 
 - [x] Power off libserv55

### Implementation notes, if any
SC old web [content review](https://docs.google.com/spreadsheets/d/1UsMmbMxyJVr03I6NHJ1449_35_OkJW3SNhdaq0nzCaI/edit?gid=1351235624#gid=1351235624).

kevinreiss:
Requested reports for points 2 and 3 today.

kevinreiss:
@acozine reached out about finishing this ticket and we now can move on the steps to add libweb2.princeton.edu to lib-adc, princeton_ansible, and add the necessary redirects.

acozine:
[...]

PR #5366, opened 2024-09-20
Migrating libweb2 to the static site

acozine:
Partial fix for #5294. 

Adds a redirect for the old libweb2 URL.
Adds nginxplus config for the new site within libstatic.

Once this is confirmed as working, we can decommission the old infrastructure.


kevinreiss (approved):

acozine:
Also alphabetizes the entries in this config file.

PR #5369, opened 2024-09-20
[Nomad] Update Consul & Fix Permissions

tpendragon:
The nomad permissions template wasn't the recommended template from https://developer.hashicorp.com/nomad/docs/integrations/consul/acl. This fixes that, and also updates Consul.

This also moves secrets for Nomad to its own vault file.


hackartisan (changes requested):
Will the same policy be used on nomad servers and clients? maybe we should have two separate files?

tpendragon:
> Will the same policy be used on nomad servers and clients? maybe we should have two separate files?

Sure, that's a good idea. It's gonna expand this PR a fair bit.

tpendragon:
> Will the same policy be used on nomad servers and clients? maybe we should have two separate files?

Done.

eliotjordan (approved):

PR #5371, opened 2024-09-20
[DACS#236] Install wapiti in localhost

christinach:
closes https://github.com/pulibrary/dacs_handbook/issues/236

Scan and create an html report


acozine:
We created a new execution environment to run the Wapiti reports from Ansible Tower - see #5387.

PR #5376, opened 2024-09-24
[cdh-prodigy-test] Update nginx config to allow large headers

laurejt:
Updating the nginx config to allow large headers for the cdh prodigy test server.

---
Co-authored-by: Rebecca Sutton Koeser <rlskoeser@users.noreply.github.com>


kayiwa (approved):
OK. Thanks @laurejt

PR #5377, opened 2024-09-24
[postfix] enable port 587

kayiwa:
this modification allows postfix to enable port 587 locally

make smarttls work


christinach (approved):
Thansk @kayiwa

PR #5387, opened 2024-09-26
adds a new EE with python3.11 for Wapiti reporting

acozine:
Wapiti requires python 3.10 or 3.11. This EE provides python 3.11.


christinach (approved):
Thanks @acozine and @christinach

PR #5389, opened 2024-09-27
Updates checkMK for move to production

acozine:
Updates the edition of CheckMK we're installing in production to the Enterprise edition.
Updates the version of CheckMK we're installing in production to 2.3.0p15 to match the OOBM system.
Pulls shared values into a 'shared.yml' file in 'group_vars/checkmk' to make it easier to update variables and easier to see how the prod and staging sites differ.


VickieKarasic (approved):
Thanks @kayiwa - this looks good

PR #5392, opened 2024-09-30
Use health check endpoint for staging

### LINKED ISSUE #5112: [DSS] Use health.json for health checks

When DSS's health checks are failing (health.json), it should not serve that application to the public.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria
 - [ ] It checks `/catalog/health.json?providers[]=database&providers[]=solrstatus`

Existing health check configuration: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/dss-staging.conf

### First Step

rladdusaw:
Resolves #5112


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #5394, opened 2024-09-30
[videoreserves] More cleanup

sandbergja:
* We no longer need the vault, now that we have no db credentials
* We no longer need to run the mysql role as part of molecule converge

Related to https://github.com/PrincetonUniversityLibrary/video_reserves/issues/81


christinach (approved):
Thanks @sandbergja

PR #5402, opened 2024-10-02
[orangelight] Add X-Forwarded-Proto https to the catalog load balancer config

sandbergja:
Without this, login and forms were failing in the catalog.  In the logs we saw: HTTP Origin header (https://catalog-staging.princeton.edu) didn't match request.base_url (http://catalog-staging.princeton.edu)


christinach (approved):
Thanks everyone!

PR #5409, opened 2024-10-02
[orangelight-staging] Update the mailcatcher installation path in catalog-staging

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #5413, opened 2024-10-03
Used better variables to loop through RedHat and debian installs

Beck-Davis:


acozine (approved):
Looks good, thanks @Beck-Davis

Issue #5401, opened 2024-10-02
New VM: mflux-qa.lib.princeton.edu


bess:
### Describe the VM(s) you need

Which project are these VMs for?
- [ ] Project: Tiger Data

How many VMs do you need?
- [ ] Number of VMs: 1

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [ ] FQDN: mflux-qa1.lib.princeton.edu

What virtual resources does each VM need?
- [ ] Number of CPUs per VM: 2
- [ ] Amount of memory per VM (in GB): 8
- [ ] Amount of storage/disk space per VM (in GB): 25

What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [ ] Port number(s) 80, 22, 443
- [ ] Service(s) http, https, sshd

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [ ] Additional resources: no

### Relevant deadlines
If the new VMs must be ready by a certain date, please enter it here.
- [ ] No hard deadline

### Implementation notes, if any
This should use the rocky linux server template

PR #5415, opened 2024-10-03
Add mflux-qa1 to inventory

Beck-Davis:
#5401


VickieKarasic (approved):
Looks good - thanks @Beck-Davis

PR #5416, opened 2024-10-03
move abid-staging to the dev LBs

acozine:
As part of our migration to the dev load balancers, this PR:

- changes the server definitions for abid-staging (the changes to `abid-staging2` - from a routable to a non-routabel IP, and from .princeton.edu to .lib.princeton.edu - are already done)
- moves the site config into the dev subfolder so it gets deployed to the dev LBs


kayiwa (commented):
thanks @acozine

PR #5424, opened 2024-10-06
[Nomad] Update DNS to allow podman containers to use Consul DNS.

tpendragon:
DNSMasq on the hosts has to listen to everything but the LAN interface, so it can resolve queries coming from the podman interface.

### DIFF
--- a/roles/pul_nomad/tasks/main.yml
+++ b/roles/pul_nomad/tasks/main.yml
@@ -215,6 +215,47 @@
     - "consul_node_role == 'client'"
     - created_agent_token_client
 
+- name: 'pul_nomad | Configure DNS for consul'
+  become: true
+  ansible.builtin.lineinfile:
+    path: /etc/resolv.conf
+    regexp: ^nameserver 127.0.0.1
+    line: nameserver 127.0.0.1
+    create: true
+    owner: root
+    group: root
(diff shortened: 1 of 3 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

roles/pul_nomad/tasks/main.yml
```diff
@@ -215,6 +215,48 @@
...
+    group: "root"
+    mode: "0644"
+  become: true
+  notify: restart dnsmasq
+  when:
+    - "ansible_os_family == 'RedHat'"
+  tags: dnsmasq
+
+# Allows DNSMasq to listen on the podman interface, so containers can query
+# consul. ENS160 is the WLAN interface - listen on everything else.
```
hackartisan, line 246: why do we listen on "everything else"? rather than a specific thing? what is included in "everything else" and how do we know that ENS160 is the WLAN interface?

tpendragon: We listen on "everything else" because the podman interface doesn't exist until nomad starts a pod, so if I tell it to just listen to localhost and podman it errors and crashes.

We know ENS160 is the WLAN interface because that's the interface when I looked at them, and we build them from a common image.

Issue #5118, opened 2024-07-05
Decommission Lib-monitor1


dphillips-39:
### What maintenance needs to be done?
We are not using Lib-monitor1 any more. Let's decommission it.

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### Relevant deadlines
If this maintenance relates to an End of Life date or other deadline, please enter it here.
- [x] None

### Why is this maintenance needed?
- We would like to make efficient use of our resources and we do not want to maintain machines that are no longer in use. 

### Acceptance criteria
 
- [x]  delete nginx lib-monitor1 [configuration](https://github.com/pulibrary/princeton_ansible/blob/b6e4b9c032fba8c32638857860ec9acacd744537/roles/nginxplus/files/conf/http/lib-monitor.conf#L6)
- [x]  delete aliases on lib-adc
- [x]  delete lib-monitor1 vm
- [x]  delete lib-monitor1 from [inventory](https://github.com/pulibrary/princeton_ansible/blob/b6e4b9c032fba8c32638857860ec9acacd744537/inventory/all_projects/_orphans#L8)
- [x]  delete DNS for the VM
- [x]  delete firewall rules

acozine:
We disconnected the lib-monitor1 VM from the network on July 5, 2024.

PR #5427, opened 2024-10-07
Removed lib-monitor

dphillips-39:
Removed Lib-monitor from inventory and lib-monitor.conf

Related to #5118


inventory/all_projects/_orphans
```diff
@@ -5,7 +5,6 @@ issuetracker-prod1.princeton.edu
 lib-adc1.princeton.edu
 lib-adc2.princeton.edu
 [monitor_backends]
```
acozine, line 7: We should probably remove the group as well:
```suggestion
```

Beck-Davis: Okay, we'll remove it and do another commit on this branch when we come back to this ticket tomorrow. Thanks @acozine, I wasn't sure if we would need this group later.

dphillips-39: This group was removed

acozine (approved):
Looks good, thanks @dphillips-39

PR #5429, opened 2024-10-07
[orangelight] Remove catalog-qa1 from the load balancer

sandbergja:
Something is broken with passenger on catalog-qa1, and the health checks are filling up the disk with errors quickly


VickieKarasic (approved):
Thanks for addressing this, @sandbergja!

PR #5432, opened 2024-10-09
[gitlab] add mailing ability

kayiwa:
disable TLS on the gitlab server (uses TLS by default)
add the server to the mail relay allow list


carolyncole (approved):
Looks go to me @kayiwa.  Thanks!

PR #5441, opened 2024-10-10
[ufw_firewall] add VM firewall rules

### LINKED ISSUE #4965: automate local firewall rules for vms in the .lib domain

Create firewall rules for all our virtual machines. We do not use PanOS on our [*.lib.* domain](https://datatracker.ietf.org/doc/html/rfc1918) and therefor will need to add the same rules on the host firewall. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

- While the non-routable ip addresses are only reachable to us having firewall rules is probably a good idea across all our infrastructure

### Acceptance criteria
 - [ ] A UFW role
 - [ ] Variables for the ports to open

kayiwa:
- **a parametrized role for a hosts firewall**
- **add documentation for the role**
- **our lowest version is no longer bionic**
- **add ufw_firewall to roles to be tested**

closes #4965


roles/firewall/defaults/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+# defaults file for roles/ufw_firewall
+firewall_rules:
```
kayiwa, line 3: add ufw as variable to avoid confusing it with panos

roles/firewall/handlers/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# handlers file for roles/ufw_firewall
+- name: restart ufw
+  ansible.builtin.service:
+    name: ufw
+    state: restarted
+
+- name: reload firewalld
```
kayiwa, line 8: we won't be using firewalld

roles/firewall/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+# tasks file for roles/ufw_firewall
+- name: Firewall | Determine OS family
```
kayiwa, line 3: we are using ufw

carolyncole (approved):
We reviewed in Open house on 10/28

Issue #969, opened 2019-11-04
lib-static-staging1 & lib-static-prod1 Hardware


carolyncole:
We are currently hosting a number of static html content on our various servers like libphp-prod.

We would like to create a vm to move this content to.

It would need certificates for https access from behind the load balancer.

kayiwa:
BEST TICKET EVER!

kayiwa:
VMs are ready.

PR #5443, opened 2024-10-17
upgrading the mediaflux version and pushing a tagged release to docker hub

JaymeeH:
refs [#969](https://app.zenhub.com/workspaces/rdss-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/969)

### DIFF
--- a/playbooks/mflux-docker.yml
+++ b/playbooks/mflux-docker.yml
@@ -5,10 +5,10 @@
 # to run this playbook, use the following command: ansible-playbook mflux-docker.yml
 # pass -e docker_tag={tag_number} to specify the tag for the docker image
 # pass -t publish_image to push the image to docker hub
-# to run the playbook for local development, pass -e platform_base=arm64 -e docker_base_image=rockylinux:9 -e docker_tag=dev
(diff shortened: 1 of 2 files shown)


playbooks/mflux-docker.yml
```diff
@@ -8,7 +8,7 @@
 # to run the playbook for local development, pass -e platform_base=arm64 -e docker_base_image=rockylinux:9 -e docker_tag=dev
 
 # Create a docker container from the local version of the image
-# docker create --name mediaflux-dev --publish 8888:80 --mac-address 02:42:ac:11:00:02 mediaflux-development
+# docker create --name mediaflux --publish 8888:80 --mac-address 02:42:ac:11:00:02 mediaflux-development:{tag_number}
```
carolyncole, line 11: I see this change, but you did utilize the `docker_tag` on the command line correct?

carolyncole (changes requested):
Just making sure you utilized the automatic upload with `docker_tag` if so I think we could either delete the line you modified or reference docker_tag in that line

playbooks/mflux-docker.yml
JaymeeH: Good catch. I passed the `tag_number` as the docker tag, so i'll just update the line

PR #5447, opened 2024-10-18
Preserves the network when replacing a VM

### LINKED ISSUE #5423: Replace-a-vm playbook should keep the new VM in the same network
### Expected behavior
If I replace an existing VM, the new VM should come up in the same network space as the old VM.

### Actual behavior
You must choose which network you want to put your replacement VM on, and the output about the old VM does not include any information about the network or the IP.

### Impact of this bug
If we replace a VM and the new VM comes up in the wrong network space, it will not get an IP address.

### Acceptance criteria
- [x] The `old_vm_info` variable should, if possible, show which network the old VM was on.
- [x] The default should be `keep the new VM in the same network space as the old VM.`
- [x] The Survey should make it clear that it's possible to override the previous setting. We do want to keep this functionality for the network migration, but we should make the template safe to run for VMs that are not being migrated.

acozine:
Closes #5423.

The basic requirements are met in this branch - see https://ansible-tower.princeton.edu/#/jobs/playbook/7184/output for a run in Tower.  We updated the Tower template survey about which network to put the new VM on - it's not required and there's no default, so if you do nothing, your new VM comes up in the same network as your old VM. 

We could make this even better by updating the `Print out relevant details of old VM` and `Print out relevant details of new VM` tasks to include a line about which network they were/are on.

### DIFF
--- a/playbooks/utils/replace_vm_host.yml
+++ b/playbooks/utils/replace_vm_host.yml
@@ -40,175 +40,175 @@
       ansible.builtin.debug:
         msg: Ansible will now move and power off the current {{ old_vm_info.virtual_machines[0].guest_name }} VM, then create a replacement.
 
-    - name: Move old VM to the ToBeDeleted folder
-      community.vmware.vmware_guest_move:
-        hostname: "{{ vcenter_hostname }}"
-        username: "{{ vcenter_username }}"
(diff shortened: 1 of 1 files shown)


carolyncole (approved):
Thank you!

PR #5450, opened 2024-10-21
[PAS] let the load balancer know that the app server is on the .lib network

sandbergja:
I ran this on both load balancer machines.  Before this PR, trying to go to https://slavery-staging.princeton.edu/ just gave the "Something went wrong"


christinach (approved):
thanks @sandbergja

PR #5453, opened 2024-10-22
[recap] Set prod php_version to 8.3

sandbergja:
Previously, the playbook was installing php 8.3 but php-8.1-gd, so when composer tried to check for the gd extension from PHP 8.3, it failed

Closes https://github.com/pulibrary/recap/issues/269


christinach (approved):
Thanks ! 🍄

PR #5454, opened 2024-10-22
[researchdata] Set prod php_version to 8.3

sandbergja:
Previously, the playbook was installing php 8.3 but php-8.1-gd, so when drupal tried to check for the gd extension from PHP 8.3, it failed.  This blocked deploying to prod.

This extends #5453 to researchdata


christinach (approved):
Thanks @sandbergja !

PR #5461, opened 2024-10-23
[Consul] Explicitly set the encryption key.

tpendragon:
Otherwise Consul is trying to grab the encryption key from setup, and it's not always working. This came up because consul was failing to connect to the cluster on our new box.

Error in the playbook when this happened:

```
fatal: [nomad-client-staging1.lib.princeton.edu]: FAILED! => {"changed": false, "cmd": "/usr/local/bin/consul acl token list -format json", "delta": "0:00:00.067755", "end": "2024-10-23 13:26:59.936259", "msg": "non-zero return code", "rc": 1, "start": "2024-10-23 13:26:59.868504", "stderr": "Failed to retrieve the token list: Unexpected response code: 500 (No known Consul servers)", "stderr_lines": ["Failed to retrieve the token list: Unexpected response code: 500 (No known Consul servers)"], "stdout": "", "stdout_lines": []}
```

Error in the consul log (`/var/log/consul/consul.log`):

```
Failed to join <server ip>: Remote state is encrypted and encryption is not configured
```

Issue that showed me what happened: https://github.com/hashicorp/consul-helm/issues/588

README from ansible-consul about this config option: https://github.com/ansible-collections/ansible-consul?tab=readme-ov-file#consul_raw_key

The reason it worked before:

https://github.com/ansible-collections/ansible-consul/blob/master/tasks/nix.yml#L79 pulls a key from a random consul bootstrapped server

https://github.com/ansible-collections/ansible-consul/blob/master/tasks/nix.yml#L105 saves that fact to a file - but runs it on a random node, in this case it ran it on the node that's not configured yet, so it failed to get the key.

It's easiest to just not have the magic and explicitly define the encryption key.


eliotjordan (approved):

hackartisan:
Thank you for this detailed write up @tpendragon

PR #5462, opened 2024-10-23
[Nomad/DPUL-C] Add configuration and read-only user to figgy prod db

### LINKED ISSUE #5459: Add dpul_c read-only user to figgy production database
We want our staging dpul-c to read from the figgy prod db.

eliotjordan:
Closes #5459


tpendragon (changes requested):
Need to reload the Figgy prod DB if the pg_hba changes.

Otherwise looks great!

tpendragon (approved):

Issue #5200, opened 2024-07-24
Create a playbook to update the allsearch-api's banner


maxkadel:
### User story
We want to make it so that folks like Esmé can update the banner on the allsearch-api (which currently updates the Allsearch, but may in the future update other applications as well) without necessarily having a working local environment for it, and without needing to ssh onto the servers. The hope is that this playbook could be added to ansible-tower, and users could log in there to update the information about the banner, including the banner text.

### Acceptance criteria
- [ ] The playbook should accept 5 user-supplied variables
  - [x] Environment (staging or production - default to staging)
  - [ ] Banner text - which may be a long string that may include html tags
  - [ ] alert_status - string - valid values are info, success, warning, or error
  - [x] dismissible - boolean
  - [x] autoclear - boolean
- [x] The playbook should then run the allsearch api `banner:update` rake task on one of the allsearch api servers for the given environment, using the user-supplied variables
- [ ] It should work correctly if the text includes commas (a rake task gotcha is that you have to escape commas when running it on the command line)
- [ ] The text, etc., on https://allsearch-api.princeton.edu/banner should update based on the user supplied variables

### Concrete example
When we recently needed to update the banner text on Orangelight with very little notice, it was difficult to get the change on the servers (in part because of the rake task comma escaping weirdness).

### Implementation notes, if any
- Allsearch api [banner rake task](https://github.com/pulibrary/allsearch_api/blob/main/lib/tasks/banner.rake)
- There is currently a separate [capistrano task to turn the banner on or off](https://github.com/pulibrary/allsearch_api/blob/main/lib/capistrano/tasks/banner.rake)

VickieKarasic:
We got this error with the latest iteration of the playbook: 
```
"stderr": "(in /opt/allsearch_api/releases/20241022171231)\nrake aborted!\nActiveRecord::ConnectionNotEstablished: connection to server at \"127.0.0.1\", port 5432 failed: Connection refused (ActiveRecord::ConnectionNotEstablished)\n\tIs the server running on that host and accepting TCP/IP connections?\n/opt/allsearch_api/shared/bundle/ruby/3.3.0/gems/activerecord-7.2.1.1/lib/active_record/connection_adapters/postgresql_adapter.rb:69:in `rescue in new_client'
```

Beck-Davis:
Closing. Can be reopened with clearer implementation strategy. Rake task vs. Ansible playbook?

PR #5463, opened 2024-10-23
Create playbook to update allsearch banner

Beck-Davis:
#5200


maxkadel:
- It should run in `/opt/allsearch_api/current`, not `/opt/allsearch_api/current/public`
- It occurs to me that this should only run against one machine in a given environment, since this updates the database and we don't want to do that via two machines
- Maybe we can give defaults for alert_status ("info"), dismissible (true), and autoclear (false)
- You may need to set the RAILS_ENV=staging or RAILS_ENV=production before running the rake task?

PR #5467, opened 2024-10-24
updates vSphere name of private network; prod & staging now match

acozine:
Up until now, the name of the private network was different in our vSphere prod and vSphere staging environments. @jkazmier-PUL updated vSphere so the name is the same in both envs, and it also matches OIT's designation. 

This PR updates the name of the network in the replace-a-vm playbook.
It also adds the network to the output comparing the old VM to its replacement.


carolyncole (approved):
Thanks @acozine!

PR #5470, opened 2024-10-25
fix broken replace-a-vm playbook

acozine:
The recently added diagnostic output that shows the network was failing because we were relying on a default setting for VMs in the public network.

This PR sets a network explicitly for both possible cases (public network and private network). It also removes the default settings for clarity.


carolyncole (approved):
Thanks @acozine

PR #5473, opened 2024-10-25
adds an EE with the CrowdStrike collection

acozine:
Creates a new EE for the security playbook.
Installs the crowdstrike.falcon collection and necessary tools. 
Omits some other collections we don't need for that playbook.


acozine:
EE created, uploaded, and in Tower now. Confirmed that it works for the security playbook.

kayiwa (approved):
OK! Thanks @acozine

PR #5217, opened 2024-07-31
Streamlining CrowdStrike automation

acozine:
Closes #5058. Eventually.

WIP for now. When this is done we will download the correct version of Falcon using the CrowdStrike API and the crowdstrike.falcon collection.

PR #5474, opened 2024-10-25
Updates security playbook to use crowdstrike collection

### LINKED ISSUE #5058: Use crowdstrike.falcon to install & update falcon
Currently we install an old version of falcon that we download internally, then CrowdStrike updates the version. The gap is getting large  - latest version is 7.16. 

If we use the collection instead of `command` and `systemd` tasks, we can download the appropriate version directly from CrowdStrike.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

As the version gap widens, upgrades get more difficult. We should keep up.

### Acceptance criteria
 - [ ] We download the appropriate version of Falcon sensor directly from CrowdStrike and install it on new VMs

We can use the `sensor_download_info` module to identify the appropriate version and pass the resulting hash to the `falcon_install` module to install it.

acozine:
Closes #5058.

Supersedes #5217, which had a nasty rebase that I'm not sure I got right.

Uses the crowdstrike.falcon collection instead of a handcrafted install. Depends on #5473, which adds a new EE for the security playbook.


acozine:
Tested today in Tower, works well. The EE has been uploaded, so this PR is ready to go.

PR #5478, opened 2024-10-28
break up the nfsserver tasks

kayiwa:
make the role more modular. Possibility of having all variables in the vars file
we will have more reusable variables and conditionals


maxkadel (changes requested):
I think this looks great! 

I think there might be a typo - `nfs_server_option` vs `nfs_server_options`

Also, would it be possible in the README to add an example where you override the default options?

PR #5481, opened 2024-10-28
initialize role for logging

kayiwa:
this is a galaxy scaffolding for a role for logs


kayiwa:
we will be recreating this from scratch

PR #5483, opened 2024-10-28
Restart nginx on crash.

### LINKED ISSUE #5480: [Passenger] Nginx should restart if it's killed because of OOM
## Description

We've had several boxes go down and not come back up because the nginx process was killed by `OOM` errors. While it's true we should figure out why the memory is spiking, we also shouldn't have to restart nginx - systemd should do that.

## Acceptance Criteria

- [ ] When an nginx passenger process is killed with `kill -9 <pid>` it should start again.

## Sudden Priority Justification

The longer we go without this in place, the more likely we're going to get drawn into fixing these incidents.

## First step

Google around on sustainable ways to add systemd auto-restart to nginx.

tpendragon:
Closes #5480

### DIFF
--- a/roles/passenger/files/systemd_override.conf
+++ b/roles/passenger/files/systemd_override.conf
@@ -1,0 +1,3 @@
+[Service]
+Restart=on-failure
+RestartSec=5s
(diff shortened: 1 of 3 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

PR #5491, opened 2024-10-30
Removed pdc-describe-prod2 from the load balancer

hectorcorrea:
Removes `pdc-describe-prod2` from the load balancer so that it stops taking new jobs and lets the `prod1` and `prod3` take care of new jobs.

See https://github.com/pulibrary/pdc_describe/issues/1983


kayiwa (approved):
OK! Thanks @hectorcorrea

Issue #5431, opened 2024-10-08
replace catalog-qa1 VM 


VickieKarasic:
### Describe the VM(s) you need

If you do not know the answers to some of these questions, that's okay. Please add the information you do have, and we will work together to gather any other information we need. You can also look at the [documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/new_vms_for_devs.md) in pul-it-handbook.

Which project are these VMs for?
- [ ] Project: catalog-qa

How many VMs do you need?
- [ ] Number of VMs:

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [x] FQDN: catalog-qa1.princeton.edu

What virtual resources does each VM need?
- [x] Number of CPUs per VM: 2
- [x] Amount of memory per VM (in GB): 8
- [x] Amount of storage/disk space per VM (in GB): 2 hard disks, 30 GB each 

What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [x] 22, 80
- [x] ssh, web-browsing

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)

### Relevant deadlines
If the new VMs must be ready by a certain date, please enter it here.

### Implementation notes, if any
- Blocked by #5430

VickieKarasic:
Machine has been replaced and we confirmed that the playbook in #5430 is working as it should.

PR #5497, opened 2024-10-31
[prancible] adding catalog-qa1 back to the loadbalancer 

VickieKarasic:
The machine has been rebuilt (see #5431)


sandbergja (approved):
Thanks, @VickieKarasic ! ☕

PR #5507, opened 2024-11-04
[lib_jobs] Add health checks against health endpoint

### LINKED ISSUE #5107: [lib-jobs] Use health.json for health checks

When lib-job's health checks are failing (health.json), it should not serve that application to the public.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria

Existing health check configuration: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/lib-jobs-staging.conf

### First Step

maxkadel:
Deploying bad branch to staging leads nginx to serve only the good branch

image

Staging version run on production load balancer - not run with production change against load balancer yet.
Closes #5107


kayiwa (approved):
OK! Thanks @maxkadel

Issue #4938, opened 2024-05-13
upgrade sftp servers to jammy


kayiwa:
### What maintenance needs to be done?

We need to upgrade our sftp servers to jammy jellyfish

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
 - When we attempted to set these servers up the SSSD software would not work with Active Directory. The software seems to be broken again (on focal). While the machine is bound to Active Directory. It is unable to actually authenticate successfully

### Acceptance criteria
 - [x] Ansible playbook runs and sets up and endpoint
 - [ ] All users can authenticate with AD users `getent passwd netid` provides the Active Directory UID/GID
 - [x] Document how to reset a re-bind to Active Directory

kayiwa:
related to #4879

kayiwa:
Ensure that SSSD_AD tests stop the playbook from running with useful information. https://github.com/pulibrary/princeton_ansible/pull/5511#discussion_r1836995439

kayiwa:
Blocked by this [SNow](https://princeton.service-now.com/service?id=ticket&table=sc_req_item&sys_id=5917f09a470eda547d2ae1dc416d43ce) ticket

kayiwa:
Add active directory certs. [New adventure](https://unix.stackexchange.com/questions/767029/ubuntu-failed-to-join-an-ad-over-tls-ssl-with-realmd)

acozine:
See also notes in [this private gdoc](https://docs.google.com/document/d/1EBV7a7w1URxuHbn2231qtJ5IL_rI98Ksndwk3ntaIRU/edit?tab=t.0).

kayiwa:
- [ ] grab MAC address of lib-sftp-test2
- [ ] wipe out lib-sftp-test2 
- [ ] register [machine on Active Directory](https://tools.princeton.edu/Dept/start.aspx?) (delete or reset)
- [ ] run [lib-sftp](playbooks/lib_sftp.yml)  --limit lib-sftp-test2.princeton.edu
- [ ] create private/pub keys on `almasftp` (user on prod) copy public keys to (test2 vm)
  - [ ] rsync contents of /alma (except aspace) from lib-sftp-prod1 to lib-sftp-test2
- [ ] create private/pub keys on `lib-aspacesftp` (user or prod) copy public keys to (test2 vm)
[...]

PR #5511, opened 2024-11-05
refactor the system ldap role

kayiwa:
we will create a system ldap role which can be used to bind vms to active directory
and sftp role and/or playbook can incorporate this role

This PR decouples system ldap from sftp. If it goes in it means any vm that needs to bind to active directory can use it. We've had an example of this need with the RDSS mflux vms. Our Active Directory binding is currently broken so this can be merged to main. A follow up PR for lib-sftp will be coming. 

Partially addresses #4938


roles/system_ldap/tasks/main.yml
```diff
@@ -1,97 +1,121 @@
...
     line: 'PasswordAuthentication yes'
 
-- name: system_ldap | allow users authentication
+- name: System_ldap | allow users authentication
   ansible.builtin.lineinfile:
     path: /etc/ssh/sshd_config
     state: present
     regexp: '^AllowUsers pulsys'
     line: '# AllowUsers pulsys'
   notify: restart sshd
```
acozine, line 21: Earlier in this PR the `restart sshd` handler got deleted, so this `notify` parameter won't work.

roles/system_ldap/tasks/main.yml
```diff
@@ -1,97 +1,121 @@
...
+  changed_when: false
+  when: running_on_server
+
+- name: System_ldap | verify domain join
+  ansible.builtin.command: realm list
+  register: realm_list
+  changed_when: false
+  when: running_on_server
+
+- name: System_ldap | debug realm list output
```
acozine, line 92: This will show the output to the user, but not stop the playbook if that output is wrong. Do we want to verify the domain join? Or was that put in for testing purposes?

roles/system_ldap/tasks/main.yml
```diff
@@ -1,97 +1,121 @@
...
-    password: "{{ almasftp_user_password }}"
-
-- name: system_ldap | Flush handlers to be able to use SSSD authentication
-  meta: flush_handlers
+    mode: "0700"
+    owner: "{{ ad_test_user }}"
+    group: "{{ ad_test_user }}"
+  when: running_on_server
+
+- name: System_ldap | test login with AD user
```
acozine, line 111: Does anything need to be restarted before we test the login?

roles/system_ldap/tasks/main.yml
```diff
@@ -1,97 +1,121 @@
...
-    - "/alma/bursar"
-    - "/alma/datasync_processing"
-    - "/alma/fund_adjustment"
-    - "/alma/invoice_status"
-    - "/alma/invoices"
-    - "/alma/people"
-    - "/alma/pod"
-    - "/alma/publishing"
-    - "/alma/recap"
-    - "/alma/scsb_renewals"
```
acozine, line 65: We still need these for building the SFTP servers, right? I don't see a change here that adds these values elsewhere. How would we retrieve them?

acozine (commented):
Nice! This is a great step toward flexibility in building systems - we can use an AD binding wherever we want to.

I had a couple of questions and suggestions, see below.

roles/system_ldap/tasks/main.yml
kayiwa: this will be a separate role call lib_sftp.

roles/system_ldap/tasks/main.yml
kayiwa: testing.

roles/system_ldap/tasks/main.yml
acozine: yep, maybe for now just comment them out instead of deleting them, so when we start the lib_sftp work, we can find this list easily

kayiwa: they are in a local branch which I will push up... If I knew how to git... ;-)

kayiwa: https://github.com/pulibrary/princeton_ansible/tree/i4938_sftp_role

roles/system_ldap/tasks/main.yml
kayiwa: confirmed that it doesn't. The ssh and sssd are configured. It will make a call to the AD endpoint.

roles/sssd_ad/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for system_ldap
+ad_domain: "pu.win.princeton.edu"
+ad_test_user: "{{ omit }}"
```
acozine, line 4: Supply a username, then default to omit for testing. `"{{ my_user | default(omit) }}"`

roles/sssd_ad/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
...
+ad_domain_controller: "pu.win.princeton.edu"
+ad_admin_user: "{{ omit }}"
```
acozine, line 6: and here

roles/sssd_ad/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
+---
+- name: Sssd_ad | change hostname to match AD
```
acozine, line 2: Can we add a comment about the reason for this (that Windows limits hostnames to 12 characters).

roles/sssd_ad/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+      [logging]
+        default = FILE:/var/log/krb5libs.log
+        kdc = FILE:/var/log/krb5kdc.log
+        admin_server = FILE:/var/log/kadmind.log
+
+      [libdefaults]
+        dns_lookup_realm = false
+        dns_lookup_kdc = true
+        ticket_lifetime = 24h
+        renew_lifetime = 7d
```
acozine, line 52: Nice! This should fix AD breakages due to losing the bind.

roles/sssd_ad/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+  changed_when: false
+  when: running_on_server
+
+- name: Sssd_ad | verify domain join
+  ansible.builtin.command: realm list
+  register: realm_list
+  changed_when: false
+  when: running_on_server
+
+- name: Sssd_ad | debug realm list output
```
acozine, line 95: Let's change this so it stops the playbook if the output is not correct at this stage.

roles/sssd_ad/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+  when: running_on_server
+
+- name: Sssd_ad | test login with AD user
+  ansible.builtin.command: su - {{ ad_test_user }} -c 'exit 0'
+  changed_when: false
+  ignore_errors: true  # Ignore errors if the user cannot log in yet
+  register: ad_login_result
+  when: running_on_server
+
+- name: Sssd_ad | debug AD login result
```
acozine, line 121: Make this fail if the output is not correct/complete.

roles/sssd_ad/tests/test.yml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: localhost
+  remote_user: root
+  roles:
+    - system_ldap
```
acozine, line 5: Delete this in favor of our molecule dir.

roles/sssd_ad/tasks/main.yml
kayiwa: only way to know what this output is will make this PR larger by incorporating this. https://github.com/pulibrary/princeton_ansible/tree/i4938_sftp_role

roles/sssd_ad/tasks/main.yml
kayiwa: Same here. I don't know what the output would be https://github.com/pulibrary/princeton_ansible/tree/i4938_sftp_role

PR #5514, opened 2024-11-05
[orangelight] add a health check

### LINKED ISSUE #5116: [orangelight] Use health.json for health checks

When Orangelight's health checks are failing (health.json), it should not serve that application to the public.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria

Existing health check configuration: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/imagecat_staging.conf#L35

### First Step

kayiwa:
have the load balancer configuration check the health gem generated file
to determine health of upstream VM

closes #5116

### DIFF
--- a/roles/nginxplus/files/conf/http/catalog-prod.conf
+++ b/roles/nginxplus/files/conf/http/catalog-prod.conf
@@ -79,7 +79,7 @@ server {
         proxy_http_version 1.1;
         limit_req zone=catalog-prod-ratelimit burst=80 nodelay;
         limit_req_status 429;
-        health_check interval=10 fails=2 passes=1 uri=/catalog/991234563506421;
+        health_check uri=/health.json interval=10 fails=3 passes=2;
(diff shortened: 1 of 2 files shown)


roles/nginxplus/files/conf/http/catalog-prod.conf
```diff
@@ -79,7 +79,8 @@ server {
         proxy_http_version 1.1;
         limit_req zone=catalog-prod-ratelimit burst=80 nodelay;
         limit_req_status 429;
-        health_check interval=10 fails=2 passes=1 uri=/catalog/991234563506421;
+        # health_check interval=10 fails=2 passes=1 uri=/catalog/991234563506421;
```
tpendragon, line 82: Can we get rid of the commented one?

tpendragon (changes requested):

tpendragon (approved):

PR #5515, opened 2024-11-05
[Figgy] Use health.json for health checks

### LINKED ISSUE #5103: [Figgy] Use health.json for health checks

When Figgy's health checks are failing (health.json), it should not serve that application to the public.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Acceptance criteria

Existing health check configuration: https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/figgy-staging.conf

### First Step

tpendragon:
Closes #5103

### DIFF
--- a/roles/nginxplus/files/conf/http/figgy-staging.conf
+++ b/roles/nginxplus/files/conf/http/figgy-staging.conf
@@ -50,7 +50,7 @@ server {
         proxy_cache_convert_head off;
         proxy_cache_methods GET HEAD;
         proxy_cache_key $scheme$request_method$proxy_host$request_uri;
-        health_check interval=10 fails=3 passes=2;
+        health_check uri=/health.json interval=10 fails=3 passes=2;
     }
     location /cable {
       proxy_pass http://figgy-staging;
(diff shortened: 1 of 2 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

roles/nginxplus/files/conf/http/figgy-staging.conf
```diff
@@ -50,7 +50,7 @@ server {
         proxy_cache_convert_head off;
         proxy_cache_methods GET HEAD;
         proxy_cache_key $scheme$request_method$proxy_host$request_uri;
-        health_check interval=10 fails=3 passes=2;
+        health_check uri=health.json interval=10 fails=3 passes=2;
```
maxkadel, line 53: Does it matter that this one doesn't have the slash in front of `health.json`?

maxkadel (changes requested):
Not sure if it works without the slash?

roles/nginxplus/files/conf/http/figgy-staging.conf
tpendragon: Good catch, fixed.

maxkadel (approved):
Thanks @tpendragon 🐲

PR #5523, opened 2024-11-06
[PDC_DESCRIBE] Adds server back to the load balancer

hectorcorrea:
Re-adds PDC Describe server 2 back to the load balancer so that it starts serving web requests again.

We had removed it on https://github.com/pulibrary/princeton_ansible/pull/5491 because we wanted to prevent it from taking more background jobs, but now that the Sidekiq queue on that server has been cleared it is safe to add the server back to the load balancer.

Closes https://github.com/pulibrary/pdc_describe/issues/1983


acozine:
Deployed successfully. Thanks @hectorcorrea

PR #5526, opened 2024-11-08
I5414 rhel common

### LINKED ISSUE #5414: the common role on RHEL based VMS
When running the common role on a rocky vm we need a conditional loop for name: common | install fdfind

And further investigate why name: common | Install procs fails to install successfully

kayiwa:
Make it possible for the common role to run on RHEL and Ubuntu VMs.

Closes #5414

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -5,7 +5,7 @@ duf_version: "0.8.1"
 dust_version: "0.9.0"
 fd_find_version: "9.0.0"
 configured_dependencies: []
-common_packages:
+common_ubuntu_packages:
   - acl
   - build-essential
   - curl
@@ -26,3 +26,21 @@ common_packages:
   - wget
   - zip
   - zlib1g-dev
+common_rhel_packages:
+  - curl
+  - "@Development tools"
+  - git
+  - htop
+  - jq
+  - libacl.x86_64
+  - libyaml-devel
+  - openssl-libs
+  - python3-devel
+  - python3-pip
+  - the_silver_searcher
+  - tmux
+  - tree
+  - unzip
+  - vim
+  - wget
+  - zip
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -5,42 +5,36 @@ duf_version: "0.8.1"
 dust_version: "0.9.0"
 fd_find_version: "9.0.0"
 configured_dependencies: []
-common_ubuntu_packages:
-  - acl
-  - build-essential
+common_shared_packages:
   - curl
+  - fd-find
   - git
   - htop
   - jq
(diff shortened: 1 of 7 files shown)


roles/common/defaults/main.yml
```diff
@@ -5,7 +5,7 @@ duf_version: "0.8.1"
 dust_version: "0.9.0"
 fd_find_version: "9.0.0"
 configured_dependencies: []
-common_packages:
+common_ubuntu_packages:
```
maxkadel, line 8: There's a lot of overlap between the Ubuntu and Rocky package lists - would it make sense to make a `shared_packages` list and then only have what's not in common in the Ubuntu and Rocky lists?

roles/common/tasks/main.yml
```diff
@@ -1,111 +1,90 @@
...
-    - apt-transport-https
-  when:
-    - ansible_os_family == "Debian"
 
-- name: common | install snapd
-  ansible.builtin.package:
-    name: "{{ item }}"
-    state: present
+- name: Common | Install Common packages
+  ansible.builtin.include_tasks: "{{ item }}"
```
acozine, line 10: We don't need to put these tasks in separate files and include them. The conditions will ensure that the correct packages get installed on the correct OS.

roles/common/tasks/main.yml
```diff
@@ -1,111 +1,90 @@
...
-  delay: 60
-  until: duf_repo is succeeded
+- name: Common | install duf
+  ansible.builtin.package:
+    state: present
   when:
-    - running_on_server
-    - ansible_os_family == "Debian"
+    - ansible_distribution == "Ubuntu"
+    - ansible_distribution_release == "jammy"
```
acozine, line 64: Could we add this to the ubuntu package list, instead of having a separate task for it?

roles/common/tasks/main.yml
```diff
@@ -1,111 +1,90 @@
...
+- name: Common | install fdfind
+  ansible.builtin.package:
+    name: fd-find
+    state: present
+  when:
+    - ansible_os_family == "Debian"
+    - running_on_server
 
-- name: common | install plocate (jammy)
+- name: Common | install plocate (jammy)
```
acozine, line 81: For these two also - can we add them to the Ubuntu package list so they get installed with the other apt packages?

acozine (commented):
Hooray for getting our RH-family systems working and passing CI!

I see a couple of ways we can streamline this, see below.

roles/common/tasks/main.yml
kayiwa: I should stop caring about bionic.

maxkadel (approved):
Looks good to me, thanks @kayiwa!

PR #5528, opened 2024-11-08
Add options for managing VMs

acozine:
Sometimes a VM gets into a weird state (mostly this seems to happen to Windows VMs) and using the `reboot guest OS` option does not work.

This PR offers two new options for interacting with VMs:
- power off, for a VM we don't need any more
- restart, which is the equivalent of powering a VM off and then on again

We need to update the survey on the Tower template to include these two options and make it clear that `restart` is only for stuck VMs.

Also alphabetizes the fields we pull from vSphere and includes the IP of the VM in the output.


kayiwa (approved):
OK! Thanks @acozine

PR #5538, opened 2024-11-14
[nginxplus] add jwt

kayiwa:
for nginxplus release 33 onwards the license will be deployed using JWT.
  This uploads our jwt to the endpoint

  https://docs.nginx.com/solutions/#jwt


roles/nginxplus/tasks/plus/setup-license.yml
```diff
@@ -1,16 +1,23 @@
...
-  with_items:
+    mode: "0644"
+  loop:
     - "{{ nginx_license.certificate }}"
     - "{{ nginx_license.key }}"
+
+- name: "Copy NGINX Plus jwt"
+  ansible.builtin.copy:
+    src: "files/licence/nginx-repo.jwt"
+    dest: /etc/ssl/nginx/licence.jwt
```
acozine, line 21: This looks like it follows the same pattern as the cert and license key above - same destination, same mode, also decrypted. Could we just add it to the loop in the previous task?

acozine (commented):
Thanks for getting this done! I see one way to simplify, see below.

roles/nginxplus/tasks/plus/setup-license.yml
kayiwa: If I am reading the [documentation correctly](https://docs.nginx.com/solutions/#jwt). Placing the file there will break it. See the additional information on "custom paths". On the other side of the upgrade process I think this will be OK. For now I like to stick to what the docs insist.

acozine: Ah, it should be a different path, gotcha.

PR #5542, opened 2024-11-18
I4938 sftp role

kayiwa:
bind lib-sftp role to Active Directory


kayiwa:
this work is continued on this PR #5686

PR #5543, opened 2024-11-19
upgrades mailcatcher version

acozine:
Our staging oawaiver machines were sending warnings about mailcatcher being down. Investigation showed that the init script was looking for mailcatcher 0.9.0, while the locally installed gem was 0.10.0.

### DIFF
--- a/group_vars/oawaiver/staging.yml
+++ b/group_vars/oawaiver/staging.yml
@@ -6,7 +6,7 @@ running_on_server: true
 install_mailcatcher: true
 mailcatcher_user: "pulsys"
 mailcatcher_group: "pulsys"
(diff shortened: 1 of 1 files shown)


jrgriffiniii (approved):

acozine:
A few breadcrumbs for future us. To find the currently installed gem:
```
oawaiver-staging1:~$ ls -la /usr/local/lib/ruby/gems/3.1.0/gems/ | grep mailcatcher
drwxr-xr-x   6 root root 4096 Sep 19 16:48 mailcatcher-0.10.0
```
To see the currently configured gem:
```
oawaiver-staging1:~$ sudo less /lib/systemd/system/mailcatcher.service
```

PR #5547, opened 2024-11-20
Streamline vm replace

acozine:
Because of [this bug](https://github.com/ansible-collections/community.vmware/issues/2124), we can't bring up a new VM with a custom MAC address. But we were able to streamline the replace-a-vm playbook a bit.

Also adds more debugging output to help when things go wrong.


carolyncole (approved):
Thanks @acozine

Issue #5548, opened 2024-11-21
[ResearchData-staging] Can't Deploy Drupal to Prds Staging Environment due to missing PHP extension. 


kevinreiss:
Blocks https://github.com/pulibrary/researchdata/pull/279. 

### Expected behavior
Site deploys via tower or capistrano

### Actual behavior
Include screenshots, or links and browser version if relevant.

### Relevant links and code snippets, if applicable

```
    - drupal/core 10.3.9 requires ext-gd * -> it is missing from your system. Install or enable PHP's gd extension.
```

### Steps to replicate
Deploy site, it fails. 

### Impact of this bug
Can't deploy staging site.

kevinreiss:
Still failing today with this error string
```
See full trace by running task with --trace)
The deploy has failed with an error: Exception while executing as deploy@prds-staging1: cd /var/www/researchdata/releases/20241125195407 && composer install --no-dev exit status: 2
cd /var/www/researchdata/releases/20241125195407 && composer install --no-dev stdout: Nothing written
cd /var/www/researchdata/releases/20241125195407 && composer install --no-dev stderr: Installing dependencies from lock file
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - drupal/core is locked to version 10.3.9 and an update of this package was not requested.
    - drupal/core 10.3.9 requires ext-gd * -> it is missing from your system. Install or enable PHP's gd extension.
  Problem 2
    - drupal/seven is locked to version 1.0.0 and an update of this package was not requested.
    - drupal/core 10.3.9 requires ext-gd * -> it is missing from your system. Install or enable PHP's gd extension.
    - drupal/seven 1.0.0 requires drupal/core ^9 || ^10 -> satisfiable by drupal/core[10.3.9].
```

acozine:
Closed by #5549.

PR #5549, opened 2024-11-21
Address #5548"

kevinreiss:


acozine (commented):
The first commit added a package called `php{{ php_version }}-gd` but the second commit didn't incorporate it everywhere - is that because we don't need it? Or was that an oversight?

acozine:
We ran this against staging and it fixed the can't-deploy problem outlined in #5548.

PR #5562, opened 2024-11-25
[static-tables] disable the WAF

kayiwa:
this PR disables the WAF on the following applications

catalog
solr8-prod
static-tables
static-prod (added by mistake)


maxkadel (approved):
Thanks @kayiwa  - although my dream of updating the WAF instead of disabling it when we run into problems remains.

christinach (approved):
Thanks @kayiwa

PR #5564, opened 2024-11-25
[lib-solr8d-staging] Update conf with: app_protect_enable off

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

Issue #5560, opened 2024-11-25
Move to cicognara.org domain to static_production


kevinreiss:
### What maintenance needs to be done?
The cicognara site has been replaced a static website that is already part of the static playbook and is available in both the staging and prod static environments. The existing cicognara role is still attached to the cicognara.org domain. We need to move this 

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Relevant deadlines
- [ ] Make switch Tuesday, 12/3/2024 in the afternoon. End users available to confirm production site is functional at 3 p.m. eastern. 

### Why is this maintenance needed?
- Current production cicognara environment is being retired. We need to move the production domain name to the "new" static site. 

### Acceptance criteria
 - [x] Add a "named" site for the [static version](https://github.com/pulibrary/digital-cicognara-library) of cicognara to the the production static role [group vars](https://github.com/pulibrary/princeton_ansible/blob/f59c78e907c7eb8d7866cce83097ffcc1821f401/group_vars/libstatic/production.yml). 
 - [x] Run the static playbook against production to install the added named site's repo to the environment. 
 - [x] nginxplus playbook is updated to point cicognara.org at the static production environment. Currently configured [here](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/cicognara-org.conf). We will need to move this domain over to this [ngnixplus config file](https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/lib-static-prod.conf). Models are milberg.princeton.edu or davies-project.princeton.edu.  
 - [x] nginxplus playbook serves up the "new" cicognara site from it's new doc root in the static production environment.
 - [x] Check example URLs to confirm the "new" site is up. After the switch these two URLs should look identical: https://cicognara-staging.lib.princeton.edu/catalogo/itempages/15.html and https://cicognara.org/catalogo/itempages/15.html 
 - [x] Add a redirect pattern for URLs that have the /catalog/XXX path to this new pattern /catalogo/itempages/XXX.html. 
 - [ ] Ask end users (Ming Aguilar, Rachael Nutt, Cliff Wulfmann) to confirm new production site is working as expected. 

### Implementation notes, if any
We can model the configuration for production off the configuration for the [staging site](https://cicognara-staging.lib.princeton.edu/) in libstatic. 

[...]

PR #5570, opened 2024-11-27
[cicognara] migrate to static site

VickieKarasic:
Related to #5560 

We deleted cicognara_config file

We have added a new block on static-prod for ports 80 and 443

We've pointed it to the static site config


maxkadel (approved):
Thanks! 🎉

acozine:
The playbook that's deleted here, libstatic.yml, is obsolete. There's a newer playbook called static.yml, and inventory entries for `static_<env>`.

PR #5580, opened 2024-12-03
Block some bad yandex crawlers that were overwhelming the catalog

sandbergja:


christinach (approved):
Thanks @sandbergja @christinach @kayiwa !

PR #5583, opened 2024-12-04
[denyhost] remove princeton network

kayiwa:
committed an unnamed network because it is part of rfc 1918
this caused traffic mayhem for vpn'ed users

Co-authored-by: Christina Chortaria <christinach@users.noreply.github.com>
Co-authored-by: Jane Sandberg <sandbergja@users.noreply.github.com>
Co-authored-by: Kevin Reiss <kevinreiss@users.noreply.github.com>
Co-authored-by: Max Kadel <maxkadel@users.noreply.github.com>
Co-authored-by: regineheberlein <regineheberlein@users.noreply.github.com>
Co-authored-by: Vickie Karasic <vickiekarasic@users.noreply.github.com>

PR #5582, opened 2024-12-04
Ban crawlers from VIETTEL-CAMBODIA, unnamed crawler, DIGITALOCEAN

christinach:

PR #5588, opened 2024-12-04
Add a check to make sure we don't block ourselves using the denylist

sandbergja:
We accidentally blocked ourselves at the firewall level in #5582 (fixed in #5583).  This adds a check to CI to alert us if we're about to do that again.


christinach (approved):
Thanks @sandbergja ! 🍄

PR #5591, opened 2024-12-04
Block another bad bot that sent excessive traffic to the catalog

sandbergja:
I ran this playbook on 4 December


christinach (approved):
Thanks @sandbergja ! 🍄

PR #5593, opened 2024-12-04
[orangelight] increase the passenger max pool size

sandbergja:
This increases the number of simultaneous requests that passenger will work on. The tradeoff is that more simultaneous requests will require more memory to process.

According to the passenger documentation, we could safely process about 28 simultaneous requests (rather than our current 6) with the current memory usage patterns and memory available on the boxes:
https://www.phusionpassenger.com/library/config/nginx/optimization/#tuning-the-application-process-and-thread-count

Perhaps this could make the catalog more resilient to large spikes in bot traffic?

Creating as a draft PR so that we can discuss.  I ran this on catalog3 today (December 4)


christinach (approved):
thanks @sandbergja for looking into this 🍄 !

christinach:
Ran on all production vms

PR #5601, opened 2024-12-05
[Nomad] Prune podman containers nightly.

### LINKED ISSUE #5602: [Nomad] Old images are filling up the disk on nomad staging boxes.
A DPUL-C Deployment failed to place a box. When checking the logs for that allocation in the Nomad UI it said that it was out of disk space. I went to check the box and the disk sure was full - when I navigated `du -df` deeply enough I found that it was in a container overlays directory. I figured it was probably either backed up logs (which I was pretty sure got cleaned up), or backed up image downloads.

I knew `podman prune` was a thing, so I did some looking in Nomad about automatic image pruning and found that the Docker driver auto-prunes old downloaded images (from previous deploys), but Podman doesn't, so we need a different system for that.

tpendragon:
We were filling up disk with unused images.

Pulled from https://github.com/hashicorp/nomad-driver-podman/issues/363 and https://github.com/dani-garcia/vaultwarden/discussions/4362

Closes #5602

### DIFF
--- a/roles/pul_nomad/files/auto_update_override.conf
+++ b/roles/pul_nomad/files/auto_update_override.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStartPost=
+ExecStartPost=/usr/bin/podman image prune -a -f
--- a/roles/pul_nomad/handlers/main.yml
+++ b/roles/pul_nomad/handlers/main.yml
@@ -1,2 +1,5 @@
 ---
 # handlers file for pul_nomad
+- name: reload systemd
+  ansible.builtin.systemd_service:
+    daemon_reload: true
--- a/roles/pul_nomad/tasks/main.yml
+++ b/roles/pul_nomad/tasks/main.yml
@@ -102,12 +102,46 @@
   when:
     - "ansible_os_family == 'RedHat'"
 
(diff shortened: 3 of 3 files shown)


roles/pul_nomad/files/auto_update_override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStartPost=
```
hackartisan, line 2: Do we want this key in here twice? What does that do?

hackartisan (changes requested):
it would be really nice to have a little bit in the issue about how this problem manifested. I assume there was a deploy error, anything you can share about which screen in nomad revealed the problem? Once it was clear there was a disk space issue I assume you connected to that box and did usual disk space investigation.

roles/pul_nomad/files/auto_update_override.conf
tpendragon: I have no clue, that's just what was in the issue I linked to and it worked

tpendragon:
@hackartisan Added more info to https://github.com/pulibrary/princeton_ansible/issues/5602

hackartisan (approved):
Thank you!

PR #5605, opened 2024-12-06
Increase passenger thread count to 30

rladdusaw:
After analyzing our resource usage on our Orangelight webservers we decided to increase the Passenger thread count to 30 to handle more requests with the available resources.


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

Issue #5590, opened 2024-12-04
Create Red Hat-family Linux VMware Template for PostgreSQL 13


kayiwa:
Our current Ubuntu Bionic PostgreSQL 13 server is past  its end-of-life. To ensure continued support and security, we need to migrate to a new server based on a supported operating system. This issue outlines the creation of a Red Hat Linux VMware template to facilitate deployment of a replacement PostgreSQL 13 server in our vSphere environment.

Tasks:

- [x] Create a Red Hat Linux VM (local)
- [x] Convert the VM to a Template:
- [x] Use the latest supported Red Hat Enterprise Linux release.
- [x] Install and Configure PostgreSQL 13:
- [x] Install the PostgreSQL 13 server software.
- [x] Configure PostgreSQL for optimal performance and security, considering factors like:
- [x] Authentication methods
- [x] Database encoding
- [x] Connection limits
- [x] Documentation:
- [x] Create documentation outlining the steps to deploy a new PostgreSQL 13 server from the template.

Acceptance Criteria:

- [x] A Red Hat Linux VMware template is available in vSphere.
- [x] The template includes a fully installed and configured PostgreSQL 13 server.

Benefits:

- [ ] Simplified Deployment: Streamlines the process of deploying new PostgreSQL 13 servers.
- [ ] Consistency: Ensures all new servers are configured identically, reducing potential for errors.
- [ ] Maintainability: Provides a centralized template for updates and modifications.
- [ ] Security: Utilizes a supported operating system with the latest security updates.

acozine:
Confirmed today that Rocky Linux also installs PostgreSQL 13 by default, so we can use our existing Rocky template for this work.

kayiwa:
Change names of db and role to avoid dashes

acozine:
We also need to clean up the artifacts of the migration - we have records for ansible-db1 and postgres-prod-bionic, make sure both are gone.

acozine:
Related to #5508

PR #5606, opened 2024-12-06
I5590 migrate ansible tower db

kayiwa:
Install and configure postgresql 13 on Rocky Linux

Related to #5590

### DIFF
--- a/roles/postgresql/defaults/main.yml
+++ b/roles/postgresql/defaults/main.yml
@@ -41,6 +41,9 @@ postgres_packages:
   - postgresql-client-{{ postgres_version }}
   - postgresql-contrib-{{ postgres_version }}
 
+rh_postgres_packages:
+  - postgresql-server
+
 postgres_client_packages:
   - libpq-dev
   - python3-psycopg2
@@ -50,6 +53,8 @@ postgres_service_name: "postgresql-{{ postgres_version }}"
 postgres_database_owner: "postgres"
 postgres_cluster_name: "main"
 postgres_conf_directory: "/etc/postgresql/{{ postgres_version }}/{{ postgres_cluster_name }}"
+
+rh_postgres_conf_directory: "/var/lib/pgsql/data"
 
 db_clusteradmin_password: "changethis"
 postgres_admin_password: "change"
@@ -80,6 +85,8 @@ postgres_ext_postgis_deps:
   - postgresql-{{ postgres_version }}-postgis-{{ postgres_ext_postgis_version }}-scripts
 
 postgres_data_directory: "/var/lib/postgresql/{{ postgres_version }}/{{ postgres_cluster_name }}"
+
+rh_postgres_data_directory: "/var/lib/pgsql/data"
 
 # HBA Connections
 postgres_hba_entries: []
--- a/roles/postgresql/defaults/main.yml
+++ b/roles/postgresql/defaults/main.yml
@@ -56,6 +56,10 @@ postgres_conf_directory: "/etc/postgresql/{{ postgres_version }}/{{ postgres_clu
 
 rh_postgres_conf_directory: "/var/lib/pgsql/data"
 
+# On RHEL based machines the path below will be needed
+# postgres_log_file: /var/lib/pgsql/data/pg_log/pg.csv
+postgres_log_file: "/var/lib/postgresql/{{ postgres_version }}/main/pg_log/pg.json"
+
 db_clusteradmin_password: "changethis"
 postgres_admin_password: "change"
 sudoers_dir: "/etc/sudoers.d"
--- a/roles/postgresql/defaults/main.yml
+++ b/roles/postgresql/defaults/main.yml
@@ -41,7 +41,7 @@ postgres_packages:
   - postgresql-client-{{ postgres_version }}
   - postgresql-contrib-{{ postgres_version }}
 
-rh_postgres_packages:
+dnf_postgres_packages:
   - postgresql-server
 
 postgres_client_packages:
--- a/roles/postgresql/tasks/main.yml
+++ b/roles/postgresql/tasks/main.yml
@@ -52,6 +52,7 @@
 - name: PostgreSQL | Ensure PostgreSQL is enabled and started
   ansible.builtin.service:
     name: postgresql
+    state: started
     enabled: true
   when:
     - running_on_server
--- a/roles/postgresql/tasks/main.yml
+++ b/roles/postgresql/tasks/main.yml
@@ -4,30 +4,29 @@
     pkg: ca-certificates
     state: present
   when:
-    - ansible_distribution == "Ubuntu"
+    - ansible_os_family == "Debian"
 
 - name: PostgreSQL | Add PostgreSQL repository apt-key
   ansible.builtin.apt_key:
     url: "{{ postgres_apt_key_url }}"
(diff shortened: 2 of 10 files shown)


acozine (commented):
Hooray for getting postgresql 13 working on Rocky. I have a few ideas for making this easier for Future Us to manage and understand, see below.

roles/postgresql/templates/rh_postgresql.conf.j2
```diff
@@ -0,0 +1,241 @@
...
+#ssl_crl_dir = ''
+
+
+#------------------------------------------------------------------------------
+# RESOURCE USAGE (except WAL)
+#------------------------------------------------------------------------------
+
+# - Memory -
+
+shared_buffers = {{ buffer_memory }} # min 128kB
```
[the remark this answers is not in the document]
kayiwa: It computes it based on the server running it. https://github.com/pulibrary/princeton_ansible/blob/e6d7bf6524760e972ad1b1ae222b35e31073a5c9/roles/postgresql/defaults/main.yml#L8

roles/postgresql/tasks/main.yml
```diff
@@ -3,76 +3,126 @@
...
+    name: '{{ item }}'
+    state: present
+  loop:
+    - "{{ dnf_postgres_packages }}"
+  when:
+    - postgresql_is_local
+    - ansible_os_family == "RedHat"
 
-- name: PostgreSQL | install postgresql client packages
+- name: PostgreSQL | Ensure PostgreSQL is enabled and started
```
acozine, line 52: I think to enable AND start it, we need `state: started` here as well as `enabled: true`.

roles/postgresql/tasks/main.yml
```diff
@@ -3,76 +3,126 @@
...
+  ansible.builtin.service:
+    name: postgresql
+    enabled: true
```
acozine, line 55:
```suggestion
    state: started
    enabled: true
```

roles/postgresql/defaults/main.yml
```diff
@@ -41,6 +41,9 @@ postgres_packages:
   - postgresql-client-{{ postgres_version }}
   - postgresql-contrib-{{ postgres_version }}
 
+dnf_postgres_packages:
```
acozine, line 44: Where do we use this var?

roles/postgresql/tasks/main.yml
```diff
@@ -3,76 +3,126 @@
...
+  when:
+    - running_on_server
+
+- name: PostgreSQL | Check if PostgreSQL data initialization logs exist (RedHat)
```
acozine, line 59: With the `creates` line in the next task, I don't think we need this task or the condition on the next task. IIRC, when you pass `creates`, Ansible checks for the file/dir listed, and only executes the task if it doesn't exist.

roles/postgresql/tasks/main.yml
```diff
@@ -3,76 +3,126 @@
...
+    - ansible_os_family == "RedHat"
+
+- name: PostgreSQL | Initialize PostgreSQL database (RedHat)
+  ansible.builtin.command:
+    cmd: /bin/postgresql-setup --initdb --unit postgresql
+  args:
+    creates: "/var/lib/pgsql/initdb_postgresql.log"
+  when:
+    - ansible_os_family == "RedHat"
+    - not pgsql_data_install.stat.exists
```
acozine, line 74: As mentioned above, I think this line can go, along with the `stat` task.

roles/postgresql/tasks/main.yml
```diff
@@ -81,24 +131,27 @@
...
   register: configure_postgres
-  when: postgresql_is_local
+  when:
+    - postgresql_is_local
+    - ansible_os_family == "Debian"
   notify: restart local postgres
 
 - name: PostgreSQL | rotate the logs for postgresql
-  template:
+  ansible.builtin.template:
```
acozine, line 142: Do both PostgreSQL installations write logs to the same place? Or do we need to make sure the logrotate template can handle two locations?

roles/postgresql/tasks/main.yml
```diff
@@ -145,26 +199,55 @@
...
+
+- name: PostgreSQL | configure logical replication
+  ansible.builtin.include_tasks: logical_replication.yml
   when:
     - postgresql_cluster != false
     - inventory_hostname == postgresql_cluster.leader
     - postgresql_logical_replication != false
-- include_tasks: create_db.yml
-- include_tasks: create_users.yml
-- include_tasks: create_read_only_user.yml
```
acozine, line 169: By moving these out of this mega-task, are we losing the conditions? With this change, I think the `create_db` and `create_users` includes would run on all boxes. Maybe a block is the right answer here?

roles/postgresql/tasks/main.yml
```diff
@@ -145,26 +199,55 @@
...
+    - postgres_host is not none
+    - postgresql_is_local
+    - ansible_os_family == "RedHat"
+    - running_on_server
+  become: true
+  notify:
+    - restart local postgres
+
+- name: PostgreSQL | configure access to postgres server
+  ansible.builtin.include_tasks: pg_hba.yml
```
acozine, line 224: How is this different from the task at the bottom of this file? Are we creating the pg_hba.yml list of allowed servers twice?

acozine (commented):
This is a big step forward, thanks! I saw a few things I wasn't sure about - see below.

roles/postgresql/defaults/main.yml
kayiwa: it was your suggestion ;-) and it is used here https://github.com/pulibrary/princeton_ansible/pull/5606/files/872428d142dcd542225d41e453e7c9a79d52125d#diff-68e937462c82c14f5ad880a4c3a8f2719e4a288d75e0a22641b524102414576bR47

roles/postgresql/tasks/main.yml
kayiwa: they do not (short answer) 
Modify  template for RHEL  based machines (pair with @acozine)

roles/postgresql/tasks/main.yml
```diff
@@ -145,26 +199,55 @@
...
     - inventory_hostname == postgresql_cluster.leader
-- include_tasks: warm_standby_follower.yml
+
+- name: PostgreSQL | configure warm standy follower
+  ansible.builtin.include_tasks: warm_standby_follower.yml
   when:
     - postgresql_cluster != false
     - inventory_hostname in postgresql_cluster.followers
-- include_tasks: logical_replication.yml
+
```
kayiwa, line 238: Put these back.

roles/postgresql/tasks/main.yml
acozine: Addressed by 96e44bff59a6f840d70cb4831eb231d3d392aa0d

roles/postgresql/tasks/main.yml
kayiwa: they are still in there. They now pass my linter and use FQRN https://github.com/pulibrary/princeton_ansible/blob/9afb46ced7138a474716d288ccaeda5bc40775a1/roles/postgresql/tasks/main.yml#L246

tpendragon (approved):

PR #5608, opened 2024-12-06
Add catalog5

### LINKED ISSUE #5604: [Catalog-prod] 5th web server
### Describe the VM(s) you need

A new web server for the catalog. Name would be catalog5. Same VM as catalog1,2,3,4

Which project are these VMs for?
- [x] Project: Orangelight

How many VMs do you need?
- [x] Number of VMs: 1

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [x] FQDN: catalog5.princeton.edu

What virtual resources does each VM need?
- [ ] Number of CPUs per VM:
- [ ] Amount of memory per VM (in GB):
- [ ] Amount of storage/disk space per VM (in GB):

What firewall changes should be made? (for example, port 22 for SSH, etc.) - same as previous Catalog web servers

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [ ] Additional resources:

### Other Todos

- [x] Confirm the new machine IP is registered with ponyexpress
- [x] Add the new machine IP to the list of local machines allowed to connect to bibdata's protected patron endpoint. 

Please get configuration from existing web boxes.

Beck-Davis:
Add catalog5 to inventory and nginx conf.

Closes #5604


sandbergja (approved):
Thanks, @Beck-Davis ! 🐻‍❄️

PR #5609, opened 2024-12-06
[allsearch api] compile ruby with the YJIT in staging

sandbergja:
Recent versions of Ruby 3.3 will automatically compile the yjit if they notice that the rust compiler is installed.

Rails 7.2 (which Allsearch API uses) automatically uses the YJIT if it is available.

Trying this out to see if it boosts performance at all (which we can monitor in the datadog APM).

One way to check if the YJIT is installed is with the following shell command:

```
ruby -e 'puts "YJIT is#{defined?(RubyVM::YJIT) ? "" : " not"} installed"'
```

### DIFF
--- a/roles/ruby_s/tasks/install_from_source.yml
+++ b/roles/ruby_s/tasks/install_from_source.yml
@@ -1,4 +1,10 @@
 ---
+- name: ruby_s | install the rust compiler to compile Ruby YJIT
+  ansible.builtin.apt:
+    name: rustc
+    state: present
+  when: ruby_yjit
+
 - name: ruby_s | lookup download path and checksum from ruby-lang release index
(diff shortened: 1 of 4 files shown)


roles/ruby_s/tasks/install_from_source.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: ruby_s | install the rust compiler to compile Ruby YJIT
+  ansible.builtin.apt:
+    name: rustc
+    state: present
+  when: ruby_yjit
```
kayiwa, line 6: 😎

kayiwa (approved):
OK! Thanks @sandbergja

PR #5613, opened 2024-12-09
[Pulfalight] Add nginx stats & logs for pulfalight to Datadog.

tpendragon:
This does a few things:

1. Adds nginx logs
1. Filters local health checks out of nginx logs (for all apps).
1. Forwards the real IP so they don't get filtered, and we can group on them in DataDog.

Refs pulibrary/pulfalight#1469

### DIFF
--- a/roles/passenger/templates/passenger.j2
+++ b/roles/passenger/templates/passenger.j2
@@ -23,6 +23,7 @@ server {
 
 # this filters out the load-balancer IPs, so we don't log health checks from them
 map $remote_addr $logging_enabled {
+  "127.0.0.1" 0;
   {% for ip in passenger_real_ip_from %}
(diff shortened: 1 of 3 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

roles/passenger/templates/passenger.j2
```diff
@@ -23,6 +23,7 @@ server {
 
 # this filters out the load-balancer IPs, so we don't log health checks from them
 map $remote_addr $logging_enabled {
+  "127.0.0.1" 0;
```
jrgriffiniii, line 26: 💯

sandbergja (approved):
Thanks, @tpendragon and @hackartisan !

PR #5617, opened 2024-12-10
Comment out deploy_id_rsa_private_key in bibdata commons so that we can run the bibdata playbook successfully

christinach:
See related sftp ticket: https://github.com/pulibrary/princeton_ansible/issues/4938


sandbergja (approved):
Thanks, @christinach and @rladdusaw 🐬 🐈‍⬛

PR #5624, opened 2024-12-10
Figgy worker boxes appear to have more capacity for sidekiq threads

hackartisan:
And we have a backed-up queue.

This has been run already on most of figgy's workers.


tpendragon (approved):

hackartisan:
We looked at metrics after this change and job duration increased; we think the limit is related to network capacity. i.e. adding more workers doesn't allow us to take advantage of available memory and cpu capacity.

PR #5639, opened 2024-12-17
[MFLUX] Add roles for users to query

carolyncole:
Makes sure the docker development instance can run the content queries when logged in via the cas user

Create a new instance and published it with this code `ansible-playbook playbooks/mflux-docker.yml --ask-vault-pass -e platform_base=arm64 -e docker_base_image=rockylinux:9 -e docker_tag=cas-login` for folks to use on thier machines


kayiwa (approved):
You probably want someone from RDSS to look this over. OK! Thanks @carolyncole

PR #5641, opened 2024-12-17
[nginxplus] fail2ban blacklight filter

kayiwa:
we have had queries that have a large number of facets. This filter will
look for 8 of these for blacklight. It works with blacklight 7. IPs
found will be banned

Co-authored-by: Jason Casden <cazzerson@users.noreply.github.com>
Co-authored-by: David Romani <unc-david-romani@users.noreply.github.com>

### DIFF
--- a/roles/nginxplus/files/fail2ban/nginx-bad-httpbots.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-bad-httpbots.conf
@@ -1,0 +1,8 @@
+i[nginx-f_inclusive]
+enabled = true
+port    = http,https
+filter  = nginx-f_inclusive
+logpath = /var/log/nginx/access.log
+maxretry = 3
+findtime = 3600
+bantime  = 3600
--- a/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
@@ -1,3 +1,4 @@
 [Definition]
 failregex = ^\{\"remote_ip\"\: \"<ADDR>\".{45,80}\"uri\"\:.{4,}?(?:\&?f(?:_inclusive)?%%5B.{5,}?){20,}
+nginx-f_inclusive.conf
 ignoreregex =
--- a/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
@@ -1,4 +1,3 @@
-nginx-f_inclusive.conf
 [Definition]
-failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){8,}
+failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){20,}
 ignoreregex =
--- a/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
@@ -1,4 +1,4 @@
 nginx-f_inclusive.conf
 [Definition]
-failregex = ^<HOST> .+?([&?]f(_inclusive)?%%5B.*?){8,}
+failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){8,}
 ignoreregex =
--- a/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
@@ -1,3 +1,3 @@
 [Definition]
-failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){20,}
+failregex = ^\{\"remote_ip\"\: \"<ADDR>\".{45,80}\"uri\"\:.{4,}?(?:\&?f(?:_inclusive)?%%5B.{5,}?){20,}
 ignoreregex =
--- a/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
+++ b/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
@@ -1,0 +1,4 @@
+nginx-f_inclusive.conf
+[Definition]
+failregex = ^<HOST> .+?([&?]f(_inclusive)?%%5B.*?){8,}
+ignoreregex =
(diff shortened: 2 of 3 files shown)


roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
```diff
@@ -0,0 +1,4 @@
+nginx-f_inclusive.conf
+[Definition]
+failregex = ^<HOST> .+?([&?]f(_inclusive)?%%5B.*?){8,}
```
christinach, line 3: @kayiwa, @sandbergja and I are reviewing this together and the question we have is : have you identified patterns in our logs that come from a bot and thy are 8 facets and more?

Also in the catalog using the advanced search it's not hard for a human to create queries with 8 facets and more.

roles/nginxplus/files/fail2ban/nginx-bad-httpbots.conf
```diff
@@ -0,0 +1,8 @@
+i[nginx-f_inclusive]
+enabled = true
+port    = http,https
+filter  = nginx-f_inclusive
+logpath = /var/log/nginx/access.log
+maxretry = 3
```
christinach, line 6: Can we increase the maxretry or decrease the findtime. 
Have you set the maxretry and the findtime based on the sequence of the queries you identifies in the logs?

christinach (changes requested):
@kayiwa thank you for the PR! 
It's a great start but both @sandbergja and I think that this PR as it is, it's too restrictive and will ban legitimate human users. 
Both of us tried a query in advanced search and we were banned from UNC.

roles/nginxplus/files/fail2ban/nginx-bad-httpbots.conf
kayiwa: It was deliberate aggressive at first in the hops that it would squeal.

roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
kayiwa: this was based on the UNC input.

acozine:
Current CI failure is in the idempotency check - it looks like Ansible deletes the file each time and recreates it, instead of checking whether it's there and adding or updating it if needed. Here's the sequence of tasks on the second run:

```
 TASK [nginxplus : Nginxplus | identify unwanted configuration files that we should delete] ***
  ok: [instance]
  
  TASK [nginxplus : Nginxplus | Delete unwanted configuration files] *************
  changed: [instance] => (item={'path': '/etc/fail2ban/jail.d/nginx-bad-httpbots.conf', 'mode': '0644', 'isdir': False, 'ischr': False, 'isblk': False, 'isreg': True, 'isfifo': False, 'islnk': False, 'issock': False, 'uid': 0, 'gid': 0, 'size': 166, 'inode': 3171011, 'dev': 48, 'nlink': 1, 'atime': 1734460073.7526984, 'mtime': 1734460073.4446995, 'ctime': 1734460073.7526984, 'gr_name': 'root', 'pw_name': 'root', 'wusr': True, 'rusr': True, 'xusr': False, 'wgrp': False, 'rgrp': True, 'xgrp': False, 'woth': False, 'roth': True, 'xoth': False, 'isuid': False, 'isgid': False})
  
  TASK [nginxplus : Nginxplus | Add nginx-badbots filter] ************************
  ok: [instance]
  
  TASK [nginxplus : Nginxplus | Add nginx-badbots restriction] *******************
  ok: [instance]
  
  TASK [nginxplus : Nginxplus | Add nginx-bad-httpbots restriction] **************
  changed: [instance]
  ```

File: roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf (hunk 1 above)
Comment (sandbergja) [inline roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf:3]:
```suggestion
failregex = ^<HOST> .+?([&?]f(_inclusive)?%5B.*?){8,}
```

I think we may only need one `%`

roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
```diff
@@ -0,0 +1,4 @@
+nginx-f_inclusive.conf
```
sandbergja, line 1: We need to remove this first line,  otherwise I get the error:

```
Wrong config file: File contains no section headers.
file: '/Users/sandbergj/repos/princeton_ansible/roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf', line: 1
'nginx-f_inclusive.conf\n'
```

roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
sandbergja: Scratch that, I think we actually need something like this since access.log is a json log:

```suggestion
failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){8,}
```

sandbergja (changes requested):
Some more ideas after experimenting with `fail2ban-regex` locally with this branch

christinach:
@kayiwa in our stand up we were reviewing https://app.honeybadger.io/projects/54399/faults/115717920. We're thinking to block 20 facets and more.

kayiwa:
20 facets or greater (from prancible openhouse)

christinach:
@kayiwa thanks for making the change I'll bring this to the team to review and will follow up on the PR.

rladdusaw:
We will [test this](https://github.com/pulibrary/princeton_ansible/tree/main/roles/nginxplus#testing-a-fail2ban-rule) against logs and confirm that it's not banning human users.

roles/nginxplus/files/fail2ban/nginx-f_inclusive.conf
```diff
@@ -0,0 +1,3 @@
+[Definition]
+failregex = ^\{\"remote_ip\"\: \"<HOST>\".*?\"uri\"\:.+?([\&]?f(_inclusive)?%%5B.*?){20,}
```
sandbergja, line 2: Sorry, one last suggestion!  While testing this,  I found that the 20 we requested makes the regular expression extremely slow.  For example, processing 5000 access.log entries took over 2 minutes.

The following helps to speed it up (still kinda slow, the same 5000 access.log entries take just under a minute now):
- changing `.+` and `.*` to reasonable ranges like `.{45,80}` made this quite a bit faster.
- Changing `<HOST>` to `<ADDR>`
- Using non capturing groups `(?:` instead of `(`

As a side note, [I found this blog post very useful when trying to speed up the regex](https://blog.gistre.epita.fr/posts/thomas.corbiere-2024-09-22-optimizing-regular-expressions/)

```suggestion
failregex = ^\{\"remote_ip\"\: \"<ADDR>\".{45,80}\"uri\"\:.{4,}?(?:\&?f(?:_inclusive)?%%5B.{5,}?){20,}
```

sandbergja (commented):
Thanks for all your work on this, @kayiwa !  One more suggestion to speed up the regular expression -- what do you think?  Then I want to hit the green button! :-)

sandbergja (approved):
🎉🎉🎉 🤖🤖🤖

christinach (approved):
Thank you @kayiwa

PR #5643, opened 2024-12-18
[DPUL] update ruby and node

hackartisan:
refs pulibrary/dpul#1584


kayiwa (approved):
OK! Thanks @hackartisan

PR #5644, opened 2024-12-18
[abid] remove ufw role

kayiwa:
we will use sandboxes for our ufw experiments this makes abid staging
stay up


dphillips-39 (approved):

acozine:
Should we also remove the UFW rules from `group_vars` for abid-staging?

PR #5646, opened 2024-12-18
Update Figgy Illiad variables

### LINKED ISSUE #5645: Update Figgy PDF OCR Environment variables
Part of https://github.com/pulibrary/figgy/issues/6412

1. Remove OCR_OUT
2. Update OCR_IN to point to the new Illiad file share: \\lib-fileshare.princeton.edu\ILL_OCR_Scans
3. Add values for Illiad SFTP connection

eliotjordan:
Closes #5645 

[In draft until transition to hosted Illiad]


hackartisan (changes requested):
My one thought is that we could leave the deleted env vars in place, instead introduce a new "in path" var like `OCR_ILLIAD_IN_PATH` or something, and then we could go ahead and provision the figgy boxes to have all the new variables so that the switch over would be just the one deploy. We would probably want a ticket to remove the other vars later, but it would make the switchover a bit smoother. What do you think?

hackartisan (approved):

PR #5649, opened 2024-12-19
[pulfalight] Increase passenger pool

### LINKED ISSUE #5648: increase passenger pool for pulfalight
follow up to #5618

We bumped the pool to 10 but there is still plenty of memory available. Will increase to 20.

hackartisan:
closes #5648


kayiwa (approved):
OK! Thanks @hackartisan

PR #5650, opened 2024-12-19
Update the documentation in nginxplus.yml

### LINKED ISSUE #5603: Nginx playbook does not correctly remove config files
### Expected behavior
When I run the nginxplus.yml playbook with `-e nginx_cleanup_config=true`, Ansible should remove any configs that no longer exist in the config directory. Here's the [comment](https://github.com/pulibrary/princeton_ansible/blob/ffa8f3d96cc83d7498254c43a7434efab9fc7094/playbooks/nginxplus.yml#L5C3-L5C40) that describes this.

### Actual behavior
Today we moved the catalog-staging site to the dev load balancers with #5600. Running the nginxplus.yml playbook against the production load balancers with `-e nginx_cleanup_config=true` should have removed the catalog-staging config, but it did not. Since the SSL certificates had already been removed, this caused `nginx -t` to fail with 

### Error messages
```
nginx: [emerg] cannot load certificate \"/etc/letsencrypt/live/catalog-staging/fullchain.pem\": BIO_new_file() failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/etc/letsencrypt/live/catalog-staging/fullchain.pem','r') error:2006D080:BIO routines:BIO_new_file:no such file)", "nginx: configuration file /etc/nginx/nginx.conf test failed"]
```

### Steps to replicate
Move or remove a configuration file from the nginx role, remove the certificates from the load balancers, run the nginxplus.yml playbook.

### Impact of this bug
We have to manually update the load balancers when we migrate or remove sites.

Beck-Davis:
 
 closes #5603

### DIFF
--- a/playbooks/nginxplus.yml
+++ b/playbooks/nginxplus.yml
@@ -2,8 +2,8 @@
 # you MUST run this playbook on a single host with '--limit' for example `ansible-playbook -v --limit lib-adc2.princeton.edu playbooks/nginxplus.yml`
 # to update the second load-balancer, switch which adc# host is currently active, then run the playbook on the second (formerly active, now inactive) host
 # to update configuration for existing sites, run with '-t update_conf' for example `ansible-playbook -v --limit lib-adc2.princeton.edu playbooks/nginxplus.yml -t update_conf`
-# if you want to remove obsolete config files, you must pass '-e nginx_cleanup_config=true'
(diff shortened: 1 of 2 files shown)


kayiwa (approved):
OK! Thanks @Beck-Davis

PR #5653, opened 2024-12-20
[ABID] Upgrade ruby and node

hackartisan:
refs pulibrary/abid#181


kayiwa (approved):
OK! Thanks @hackartisan

PR #5659, opened 2024-12-23
[fail2ban/nginxplus] Change format of fail2ban match

sandbergja:
Previously, we were trying to match CLF formatted logs in access.log.  But our access log is in a JSON format, so we needed to adjust the fail2ban failregex accordingl[fail2ban/nginxplus] Change format of fail2ban match

Previously, we were trying to match CLF formatted logs in access.log.  But our access log is in a JSON format, so we needed to adjust the fail2ban failregex accordingly


christinach (approved):
Thank you @sandbergja @acozine @jrgriffiniii @kevinreiss @kayiwa @christinach !

PR #5662, opened 2024-12-23
[fail2ban docs] add a note that you can test with fail2ban-regex locally

sandbergja:


kayiwa (approved):
OK! Thanks @sandbergja

PR #5663, opened 2024-12-23
adds two new nomad clients for imagecat

acozine:
Partially addresses https://github.com/pulibrary/ops-catchall/issues/121.

We created two new production nomad VMs today. This adds them to inventory.


sandbergja (approved):
Thanks, @acozine !

PR #5667, opened 2024-12-24
[catalog] Use ruby 3.3.6 with yjit on staging

sandbergja:


christinach (approved):
Thank you @sandbergja !

PR #5674, opened 2024-12-30
[catalog] reduce the number of passenger processes in staging and qa

sandbergja:
Staging and qa have less traffic and less memory than the production boxes, so it is appropriate to have fewer passenger processes.

This helps, but does not totally resolve, an alert in CheckMK that complains about overcommitting memory.  We have determined that this is not actually a problem -- the overcommitted memory does not end up getting used, so we don't run into OOM issues.


christinach (approved):
Thanks @sandbergja !

PR #5675, opened 2024-12-30
[catalog] Use ruby 3.3.6 with YJIT for all environments

sandbergja:
This has been run on catalog5


christinach (approved):
Thanks @sandbergja ! 🍄

PR #5681, opened 2024-12-31
remove the configs for research data

kayiwa:
the service has moved to an OIT hosted solution


Beck-Davis (approved):
Looks good @kayiwa 🧙🏽

PR #5684, opened 2024-12-31
[abid] roll back upgrades and fix passenger settings

hackartisan:
We can't complete the abid language upgrades due to https://github.com/pulibrary/abid/issues/186, so rolling these back while we work on that ticket.

Fixing a few passenger settings noticed in the meantime.

refs pulibrary/abid#181


kayiwa (approved):
OK! Thanks @hackartisan

PR #5685, opened 2024-12-31
[GEACCIRC] Update ruby to 3.3.6

christinach:
https://github.com/pulibrary/geaccirc/issues/57

I've run this against staging and successfully deployed the application. 
I'm running it now against production


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #5689, opened 2025-01-02
[GEACCIRC] Include app_secret_key_base in roles/default.yml so that it's defined for the test

christinach:


sandbergja (approved):
Thanks, @christinach !  🐬

PR #5693, opened 2025-01-06
[ruby_s] Run make in parallel

sandbergja:
I installed a new ruby version (3.3.6 with YJIT) on both dss-staging1 and dss-staging2. I used this commit for dss-staging2, but not dss-staging1 for the sake of comparison. Running the playbook took:
  * 15:53.25 on dss-staging1 without this commit
  * 9:55.20 on dss-staging2 with this commit


maxkadel (approved):
Thanks so much @sandbergja! 🍄

PR #5694, opened 2025-01-06
Updated the copy module to include a notify to restart fail2ban service.

aruiz1789:
Updated the copy module to include a "notify" to restart fail2ban service when changing the fail2ban config files.

The dev environments had a pending fail2ban config change at: /etc/fail2ban/filter.d/nginx-badbots.conf

Which trigged an automatic restart of the fail2ban service when running the playbook: active (running) since Mon 2025-01-06 15:22:06 UTC; 41s ago


acozine:
Looks good, thanks @aruiz1789!

PR #5696, opened 2025-01-06
Add secret key base to support the Rails 7.2 upgrade

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #5703, opened 2025-01-07
[nginx] exclude apt cache from idempotence check

sandbergja:
This step can be a bit flaky in an idempotence check.  For example: https://github.com/pulibrary/princeton_ansible/actions/runs/12656596039/job/35269514546

My guess is that some package version was updated between when `molecule converge` ran and `molecule idempotence` ran, which led to this step being "changed".

We can exclude a step from running during the idempotence checks with the molecule-idempotence-notest tag, hopefully this helps!


maxkadel (approved):
So good to know! Cool flag!

PR #5704, opened 2025-01-08
[nginxplus] add new virtual loadbalancer

kayiwa:
remove lib-adc2 from the allow list to allow traffic from it

Co-authored-by: Angel Ruiz <aruiz1789@users.noreply.github.com>


maxkadel (approved):
Thanks @kayiwa!

PR #5705, opened 2025-01-08
Update mailcatcher path to match new Ruby version, use newer version …

maxkadel:
…of mailcatcher


christinach (approved):
Thanks @maxkadel 🦄

PR #5712, opened 2025-01-08
[orangelight] Add sidekiq

maxkadel:
Connected to https://github.com/pulibrary/orangelight/issues/4609

This adds the sidekiq role to the orangelight playbook.  A few things to note:

- Some role dependencies are moved from the orangelight role to the orangelight playbook.  This is because, when they were in the role, time-consuming tasks like the `ruby_s` tasks would run twice: first for the sidekiq boxes, then again for non-sidekiq boxes.  Now that they are in the playbook, placed before the `sidekiq` role, `ruby_s` (and other tasks like `bind9`) only run once.
- The `sidekiq_worker_name` does not contain the name of the application, as it does for bibdata.  This is so we can simply run `systemctl restart sidekiq`, rather than having to figure out the application-specific sidekiq name.

### DIFF
--- a/roles/deploy_user/tasks/main.yml
+++ b/roles/deploy_user/tasks/main.yml
@@ -22,13 +22,6 @@
   run_once: true
   tags: update_keys
 
-- name: DEBUG DON'T COMMIT ME - inventory
-  ansible.builtin.debug:
-    var: hostvars[inventory_hostname]
-
-- name: DEBUG DON'T COMMIT ME - keys
-  ansible.builtin.debug:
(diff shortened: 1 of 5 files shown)


roles/deploy_user/tasks/main.yml
```diff
@@ -22,6 +22,7 @@
   run_once: true
   tags: update_keys
```
kayiwa, line 23: cool!

kayiwa (approved):
OK! Thanks @maxkadel

PR #5714, opened 2025-01-09
Decommission Cicognara files_retake

dphillips-39:
Deleted Cicognara 1 & 2 from inventory and playbooks/cicognara.yml


kayiwa (approved):
OK. Thanks @dphillips-39

PR #5715, opened 2025-01-09
[lib-sftp] make /alma/aspace directory group-writeable

sandbergja:
Two users write to the /alma/aspace directory: the alma user and the aspace user.  They are both part of the same group, but the group needs write permissions to the directory.


kayiwa (approved):
OK! Thanks @sandbergja

PR #5719, opened 2025-01-09
[orangelight] Update mailcatcher paths

sandbergja:
Also, add two default variables to the ruby_s role to automatically update the mailcatcher path when the ruby_version_override variable changes. These default variables could be overriden in the group_vars if needed.

I ran this on staging and qa, and confirmed that the systemd service is now looking at the new path.


maxkadel (approved):
Thanks @sandbergja !

PR #5726, opened 2025-01-10
[EZProxy] I4788 ezproxy test

VickieKarasic:


group_vars/ezproxy/common.yml
```diff
@@ -0,0 +1,6 @@
+---
+shib_admin_user:
```
acozine, line 2: where is this var being used/consumed?

roles/ezproxy/tasks/main.yml
```diff
@@ -52,6 +34,13 @@
   when:
     - running_on_server
 
+- name: Ezproxy | install ezproxy startup script (requires root)
+  ansible.builtin.command: /var/local/ezproxy/ezproxy -si
```
acozine, line 38: Does this command create a file or install a service or something like? something we could use as a condition to make the command only run when it needs to? Then we wouldn't need the `changed_when: false` . . . or at the least we could set `changed_when:` to something more specific

roles/ezproxy/templates/princeton_allow.txt.j2
```diff
@@ -0,0 +1,10 @@
+# {{ ansible_managed | comment }}
```
acozine, line 1: Is this the only reason to make this a template?

roles/ezproxy/tasks/main.yml
```diff
@@ -174,47 +142,53 @@
...
 - name: Ezproxy | add shib config user
   ansible.builtin.template:
     src: "shibuser.txt.j2"
     dest: /var/local/ezproxy/shibuser.txt
-    mode: 0644
+    mode: "0644"
     owner: root
     group: root
 
+- name: Ezproxy | add shib config user
```
acozine, line 153: I don't really understand what this task does from the name. Looking at the template, it looks to be adding IPs or IP ranges, not users . . . was this copy pasta?

roles/ezproxy/tasks/main.yml
```diff
@@ -174,47 +142,53 @@
...
   changed_when: false
   when:
     - running_on_server
+
+- name: Ezproxy | reload systemd to recognize new service
+  ansible.builtin.systemd:
+    daemon_reload: true
+  when:
+    - running_on_server
   notify: reload ezproxy
```
acozine, line 177: In this task we are reloading systemd as a whole and if this task returns `changed` we will run a handler to reload ezproxy specifically . . . but a task or two below this one, we are also restarting ezproxy. Do we still need to notify the reload task here?

acozine (commented):
Thanks for updating the EZproxy automation so we can rebuild our EZproxy servers! I had a couple of questions/suggestions, see below.

group_vars/ezproxy/common.yml
kayiwa: https://github.com/pulibrary/princeton_ansible/blob/444f7194df09861075e1c80376e79c53dabdac4f/roles/ezproxy/templates/shibuser.txt.j2#L16

roles/ezproxy/tasks/main.yml
kayiwa: we get very little to go for what is does from [the documentation](https://help.oclc.org/Library_Management/EZproxy/Install_and_update_EZproxy/Install_and_update_self-hosted_EZproxy/EZproxy_for_Linux_install) :shrug:

VickieKarasic (commented):
Thanks for this @kayiwa

PR #5729, opened 2025-01-10
[orangelight] Ruby 3.4.1

sandbergja:
I ran this on catalog-staging and catalog-qa2

### DIFF
--- a/roles/orangelight/tasks/main.yml
+++ b/roles/orangelight/tasks/main.yml
@@ -2,21 +2,21 @@
 # tasks file for roles/orangelight
 
 - name: orangelight | update rubygems
-  ansible.builtin.command: gem update --system 3.5.10
+  ansible.builtin.command: gem update --system 3.6.1
   become: true
   register: update_rubygems
   changed_when: '"Latest version already installed. Done." not in update_rubygems.stdout'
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja !

roles/orangelight/tasks/main.yml
```diff
@@ -2,17 +2,17 @@
...
 - name: orangelight | install default bundler
-  ansible.builtin.command: gem install --default bundler:2.5.10
+  ansible.builtin.command: gem install --default bundler:2.6.1
   become: true
   register: install_default_bundler
-  changed_when: '"Fetching bundler-2.5.10.gem" in install_default_bundler.stdout'
+  changed_when: '"Fetching bundler-2.6.1.gem" in install_default_bundler.stdout'
   tags: rubygems
 
 - name: orangelight | remove old default bundler
```
christinach, line 18: @sandbergja I think we need to remove the default bundler 2.5.10

sandbergja: Ah!  Thanks for catching that.  I force pushed.

christinach (approved):

PR #5737, opened 2025-01-14
[check_mk] sftp check

### LINKED ISSUE #5728: Create a checkmk local check to validate an SFTP connection

Create a [local check](https://docs.checkmk.com/latest/en/localchecks.html) for the bibdata-staging and lib-jobs-staging servers.  The check should make sure that it can successfully connect via SFTP using the host and credentials in the `SFTP_PASSWORD`, `SFTP_USERNAME`, `SFTP_HOST`.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

This would allow us to know about any potential connection issues early, before they interfere with business-critical processes.  These connection issues could be anything from a network problem to an incorrect credential to a typo in an environment variable to an outdated SSH fingerprint.

### Acceptance criteria
 - [ ] Checkmk says "OK" when the connection is good
 - [ ] Checkmk says "WARN" or "CRIT" when the connection is bad (for example, we use the wrong username or password)

* See our [existing local checks for inspiration](https://github.com/pulibrary/princeton_ansible/tree/2faf25f5bf11475da1af4d9745a13ae3ee55d0c7/group_vars/checkmk/local_check_templates)
* In case it is helpful, here is a ruby one-liner that can check if an sftp connection is good with the provided environment variables (would need to `gem install net-sftp` first):
```
ruby -rnet/sftp -e "puts Net::SFTP.start(ENV['SFTP_HOST'], ENV['SFTP_USERNAME'], :password => ENV['SFTP_PASSWORD']) {puts 'connection is good'}"
```

kayiwa:
most of our authentication uses keybased auth. we have services that use active directory service for authentication. This check will allow periodic checks to ensure password authentication works.

related #5728


group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
```diff
@@ -0,0 +1,26 @@
+{{ ansible_managed | comment }}
+#!/bin/bash
+
+# Exit codes
+EXIT_OK=0
+EXIT_WARN=1
+EXIT_CRIT=2
+EXIT_UNKNOWN=3
```
sandbergja, line 8: I love having these codes as variables here!

group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
```diff
@@ -0,0 +1,26 @@
...
+
+# Check if all credentials are defined
+if [[ -z "{{ sftp_host }}" || -z "{{ sftp_user }}" || -z "{{ sftp_password }}" ]]; then
```
sandbergja, line 11: How will we pass in these ansible variables?

group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
```diff
@@ -0,0 +1,26 @@
...
+  echo "Critical: SFTP credentials (host, username, password) are not defined!"
+  exit $EXIT_CRIT
+fi
+
+# Attempt SFTP connection
+sftp -oBatchMode=yes -oStrictHostKeyChecking=no "{{ sftp_user }}"@"{{ sftp_host }}" <<< "{{ sftp_password }}" 2>&1
```
sandbergja, line 17: 👍🏻

group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
```diff
@@ -0,0 +1,26 @@
...
+
+# Check exit code of sftp command
+if [[ $? -eq 0 ]]; then
+  echo "OK: Successful connection to SFTP server {{ sftp_host }}"
+  exit $EXIT_OK
+else
+  echo "Critical: Connection to SFTP server {{ sftp_host }} failed!"
```
sandbergja, line 24: Same for this one, I think checkmk wants the output like:
```suggestion
  echo "$EXIT_CRIT \"SFTP\" - Connection to SFTP server {{ sftp_host }} failed!"
```

sandbergja (changes requested):
Thanks so much, @kayiwa !  This looks so nice!  I had one question about how we will use this, and then some format suggestions based on what I understood of the checkmk documentation.

group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
kayiwa: 🤦🏿‍♂️  I wrote the shell script before coffee and then yaml-ized it.

kayiwa: I used perl. This is a pattern worth further discussion. The previous templates assume we would use what is on the OS but this may possibly not scale.

sandbergja (commented):
It looks like checkmk wants a [really specific output](https://docs.checkmk.com/latest/en/localchecks.html) to be able to show up in service discovery.  Other than that, I love this PR!  Thanks so much @kayiwa !

group_vars/checkmk/local_check_templates/sftp_with_pass_check.j2
kayiwa: will bring this up at Ansible Open House possibly

sandbergja (approved):
Thanks, @kayiwa and @VickieKarasic !

PR #5746, opened 2025-01-15
removed file lib-sc-staging.conf

dphillips-39:
Removed file lib-sc-staging.conf


kayiwa (approved):
OK. Thanks @dphillips-39

PR #5749, opened 2025-01-15
[fail2ban] Use <ADDR> rather than <HOST> when we only need an IP address

sandbergja:
The [fail2ban documentation](https://github.com/fail2ban/fail2ban/wiki/Best-practice) says that if we don't need DNS resolution, we should use `<ADDR>` instead of `<HOST>`.

I confirmed using `fail2ban-regex` that it produces the same results on a sample of 10,000 log entries, and that it is consistently a little faster.


christinach (approved):
Thanks @sandbergja 🍄

PR #5754, opened 2025-01-16
[check_mk] sftp check

kayiwa:
our initial install had used cpanimus but and we didn't rememeber this
until the second vm was added

related to #5728

### DIFF
--- a/playbooks/utils/checkmk_add_sftp_check.yml
+++ b/playbooks/utils/checkmk_add_sftp_check.yml
@@ -9,15 +9,20 @@
     - ../group_vars/sftp/{{ runtime_env | default('staging') }}.yml
 
   tasks:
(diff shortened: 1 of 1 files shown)


playbooks/utils/checkmk_add_sftp_check.yml
```diff
@@ -9,15 +9,20 @@
...
   tasks:
-    - name: Install required Perl modules
+    - name: Set up Perl modules
       ansible.builtin.apt:
         name: "{{ item }}"
         state: present
         update_cache: true
       loop:
-        - libnet-sftp-foreign-perl
+        - cpanminus
```
sandbergja, line 18: 👍🏻

sandbergja (approved):
Thanks, @kayiwa !

PR #5755, opened 2025-01-16
Remove ojs-prod1.princeton.edu from the load balancer

christinach:
The VM doesn't exist anymore and this caused 'nginx -t' to fail


sandbergja (approved):
Thanks, @christinach @sandbergja ! 🐬 🍄

PR #5756, opened 2025-01-16
Fix syntax errors in fail2ban configuration

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬 🍄

PR #5760, opened 2025-01-17
[fail2ban] Reject requests with 15 facets, rather than 20

sandbergja:
@christinach and I found that the current fail2ban configuration is stressing our load balancers -- maxing out 2 CPUs and it is having trouble keeping up with the volume.

While testing locally with fail2ban-regex, looking for 15 facets applied is more than 20 times faster than looking for 20 facets. This reduces the number of facets accordingly, while relaxing the number of such requests that an IP needs to make before being banned.


christinach (approved):
Thanks @sandbergja 🍄 !

kayiwa:
Bye 🤖

Issue #5764, opened 2025-01-18
Nginx Configuration Warnings: Deprecated "listen ... http2" directive and


kayiwa:
This ticket documents multiple warnings encountered while running sudo nginx -t to test the Nginx configuration.

**Warnings:**

- Deprecated "listen ... http2" directive: The warnings indicate that several configuration files are using the deprecated listen ... http2 directive. The recommended replacement is the http2 directive. This should be updated across all affected configurations.
  - List of affected files:
    - /etc/nginx/conf.d/figgy-prod.conf (line 27)
    - /etc/nginx/conf.d/figgy-staging.conf (line 25)
    - /etc/nginx/conf.d/fpul-prod.conf (line 23)
    - ... (list continues - please see full output for details)
- Protocol options redefined: The warnings also mention multiple instances of redefined protocol options for 0.0.0.0:443. This could potentially lead to unexpected behavior. Further investigation is needed to identify the root cause and implement a proper fix.
  - List of affected files (where applicable):
    - /etc/nginx/conf.d/figgy-staging.conf (line 25)
    - ... (list continues - please see full output for details)
- server_names_hash warnings: The configuration test mentions warnings related to server_names_hash. The output suggests that increasing either server_names_hash_max_size or server_names_hash_bucket_size might be necessary.
- Conflicting server name: The test identified a conflicting server name "cdh-test-derrida.princeton.edu" on 0.0.0.0:443. This needs to be addressed to avoid potential issues.
**- Proposed Resolution:**

1. Update all occurrences of the deprecated listen ... http2 directive to the recommended http2 directive in the affected configuration files.
2. Investigate the cause of redefined protocol options for 0.0.0.0:443 and implement a solution to ensure proper configuration.
3. Increase server_names_hash_max_size or server_names_hash_bucket_size in the Nginx configuration based on the recommendation from the test output.
4. Resolve the conflicting server name "cdh-test-derrida.princeton.edu" on 0.0.0.0:443.

- Additional Notes:
-  swap out all configs that have ` listen 443 ssl http2; ` to have `listen 443 ssl;`
- investigate server_hash problem

The complete output of sudo nginx -t
```bash
nginx: [warn] the "listen ... http2" directive is deprecated, use the "http2" directive instead in /etc/nginx/conf.d/figgy-prod.conf:27
nginx: [warn] the "listen ... http2" directive is deprecated, use the "http2" directive instead in /etc/nginx/conf.d/figgy-staging.conf:25
[...]

PR #5768, opened 2025-01-21
[cdh] ssl error

kayiwa:
the cdh-test-derrida was giving us a TLS error. This change removes it

related to #5764


Beck-Davis (approved):
Thank you @kayiwa

Issue #4788, opened 2024-03-25
EZproxy Test Can't Use Shibboleth


kevinreiss:
Shibboleth doesn't work on EZproxy test, we need to review the set-up and prod and replicate it on test. There is an entry in the campus IDP for the ezproxy-test.princeton.edu domain, but we may need to reach out to the IAM group to update something. 

Shibboleth settings are the /var/local/ezproxy/shibuser.txt and in /var/local/ezproxy/config.txt.

VickieKarasic:
- We renewed the cert on ezproxy-test.princeton.edu since it had expired in April 2024. 
- We also put in a ticket with OIT's IAM group to add a new service provider for Shibboleth - see [CHG0105200](https://princeton.service-now.com/now/nav/ui/classic/params/target/change_request.do%3Fsys_id%3Dcfa3949847f156107d2ae1dc416d43cb%26sysparm_stack%3D%26sysparm_view%3D)
- We'll move forward on this once we hear back from OIT

VickieKarasic:
I met with Jennifer Tuorto regarding [INC0634311](https://princeton.service-now.com/now/nav/ui/search/0f8b85d0c7922010099a308dc7c2606a/params/search-term/INC0634311/global-search-data-config-id/c861cea2c7022010099a308dc7c26041/back-button-label/Tasks/search-context/now%2Fnav%2Fui). OIT has updated the certificate on their end for all three iterations of our test site: 

```
ezproxy-test.princeton.edu
http://ezproxy-test.princeton.edu
https://ezproxy-test.princeton.edu
```

It looks Failing at the return site on the SAML trace that Jennifer provided: 
```
[...]

PR #5779, opened 2025-01-24
[ezproxy] update ezproxy-test to latest version

VickieKarasic:
upgrade version to 7.2.12 to match ezproxy-prod

update group_vars with new domain name


acozine:
Related to #4788.

PR #5785, opened 2025-01-27
[checkmk] update password

kayiwa:
modify the troubleshooting user password

we need to update the password


kayiwa:
I thought I accidentally leaked the troubleshooting user password on an checkmk support ticket.

PR #5801, opened 2025-01-30
[postgresql] remove enable postgresql

kayiwa:
the enable postgresql task reports a failure even though the service is
    still running. We comment it out and the role runs successfully

Co-authored-by: Beck Davis <beck-davis@users.noreply.github.com>


christinach (approved):
Thanks @kayiwa and @Beck-Davis

PR #5802, opened 2025-01-30
Update bundler 2.6.3

christinach:


sandbergja (approved):
Thanks @christinach ! 🐬

PR #5806, opened 2025-01-31
[pulmonitor] add new server

kayiwa:
we need to distribute the load to allow us to scale. We add a new remote
location host that will host non-production vms

create a second site for none production servers


VickieKarasic (approved):
This looks good @kayiwa - thank you!

Issue #5753, opened 2025-01-16
Add allowPaths to solr 9 configuration to fix backups


hackartisan:
If you run a new backup (look at the crontab and copy the backup command, run it as deploy user) and then consult `/tmp/solr_backup.log` you will see request ids and can use those to check the status of the backup like so:

`curl "http://localhost:8983/solr/admin/collections?action=REQUESTSTATUS&requestid=geniza-202501161456"`

Currently we get the following status:

```
{
  "responseHeader":{
    "status":0,
    "QTime":4},
  "success":{
    "lib-solr-staging1.princeton.edu:8983_solr":{
      "responseHeader":{
        "status":0,
        "QTime":0}}},
  "geniza-2025011614562084550127003914":{
    "responseHeader":{
      "status":0,
      "QTime":0},
    "STATUS":"failed",
    "msg":"Failed to backup core=geniza_shard1_replica_n1 because org.apache.solr.common.SolrException: Path /mnt/solr_backup/solr9/staging/20250116/geniza-20250116.bk/geniza must be relative to SOLR_HOME, SOLR_DATA_HOME coreRootDirectory. Set system property 'solr.allowPaths' to add other allowed paths."},
  "failure":{
    "lib-solr-staging1.princeton.edu:8983_solr":{
      "responseHeader":{
        "status":0,
        "QTime":0},
      "STATUS":"failed",
      "msg":"Failed to backup core=geniza_shard1_replica_n1 because org.apache.solr.common.SolrException: Path /mnt/solr_backup/solr9/staging/20250116/geniza-20250116.bk/geniza must be relative to SOLR_HOME, SOLR_DATA_HOME coreRootDirectory. Set system property 'solr.allowPaths' to add other allowed paths."}},
  "Operation backup caused exception:":"java.nio.file.NoSuchFileException:java.nio.file.NoSuchFileException: /mnt/solr_backup/solr9/staging/20250116/geniza-20250116.bk/geniza/zk_backup_0",
  "exception":{
    "msg":"/mnt/solr_backup/solr9/staging/20250116/geniza-20250116.bk/geniza/zk_backup_0",
    "rspCode":-1},
  "status":{
    "state":"failed",
    "msg":"found [geniza-202501161456] in failed tasks"}}
```

It seems like there's something we could do in solr9 playbook to make this work.

See more at https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html

### Actual behavior
[...]

PR #5807, opened 2025-02-01
[solrcloud] solr9 backups

kayiwa:
reintroduce behavior of solr pre 8.6 which allowed arbitrary paths. This will enable backups to continue. 

related to #5753

### DIFF
--- a/roles/solrcloud/tasks/config.yml
+++ b/roles/solrcloud/tasks/config.yml
@@ -1,5 +1,5 @@
 ---
-- name: solrcloud | Check for solr znode
+- name: Solrcloud | Check for solr znode
   ansible.builtin.command: >
     {{ solr_installation }}/bin/solr zk ls /
     -z {{ solr_zookeeper_hosts_string }}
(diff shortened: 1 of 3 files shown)


roles/solrcloud/tasks/config.yml
```diff
@@ -1,65 +1,65 @@
...
+    mode: "0644"
     force: true
   notify: restart SolrCloud
 
-- name: solrcloud | Configure pre 7.4 logging
+- name: Solrcloud | Configure SolrCloud 9 properties
   ansible.builtin.template:
-    src: 'log4j.properties.j2'
-    dest: '/solr/log4j.properties'
+    src: 'solr9.xml.j2'
```
hackartisan, line 48: should this only happen if you're installing solr 9? Or does this role only do solr 9 at this point?

roles/solrcloud/tasks/config.yml
```diff
@@ -1,65 +1,65 @@
...
-    mode: 0644
+    mode: "0644"
     force: true
   notify: restart SolrCloud
 
-- name: solrcloud | Configure SolrCloud properties
+- name: Solrcloud | Configure SolrCloud properties
   ansible.builtin.template:
     src: 'solr.xml.j2'
     dest: '{{ solr_data_dir }}/solr.xml'
```
hackartisan, line 39: Can the new keys / values not be added into this solr.xml?

hackartisan (changes requested):
Thanks for working on this! I'm a little confused about why there's a new config file instead of updating the solr.xml template that's in there. And also how / whether these tasks should run differently for different versions of solr

Also can you please add a ref to the issue in the PR description?

roles/solrcloud/tasks/config.yml
kayiwa: no we do NOT!

hackartisan (approved):

Issue #5651, opened 2024-12-19
Decommission library-migrated.lib.princeton.edu


VickieKarasic:
### What maintenance needs to be done?

Decommission library-migrated.lib.princeton.edu, which is the following three servers: 

- library-prod1.lib.princeton.edu
- library-prod3.lib.princeton.edu
- library-prod4.lib.princeton.edu

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Relevant deadlines
If this maintenance relates to an End of Life date or other deadline, please enter it here.
- [ ] Would like these to be decommissioned by 01/31/2025, which is what was communicated to staff members

### Why is this maintenance needed?
These vms are no longer needed, since we no longer host and maintain the drupal library website. Archival versions of the site can still be found in both [Archive-It](https://wayback.archive-it.org/5151/*/https://library.princeton.edu/) and the [Wayback Machine](https://web.archive.org/web/20240000000000*/https://library.princeton.edu/).

### Acceptance criteria
 - [ ] Delete load balancer configurations
   - [x] [Library Prod](https://github.com/pulibrary/princeton_ansible/blob/595d953ae0679ae94d20eb1e83c2a4949d1ffaca/roles/nginxplus/files/conf/http/library-prod.conf)
   - [x] Delete TLS certs at `/etc/letsencrypt/archive/library-migrated.lib/`
 - [x] Delete VMs in the inventory
   - [x] inventory/all_projects/libwww
 - [x] delete library-prod1.lib.princeton.edu vm
 - [x] delete library-prod3.lib.princeton.edu vm
 - [x] delete library-prod4.lib.princeton.edu vm
 - [x] delete DNS for the VMs
 - [x] delete firewall rules

### Implementation notes, if any

- [x] Network disconnect these on 1/2/25
- [x] Power VMs off on 1/15/25
- [ ] Fully decommission by 1/31/25

VickieKarasic:
library-prod and library-prod2 DNS records in Network Registration

VickieKarasic:
Closed by #5804

acozine:
Hm, for some reason, Vickie's comment above didn't close this. Today we revoked the certificate for `library-migrated.lib.princeton.edu`. Closing.

PR #5809, opened 2025-02-03
removed references to deleted undefined hosts

Beck-Davis:
references to libwww_production in inventory/by_team/dacs and inventory/by_service/apache were causing errors when running playbooks

related to #5651


kayiwa (approved):
OK! thanks @Beck-Davis

PR #5819, opened 2025-02-05
[solr9cloud] templatize solr.xml

kayiwa:
modify our solr.xml config template. This allows us to be able to have a
conditional loop that accomodates the breaking changes from solr >=
8.6.0

related to #5753

Co-authored-by: Angel Ruiz <aruiz1789@users.noreply.github.com>
Co-authored-by: Denzil Phillips <dphillips-39@users.noreply.github.com>


roles/solrcloud/templates/solr.xml.j2
```diff
@@ -28,6 +28,17 @@
 
 <solr>
 
+  {% if solr_cloud_download_version >= "8.6.0" %}
```
hackartisan, line 31: This looks like it's going to compare lexically, so 8.12.0 would not be greater than 8.6.0. Am I reading that right? Maybe the solr 8 line will never release anything 10 or higher and it won't be an issue?

kayiwa: you're right. I am not keen on writing a function (which seems the only solution otherwise).

kayiwa: https://www.shellhacks.com/ansible-compare-version-numbers-examples/ and https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html

possible 

```
when: solr_version | version_compare('8.6', '>')
```

acozine:
Closing in favor of #5831.

PR #5824, opened 2025-02-07
add a remote variable

kayiwa:
this allows us to use the remote variable


kayiwa:
set the `checkmk_agent_site` to the `runtime_env`

kayiwa:
use staging for staging.

PR #5830, opened 2025-02-08
this is how the solr9 prod boxes are set up now, change was done by hand

acozine:
WIP.

I was experimenting with the three Solr9 production boxes. This branch reflects the current setup, but I'm not sure it's correct yet.


acozine:
This was just a way to remember what we did manually. Actual work has been integrated into other PRs.

Issue #5740, opened 2025-01-14
[CheckMK] Monitor Solr backups


acozine:
### User story
As a member of the IT team, I want to be sure our Solr backups are happening. I definitely want to know right away if a Solr backup fails.

### Acceptance criteria

- [x] We have a check in CheckMK that monitors the creation of Solr backups
- [x] If we prevent a backup from happening, we get an alert

### Concrete example
Related to https://github.com/pulibrary/pul_solr/issues/447.

### Implementation notes, if any
We have some documentation about our Solr backups in the [README](https://github.com/pulibrary/pul_solr/blob/main/README.md#solrcloud-backups) for the pul_solr repo.

hackartisan:
correllary:  https://github.com/pulibrary/pul_solr/issues/409

PR #5831, opened 2025-02-09
I5740 solr backup

kayiwa:
the solr9 cloud backup is failing. Recent PR's suggest that solr9 cloud and zookeeper have not been working in ways we expect.

Related #5740


roles/solrcloud/README.md
```diff
@@ -1,3 +1,145 @@
...
+- Automatic startup: Enabled
+- Network: Listens on all interfaces (0.0.0.0)
+
+## Important Paths
+
+- Home directory: `/solr`
+- Data directory: `/solr/data`
+- Installation directory: `/opt/solr`
+- PID directory: `/var/run/solr`
+- Log directory: `/var/log/solr`
```
acozine, line 112: This is fantastic information, but I wonder if it would serve us better in documentation for Solr/Zookeeper, rather than in the role README. If I need to find the solr data directory for troubleshooting something, I wouldn't think to look here first.

roles/solrcloud/defaults/main.yml
```diff
@@ -1,48 +1,37 @@
...
-solr_group: deploy
-
-# start on boot
-solr_service_enabled: true
-solr_service_state: started
-
+# defaults file for roles/solrcloud
+solr_cloud_download_version: 8.4.0
+solr_cloud_version: "{{ solr_version }}"
+solr_version: "{{ solr_cloud_download_version }}"
```
acozine, line 4: If we can consolidate the variables, we should do it here too.

roles/solrcloud/defaults/main.yml
```diff
@@ -1,48 +1,37 @@
...
-solr_gc_tune: "-XX:NewRatio=3 \
-  -XX:SurvivorRatio=4 \
-  -XX:TargetSurvivorRatio=90 \
-  -XX:MaxTenuringThreshold=8 \
-  -XX:ConcGCThreads=4 \
-  -XX:ParallelGCThreads=4 \
-  -XX:PretenureSizeThreshold=64m \
-  -XX:+ParallelRefProcEnabled"
-
+solr_gc_tune: "-XX:NewRatio=3 -XX:SurvivorRatio=4 -XX:TargetSurvivorRatio=90 -XX:MaxTenuringThreshold=8 -XX:ConcGCThreads=4 -XX:ParallelGCThreads=4 -XX:PretenureSizeThreshold=64m -XX:+ParallelRefProcEnabled"
```
acozine, line 30: I find the old presentation, with line breaks, makes it easier to find what we're setting. We can debate.

roles/solrcloud/defaults/main.yml
```diff
@@ -1,48 +1,37 @@
...
 # Enable verbose GC logging
 # these are the Java 9 settings
 # GCID and GCCause are always logged
-gc_log_settings: "-verbose:gc \
-  -Xlog:gc=debug:file=/solr/logs/gc.txt:time,uptime,pid,tid,level,tags \
-  -Xlog:gc+heap=debug \
-  -Xlog:gc+age=trace \
-  -Xlog:safepoint"
-
+gc_log_settings: "-verbose:gc -Xlog:gc=debug:file=/solr/logs/gc.txt:time,uptime,pid,tid,level,tags -Xlog:gc+heap=debug -Xlog:gc+age=trace -Xlog:safepoint"
```
acozine, line 34: Same here. Were the linebreaks causing issues? Or do folks just prefer not to include them?

roles/solrcloud/meta/main.yml
```diff
@@ -12,8 +12,5 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - bionic
-dependencies:
-  - role: 'common'
-  - role: 'deploy_user'
-  - role: 'zookeeper'
```
acozine, line 19: Do we have these in the molecule converge playbook?

roles/solrcloud/molecule/default/converge.yml
```diff
@@ -5,7 +5,7 @@
     - running_on_server: false
   become: true
   pre_tasks:
-    - name: update cache
+    - name: Update cache
       ansible.builtin.apt:
         update_cache: true
         cache_valid_time: 600
```
acozine, line 11: Since we no longer have any role dependencies, if we want other roles to run in testing, we need to add them below, I think.

roles/solrcloud/molecule/default/verify.yml
```diff
@@ -3,12 +3,18 @@
...
+        name: "{{ item }}"
+        state: present
+      check_mode: true
+      register: pkg_status
+      loop:
+        - ruby-switch
+        - ruby2.6-dev
+        - ruby2.6
+
+    - name: Test for ruby packages
```
acozine, line 17: Hm, this seems minimalist - shouldn't we verify more?

roles/solrcloud/tasks/config.yml
```diff
@@ -1,71 +1,58 @@
 ---
+
 - name: Solrcloud | Check for solr znode
   ansible.builtin.command: >
     {{ solr_installation }}/bin/solr zk ls /
-    -z {{ solr_zookeeper_hosts_string }}
+    -z {{ zookeeper_hosts | join(',') }}
```
acozine, line 6: If we consolidate the zk variables, make sure this line remains correct.

roles/solrcloud/tasks/config.yml
```diff
@@ -1,71 +1,58 @@
...
   register: znode_check
   ignore_errors: true
   changed_when: false
 
 - name: Solrcloud | Create zk chroot / solr znode using bin/solr
   ansible.builtin.command: >
     {{ solr_installation }}/bin/solr zk mkroot {{ solr_znode }}
-    -z {{ solr_zookeeper_hosts_string }}
+    -z {{ zookeeper_hosts | join(',') }}
```
acozine, line 14: Here too.

roles/solrcloud/tasks/config.yml
```diff
@@ -1,71 +1,58 @@
...
     group: '{{ solr_group }}'
     mode: "0644"
     force: true
-  notify: restart SolrCloud
-  when:
-    - '"9" in solr_cloud_download_version'
+  notify: Restart solr
 
-- name: solrcloud | Configure pre 7.4 logging
+- name: Solrcloud | Configure pre 7.4 logging
```
acozine, line 47: It's odd that this task name references 7.4 - is that a Java version? a Solr version? Have we moved beyond the need for this?

roles/solrcloud/tasks/install.yml
```diff
@@ -1,54 +1,38 @@
...
-- name: solrcloud | create solr dirs
-  ansible.builtin.file:
-    path: '{{ item }}'
-    state: directory
-    owner: '{{ solr_user }}'
-    group: '{{ solr_group }}'
-    mode: 0755
-  loop:
-    - '{{ solr_log_dir }}'
-    - '{{ solr_data_dir }}'
```
acozine, line 13: Is this the only place we used these vars? If so, then we don't need them and can remove them from defaults and group_vars.

roles/solrcloud/tasks/install.yml
```diff
@@ -1,54 +1,38 @@
...
     dest: '/tmp/{{ solr_cloud_package }}'
     timeout: 300
+    mode: "0644"
   register: solr_downloaded
-  when:
-    service_solrcloud_status is failed
-  retries: 3
-  delay: 10
   until: solr_downloaded is succeeded
   ignore_errors: true
```
acozine, line 18: hm, this seems risky - what happens when the URL changes? I'd think we want to know if this step fails.

roles/solrcloud/tasks/install.yml
```diff
@@ -1,54 +1,38 @@
...
 
-- name: solrcloud | copy local copy if download fails
+- name: Solrcloud | copy local copy if download fails
```
acozine, line 20: ah, I see, we have a fallback . . . hmmmmmmm

roles/solrcloud/tasks/main.yml
```diff
@@ -1,78 +1,84 @@
...
-        address: "{{ lib_zk2_host }}",
-      }
-    - {
-        hostname: "{{ lib_zk3_host_name }}.princeton.edu",
-        address: "{{ lib_zk3_host }}",
-      }
-  tags:
-    - hosts
-  when: lib_zk1_host_name != "localhost" and lib_zk2_host_name != "localhost" and lib_zk3_host_name != "localhost"
+#- name: Solrcloud | Create mount to diglibdata shares
```
acozine, line 28: Why are we not doing this?

roles/solrcloud/tasks/main.yml
```diff
@@ -1,78 +1,84 @@
...
+- name: Solrcloud | Create Solr directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    owner: "{{ solr_user }}"
+    group: "{{ solr_group }}"
+    mode: "0755"
+  with_items:
+    - "{{ solr_home }}"
+    - "{{ solr_data_dir }}"
```
acozine, line 51: ah, we ARE still using these vars, they're just later in the process . . .

roles/solrcloud/tasks/main.yml
```diff
@@ -1,78 +1,84 @@
...
     - install
 
-- name: solrcloud | configure solr cloud
+- name: Solrcloud | Configure Solr systemd service
+  ansible.builtin.template:
+    src: solr.service.j2
+    dest: /etc/systemd/system/solr.service
+    mode: "0644"
+
+- name: Solrcloud | Configure Solr
```
acozine, line 65: We already do this in the config.yml tasks. Why do it again here?

roles/solrcloud/tasks/main.yml
```diff
@@ -1,78 +1,84 @@
...
+  notify: Restart solr
+
+- name: Solrcloud | Start and enable Solr service
+  ansible.builtin.systemd:
+    name: solr
+    state: started
+    enabled: true
+    daemon_reload: true
+
+- name: Solrcloud | configure solr cloud
```
acozine, line 81: do we want to do configuration tasks before we start and enable the service/

roles/solrcloud/templates/solr.in.sh.j2
```diff
@@ -1,51 +1,40 @@
...
+# By default the start script uses "localhost"; override the hostname here
+SOLR_HOST="{{ inventory_hostname }}"
+# Location where Solr should write logs to; should agree with the file appender
+# settings in server/resources/log4j.properties
+SOLR_LOGS_DIR="{{ solr_log_dir }}"
 
-# Increase Java Heap as needed to support your indexing / query needs
-SOLR_HEAP='{{ solr_heap }}'
+# Java configuration
+{% if solr_version == '9.2.0' %}
```
acozine, line 17: This seems quite strict. Do we think future 9.x versions will require different Java versions?

roles/solrcloud/templates/solr.in.sh.j2
```diff
@@ -1,51 +1,40 @@
...
+JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64
+{% else %}
```
acozine, line 19: With this logic, if we installed 9.3.0, or even 9.2.1, we would set Solr to use Java 11. I don't think that's what we want.

roles/solrcloud/templates/solr.in.sh.j2
```diff
@@ -1,51 +1,40 @@
...
+JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
+{% endif %}
 
-# Expert: If you want finer control over memory options, specify them directly
-# Comment out SOLR_HEAP if you are using this though, that takes precedence
-# SOLR_JAVA_MEM=
+# ZooKeeper configuration
+ZK_HOST="{{ zookeeper_hosts | join(',') }}"
```
acozine, line 24: Nice! This is much tidier.

roles/solrcloud/templates/solr.xml.j2
```diff
@@ -1,53 +1,32 @@
...
-     http://www.apache.org/licenses/LICENSE-2.0
+<solr>
 
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- See the License for the specific language governing permissions and
- limitations under the License.
--->
+{% if solr_cloud_download_version is version_compare('8.6.0', '>') %}
```
acozine, line 4: I think we want "greater than or equal to" not "greater than" here.

roles/solrcloud/vars/main.yml
```diff
@@ -1,4 +1,6 @@
 ---
+# vars file for roles/solrcloud
+java_package: openjdk-11-jdk
```
acozine, line 3: This entry may override other settings for this var. I'd feel more comfortable if this is set in defaults/main.yml, because any other value will override a role default.

roles/zookeeper/defaults/main.yml
```diff
@@ -1,69 +1,19 @@
...
-
-# Rolling file appender setttings
-zookeeper_rolling_log_file_max_size: 10MB
-zookeeper_max_rolling_log_file_count: 10
-
-# List of dicts (i.e. {zookeeper_hosts:[{host:,id:},{host:,id:},...]})
-zookeeper_hosts_dict:
-  - host: '{{ lib_zk1_host_name }}'  # The leader in the Zookeeper ensemble
+# defaults file for roles/zookeeper
+zk_version: "3.4.13"
```
acozine, line 3: Are we controlling the zookeeper version? or just installing whatever the latest package version is? We probably need to think about this for the weekly package updates - could an upgrade break something?

roles/zookeeper/meta/main.yml
```diff
@@ -12,6 +12,5 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - bionic
-dependencies:
-  - role: 'openjdk'
```
acozine, line 17: here again, do we need to add the openjdk role to the converge playbook?

roles/zookeeper/molecule/default/verify.yml
```diff
@@ -3,17 +3,18 @@
...
+      loop:
+        - ruby-switch
+        - ruby2.6-dev
+        - ruby2.6
 
-  - name: test for zookeeper packages
-    assert:
-      that:
-        - not pkg_status.changed
+    - name: Test for ruby packages
```
acozine, line 17: Here again, it seems odd that we are checking for ruby packages instead of checking that zookeeper is installed and running. Was this copy pasta?

roles/zookeeper/templates/zoo.cfg.j2
```diff
@@ -1,23 +1,16 @@
...
-server.{{server.id}}={{server.ip}}:2888:3888
-{% else %}
-server.{{server.id}}={{server.host}}:2888:3888
-{% endif %}
-{% else %}
-server.{{loop.index}}={{server}}:2888:3888
-{% endif %}
+# List of servers
+{% for node in zk_nodes %}
+server.{{ node.id }}={{ node.host }}:{{ zk_leader_port }}:{{ zk_election_port }}
```
acozine, line 11: Nice! I haven't seen where we are setting that `myID` file yet, though.

acozine (commented):
Phew! This was a ton of work, thank you @kayiwa!

I've tried to follow the logic of these changes, asked a bunch of questions, and made some suggestions. Let's walk through them together.

acozine:
Oh, one more thing. I don't see where we're adding `$SOLR_OPTS -Dsolr.allowPaths=/solr/data` into the Solr9 ini file. Without that line, the line in `solr.xml` that defines `<str name="allowPaths">${solr.allowPaths:}</str>` will not work.

group_vars/solr9cloud/production.yml
```diff
@@ -1,42 +1,34 @@
----
```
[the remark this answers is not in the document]
kayiwa: linter did this and I agree with it. It is a yaml document separator.

group_vars/zookeeper/staging.yml
```diff
@@ -1,23 +1,9 @@
...
-        source: zookeeper
-        service: zookeeper
-          #To handle multi line that starts with yyyy-mm-dd use the following pattern
-        log_processing_rules:
-          - type: multi_line
-            name: start_with_date
-            pattern: \d{4}\-(0?[1-9]|1[012])\-(0?[1-9]|[12][0-9]|3[01])
+---
+# List of ZooKeeper nodes
+zk_nodes:
```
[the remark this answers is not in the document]
kayiwa: I propose making this an issue once this goes into main. Or once I am sure the solr8 will not break.

roles/zookeeper/defaults/main.yml
kayiwa: I also think this can go. We were considering the need of building zookeeper from source.

roles/zookeeper/molecule/default/verify.yml
kayiwa: copy pasta from `roles/example`

roles/zookeeper/templates/zoo.cfg.j2
kayiwa: `/var/lib/zookeeper/myid` (preferred by upstream but not package maintainer)

roles/solrcloud/defaults/main.yml
```diff
@@ -1,48 +1,38 @@
...
-solr_service_enabled: true
-solr_service_state: started
-
+# defaults file for roles/solrcloud
+solr_cloud_download_version: 8.4.1
+solr_cloud_version: "{{ solr_version }}"
+solr_version: "{{ solr_cloud_download_version }}"
+solr_user: "deploy"
+solr_group: "deploy"
+solr_data_home: "/solr/data"
```
kayiwa, line 8: this is where all the data solr generates (our stuff) lives

roles/solrcloud/defaults/main.yml
```diff
@@ -1,48 +1,38 @@
...
-  -XX:SurvivorRatio=4 \
-  -XX:TargetSurvivorRatio=90 \
-  -XX:MaxTenuringThreshold=8 \
-  -XX:ConcGCThreads=4 \
-  -XX:ParallelGCThreads=4 \
-  -XX:PretenureSizeThreshold=64m \
-  -XX:+ParallelRefProcEnabled"
-
+solr_gc_tune: "-XX:NewRatio=3 -XX:SurvivorRatio=4 -XX:TargetSurvivorRatio=90 -XX:MaxTenuringThreshold=8 -XX:ConcGCThreads=4 -XX:ParallelGCThreads=4 -XX:PretenureSizeThreshold=64m -XX:+ParallelRefProcEnabled"
+solr_gc_log_opts: "-verbose:gc -Xlog:gc=debug:file=/solr/logs/gc.txt:time,uptime,pid,tid,level,tags -Xlog:gc+heap=debug -Xlog:gc+age=trace -Xlog:safepoint"
```
kayiwa, line 31: make this follow the line break garbage collection method

PR #5848, opened 2025-02-18
Update staging inventory

Beck-Davis:
Adds subgroups to production to subdivide tower runs.

OS updates and patching have hanging/failing in tower. This could be due to running it against too many VMs. We're going to try running it against subgroups.

See similar updates for production inventory in PR #5847


dphillips-39 (changes requested):
Typo line 105

Beck-Davis:
> Typo line 105

Fixed.

dphillips-39 (approved):
Thanks Beck

PR #5860, opened 2025-02-21
Match mflux name to what's in inventory/by_environment/production

maxkadel:
This fixes a yml parse warning


carolyncole (approved):
Thanks @maxkadel

PR #5874, opened 2025-02-25
[Bibdata] add app_protect_enable off in bibdata prod nginx conf

christinach:


kayiwa (approved):
OK! Thanks @christinach

PR #5879, opened 2025-02-26
[pas] i5867 add slavery-prod2

VickieKarasic:
- adding slavery-prod2 to inventory
- updating nginxplus conf file with slavery-prod2.princeton.edu


christinach (approved):
Thanks @VickieKarasic !

PR #5885, opened 2025-02-26
Create a Traefik Bot Wall for LAE Staging

tpendragon:
This is most of what's necessary to create a captcha for all catalog paths for LAE Staging.

In this setup a request goes from Load Balancer -> Traefik -> LAE

Traefik's plugin (thank you Joe Corall) determines if a given path (/catalog) has been visited a given number of times (2) by a given IP range and if so, it redirects the user to a Cloudflare Turnstile. If they pass the turnstile, it redirects them back. Feedback from colleagues at other institutions have shown us that bots can not pass this turnstile.

Pros:

* No code changes for our Blacklight applications.
* Very configurable.
* As resilient as Nomad - if one node goes down, another will quickly replace it automatically.
* We can delay our decision about paying Cloudflare for enterprise service.
* ???

Cons:

* We'd have to put the servers we're load balancing across into this Traefik configuration.
* We gotta add some deployment mechanism into princeton_ansible, maybe a sub-directory for Nomad stuff?
* ???

Open Questions:

1. Should we run the Bot Wall on its own machine, or are we cool using Nomad for this?
2. Should we have a bot wall per application, or feed all our Blacklight apps through this?
3. Would it be easier to just implement the gem Jrochkind's working on?
4. Do we need a customized cloudflare turnstile page per application, or can we make one generic enough for every app we might use this on?

To try this:

On an IP you haven't tried it on yet go to
https://lae-staging.princeton.edu (on VPN), and browse around in the search.


roles/nginxplus/files/conf/http/lae-staging.conf
```diff
@@ -13,11 +13,14 @@ limit_req_zone $external_traffic zone=lae-staging-ratelimit:10m rate=10r/s;
 
 upstream lae-staging {
     zone lae-staging 64k;
-    server lae-staging2.princeton.edu resolve;
+    server service.consul service=node-0.traefik-wall-staging resolve max_fails=0;
+    server service.consul service=traefik-wall-staging resolve max_fails=0 backup;
```
tpendragon, line 17: This forwards all requests through one node unless it's down. Another option would be to use the ip_hash algorithm instead, so a given IP lands on a given node.

nomad_files/traefik-wall.hcl
```diff
@@ -0,0 +1,111 @@
...
+        image        = "docker.io/library/traefik:v3.3"
+        network_mode = "host"
+
+        volumes = [
+          "local/traefik.toml:/etc/traefik/traefik.toml",
+          "local/dynamic.toml:/etc/traefik/dynamic.toml",
+        ]
+      }
+
+      template {
```
tpendragon, line 45: Everything in this "template" block is the "static" configuration for Traefik.

nomad_files/traefik-wall.hcl
```diff
@@ -0,0 +1,111 @@
...
+    filename = "/etc/traefik/dynamic.toml"
+[experimental.plugins.captcha-protect]
+modulename = "github.com/libops/captcha-protect"
+version = "v1.3.2"
+EOF
+
+        destination = "local/traefik.toml"
+      }
+
+      template {
```
tpendragon, line 75: Everything in this template block is the "dynamic" configuration for Traefik. This is using the "file" provider - Traefik is usually configured via docker tags, but we can use a file instead.

nomad_files/traefik-wall.hcl
```diff
@@ -0,0 +1,111 @@
...
+  goodBots = "apple.com,archive.org,duckduckgo.com,facebook.com,google.com,googlebot.com,googleusercontent.com,instagram.com,kagibot.org,linkedin.com,msn.com,openalex.org,twitter.com,x.com"
+  persistentStateFile = "/tmp/state.json"
+  ipForwardedHeader = "X-Real-IP"
+  rateLimit = 2
+  protectParameters = "true"
+[http.services]
+  [http.services.lae-staging]
+    [http.services.lae-staging.loadBalancer]
+      [[http.services.lae-staging.loadBalancer.servers]]
+        url = "http://lae-staging2.princeton.edu:80"
```
tpendragon, line 98: For our nomad apps we could probably use a Consul provider instead, so we don't have to explicitly configure hosts, and we could add new ones without rebooting Traefik.

roles/nginxplus/files/conf/http/lae-staging.conf
tpendragon: We could also put all the regular passenger servers here as backup nodes so if Nomad crashes the sites still work, they just lose turnstile.

tpendragon:
@hackartisan @kayiwa  I'm out of time today, but I think the PR is ready again for review, and we need a follow-up ticket for some docs I left TODOs for and to figure out the high-challenge/low-challenge setup. https://github.com/pulibrary/princeton_ansible/tree/traefik_bot_wall/nomad is a lot better now I think.

There's a project directory for `traefik-wall` now, with a basic README, all the traefik configs are in YML instead of TOML, and the HCL files are pulling from files in the repository instead of inlining a bunch of configuration.

nomad/traefik-wall/README.md
kayiwa, line 1: 💯

kayiwa (approved):
OK! Thanks @tpendragon

PR #5886, opened 2025-02-26
Generate a CSV file of our GitLab issues for Airtable

acozine:
Fix for https://gitlab.lib.princeton.edu/ops/team-handbook/-/issues/24.

This playbook works to generate CSV that can be read by the Airtable connection.

Remaining work:
- Merge a skeleton playbook so we can create a Tower template and test from this branch
- Confirm a writable path on the EE to store the file during playbook execution
- Get email from Tower working to send it as an attachment - my guess is that the request to lib-ponyexpr will come from the exec-node, but I have no proof yet

PR #5887, opened 2025-02-26
adds skeleton playbook so we can set up a Tower template

acozine:
We cannot set up a template in Tower until the playbook exists in the main branch.

This branch adds a minimalist playbook to make it possible to test this functionality in Tower.

Related to #5886


kayiwa (approved):
OK! Thanks @acozine

Issue #5820, opened 2025-02-05
Failing playbook in tower should alert slack or send email that it failed


christinach:
### Maintenance
We have scheduled various playbooks to run in tower. 
When a playbook fails to finish there is an output error in AnsibleTower but nothing in Slack to indicate that it failed. 

## Acceptance Criteria
- [x] Send an alert in slack or an email that the playbook has failed

kevinreiss:
Some structural issues to having this happen consistently. You can only report failures when you Rescue an error from a block and all our playbooks don't use this pattern consistently.

acozine:
Tower has the concept of alerts for jobs - we might be able to use that instead of changing the playbooks. See https://docs.ansible.com/ansible-tower/latest/html/userguide/notifications.html. I'm happy to pair with someone to try and get this done.

christinach:
@acozine and I looked into adding notification in tower when a job fails. The email notification that @acozine had already tested is working. The slack notification needs some more investigation. For now I’m adding our team’s emails in the DACS email notifications.

christinach:
notifications are sent through email
[DACS email notification template](https://ansible-tower.princeton.edu/#/notification_templates/1/details)

PR #5892, opened 2025-02-28
add tower boxes to postfix allow list

acozine:
Partially addresses #5820.

This change adds the Tower VMs to support email notifications. Email notifications come from the Tower UI rather than from playbook/template execution. I've already run the playbook from the branch and tested the alert in the Tower UI. Last step for #5820 is to specify the conditions for alerting.


kayiwa (approved):
OK! thanks @acozine

Issue #5859, opened 2025-02-21
Remove static_tables from libstatic


maxkadel:
### What maintenance needs to be done?
Remove any static_tables configuration from libstatic - e.g. [roles/libstatic/defaults/main.yml](https://github.com/pulibrary/princeton_ansible/blob/b86fe32a4363df5fafc30b045aa907a60f701128/roles/libstatic/defaults/main.yml#L99-L104)

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### Why is this maintenance needed?
- This site now has its own servers and playbook, and including it in libstatic is redundant and confusing.

### Acceptance criteria
 - [x] Any configuration related to static_tables is removed from the libstatic role and playbook
 - [x] After running the libstatic playbook after these updates, the other sites are still functional

### Implementation notes, if any
- Even after removing the configuration, there will probably still be artifacts of the static_tables site on the application servers. It may or may not be worth removing these artifacts to avoid future confusion, it's up to whoever picks up this ticket.

aruiz1789:
merged to production and confirm sites functionality.

PR #5894, opened 2025-02-28
Remove static tables referrence from libstaic role.

aruiz1789:
At least partially addresses #5859.


maxkadel (approved):
Thanks @aruiz1789!

PR #5899, opened 2025-03-03
[getit] remove conf files

Beck-Davis:
Removed conf files for decommissioning issue [53 on ops-catchall](https://github.com/pulibrary/ops-catchall/issues/53)


VickieKarasic (approved):
thanks @Beck-Davis!

PR #5903, opened 2025-03-04
manually restart nomad boxes

kayiwa:
Prevent nomad-boxes from autorebooting.


tpendragon:
I don't think we need this now, they should be able to restart fine now that we've ensured consul is running before nomad starts.

PR #5905, opened 2025-03-04
Pulsys keys playbook fails if no hosts match at all

### LINKED ISSUE #5243: Have "Update pulsys user keys" job report failure on Tower if no hosts are found
I'm not sure how possible this is, but I was recently trying to update the pulsys user keys on a box via Tower, and was misled because the job reported success even though it did not find any servers that matched my parameters. This isn't as much a problem when running from the command line, but it's easier to miss when running from Tower. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

- The job to update the pulsys user keys is one that can often only be run from Tower. In this context, it can be easy to miss if it does not actually run on your host, since the output doesn't update without extra steps and the job reports success overall. It would be more intuitive if this particular job failed if it did not find any matching hosts.

### Acceptance criteria
 - [ ] Ansible Tower reports a failure if "Update pulsys user keys" is run and doesn't match any hosts

I'm not sure if we would want to change the overall [HOST_PATTERN_MISMATCH](https://docs.ansible.com/ansible/2.9/reference_appendices/config.html#host-pattern-mismatch) configuration, or whether it would be a way to configure it by playbook?

acozine:
Closes #5243.

The pulsys user playbook sometimes runs in Tower without adding keys to any VMs because no hosts match either play. This PR ~~adds variables for matching hosts~~ sets a fact when each play runs, and adds a third play that forces failure if neither of the first two plays ~~has any matching hosts~~ runs.

~~Could use a test run @maxkadel.~~

Update: I figured out the most common use case - if I use `limit` to run on a specific VM or group and then select an `environment` setting that does not include the VM or group, the playbook will behave as described in #5243. Here's the [Tower run](https://ansible-tower.princeton.edu/#/jobs/playbook/13669/output) that tested the functionality of this branch.

### DIFF
--- a/playbooks/pulsys_user.yml
+++ b/playbooks/pulsys_user.yml
@@ -133,6 +133,6 @@
          msg: The playbook did not add keys to any hosts, please try again.
       failed_when:
         - first_play_had_hosts is undefined
-        - second_play_had_hosts is undefinend
+        - second_play_had_hosts is undefined
       run_once: true
       delegate_to: localhost
--- a/playbooks/pulsys_user.yml
+++ b/playbooks/pulsys_user.yml
@@ -112,17 +112,19 @@
       when: ansible_play_hosts_all is search("princeton.edu")
 
 - name: Fail if neither play ran
-  hosts: localhost
+  hosts: all
 
   tasks:
 
     - name: view first play var
       ansible.builtin.debug:
         var: first_play_had_hosts
+      run_once: true
 
     - name: view second play var
       ansible.builtin.debug:
         var: second_play_had_hosts
+      run_once: true
 
     - name: Fail playbook when neither play matched a host
       ansible.builtin.fail:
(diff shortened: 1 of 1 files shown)


playbooks/pulsys_user.yml
```diff
@@ -48,6 +48,11 @@
         mode: '0600'
         backup: true
 
+    - name: set fact for hosts matched in the first play
+      ansible.builtin.set_fact:
+        first_play_had_hosts: true
+      when: ansible_play_hosts_all is search("princeton.edu")
```
tpendragon, line 54: Can you do `ansible_play_hosts_all|length>0` ?

acozine: Ah, good idea! I was thinking about whether `[ ]` would have length of greater than 0, and then the lightbulb came on. We don't need to set a condition at all! We can just set the variable. If we get `no hosts matched` the entire play gets skipped, so the variable won't get set.

maxkadel (approved):
Looks good, thanks so much @acozine !

PR #5906, opened 2025-03-05
#i5014 new feature/add warnings to replace vm playbook

dphillips-39:


acozine:
This playbook has changed since this draft PR, and the original ticket has been closed. If we want to add warnings, let's create a new ticket with specific requirements.

PR #5907, opened 2025-03-05
Follow-up to #5902: Added additional testing sandbox to inventory

dphillips-39:


kayiwa (approved):
OK! Thanks @dphillips-39

PR #5909, opened 2025-03-06
Add GitLab configuration for machine user

maxkadel:
Add GitLab configuration so that Special Collections machine user can push changes to EADs to GitLab.

Connected to https://github.com/pulibrary/lib_jobs/issues/877


regineheberlein (approved):
Thanks @maxkadel !

PR #5910, opened 2025-03-06
Don't use become/sudo in check_os_versions.yml

acozine:
With `become` set to true in the check_os_versions playbook, we see this error on some machines:

`fatal: [prds-dataspace-dtn1]: FAILED! => {"msg": "Missing sudo password"}`

The playbook gathers facts and reports them. We shouldn't need sudo/become for that.


kayiwa (approved):
OK! Thanks @acozine

PR #5915, opened 2025-03-09
[dpul] use systemd image

kayiwa:
the #5740 work updates openjdk to use the systemd image
the dpul role depends on openjdk. We also use the official redis
repository and update the image to a systemd capable one

### DIFF
--- a/roles/openjdk/molecule/default/verify.yml
+++ b/roles/openjdk/molecule/default/verify.yml
@@ -3,16 +3,16 @@
   hosts: all
   gather_facts: false
   tasks:
-  - name: check openjdk package status
-    ansible.builtin.apt:
-      name: "{{ item }}"
-      state: present
-    check_mode: true
-    register: pkg_status
-    loop:
-      - openjdk-8-jdk-headless
+    - name: Check openjdk package status
+      ansible.builtin.apt:
+        name: '{{ item }}'
+        state: present
+      check_mode: true
+      register: pkg_status
+      loop:
+        - openjdk-17-jre-headless
 
-  - name: test for openjdk packages
-    ansible.builtin.assert:
-      that:
-        - not pkg_status.changed
+    - name: Test for openjdk packages
+      ansible.builtin.assert:
+        that:
+          - not pkg_status.changed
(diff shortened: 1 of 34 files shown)


roles/openjdk/tasks/main.yml
```diff
@@ -1,27 +1,22 @@
 ---
 # tasks file for roles/openjdk
-- name: openjdk | install java (package)
+- name: Openjdk | install java (package)
   ansible.builtin.apt:
     name: "{{ java_openjdk_package }}"
     state: present
-  when:
-    - running_on_server
+    update_cache: true
```
acozine, line 7: Didn't we remove this to prevent failures related to <random_package>? I'm okay with reinstating it if we need it, but I'm having a minor flashback.

roles/openjdk/molecule/default/verify.yml
```diff
@@ -3,16 +3,16 @@
...
-    ansible.builtin.apt:
-      name: "{{ item }}"
-      state: present
-    check_mode: true
-    register: pkg_status
-    loop:
-      - openjdk-8-jdk-headless
+    - name: Check openjdk package status
+      ansible.builtin.apt:
+        name: '{{ item }}'
```
acozine, line 8: Weird that now the linter wants single quotes. But working code wins!

roles/redis/meta/main.yml
```diff
@@ -7,10 +7,11 @@ galaxy_info:
...
+  min_ansible_version: 2.9
 
   platforms:
     - name: Ubuntu
       versions:
-        - 16.04
-        - 18.04
+        - 22.04
+dependencies:
+  - role: common
```
acozine, line 17: Can we do this in the playbook instead of in the metadata?

roles/redis/tasks/main.yml
```diff
@@ -19,22 +26,22 @@
...
     src: "redis_ulimit.conf.j2"
     dest: "/etc/security/limits.d/redis.conf"
     force: true
-    mode: 0644
+    mode: "0644"
 
-- name: redis | Set vm.overcommit
-  ansible.builtin.sysctl:
+- name: Redis | Set vm.overcommit
+  ansible.posix.sysctl:
```
acozine, line 37: ah, sneaky

acozine (commented):
Overall looks good, and this is NOT a blocking review. Just a couple of things jumped out at me, see below.

aruiz1789 (approved):

PR #5917, opened 2025-03-11
[Traefik Wall] Fix wireless IPs being hit with captcha.

tpendragon:
Anyone on our wireless shouldn't be hit with captcha - this upgrades the plugin and fixes the config so it's fine.

### DIFF
--- a/nomad/traefik-wall/deploy/traefik.tpl.yml
+++ b/nomad/traefik-wall/deploy/traefik.tpl.yml
@@ -22,4 +22,4 @@ experimental:
   plugins:
     captcha-protect:
       modulename: github.com/libops/captcha-protect
-      version: v1.5.0
+      version: v1.6.1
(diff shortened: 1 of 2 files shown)


nomad/traefik-wall/deploy/traefik.tpl.yml
```diff
@@ -10,6 +10,7 @@ entryPoints:
       trustedIPs:
         - 128.112.200.245/32
         - 128.112.201.34/32
+        - 10.0.0.0/8
```
kayiwa, line 13: wanna just add these then?

https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/all/vars.yml#L141

tpendragon: Oh is https://www.net.princeton.edu/ip-network-ranges.html not right? That's what I have in exemptIps

tpendragon: It turns out I didn't need this, just previous exemptIps, and upgrading the plugin.

kayiwa (approved):
OK! Thanks @tpendragon

nomad/traefik-wall/deploy/traefik.tpl.yml
kayiwa: The network ranges you have a correct for your use case. The group vars are "the library segment" of that range.

PR #5919, opened 2025-03-11
[openjdk] add openjdk tests

kayiwa:
we want to ensure the openjdk role is tested
remove the running on server
we manually create the directory for manuals as a pre-task

related to #5740


kayiwa:
closed by #5915

PR #5920, opened 2025-03-11
I5722 replace vm

aruiz1789:


acozine (commented):
This makes the playbook much tidier. Just one suggestion for future-proofing, see below.

playbooks/utils/replace_vm_host.yml
```diff
@@ -97,108 +92,73 @@
...
+             {{ old_vm_info.virtual_machines[0].allocated.cpu }} CPUs
+             {{ (old_vm_info.virtual_machines[0].allocated.memory | int / 1024) }} GB of memory
+             {{ old_vm_info.virtual_machines[0].allocated.storage | human_readable }} of disk allocated
       ansible.builtin.debug:
-        var: new_vm_deets
+        msg: "{{ msg.split('\n') }}"
 
-    - name: delete NIC with automatic mac address
-      community.vmware.vmware_guest_network:
+    - name: Remove old VM using UUID
```
acozine, line 115: I would keep this step as the very last in the playbook, so we don't delete the old VM until we are 100% sure everything else worked and the rest of the playbook succeeded. We don't know what might break in future, and I'll sleep better knowing that if anything else starts failing, we will keep the old VM for forensics.

kayiwa (approved):
OK! Thanks @aruiz1789

playbooks/utils/replace_vm_host.yml
aruiz1789: will do.

Issue #5918, opened 2025-03-10
Adapt current configuration in ansible for use in the catalog application servers


kevinreiss:
- [x] Traefik configuration based on the existing pulfalight/lae configurations is active in staging. 
- [x] If successful create a ticket for implementing this in production. 

## Implementation Notes 
PRs to review

1. [Finding Aids](https://github.com/pulibrary/princeton_ansible/pull/5911)
2. [Lae](https://github.com/pulibrary/princeton_ansible/pull/5885)

PR #5925, opened 2025-03-14
[catalog] add traefik-wall

kayiwa:
- **add catalog-staging to traefik-wall**
- **catalog has two staging servers**

related to #5918 

Not sure this is effective on staging given how it is restricted to on-prem


kayiwa:
run successfully on dev load balancers

hackartisan (approved):

PR #5933, opened 2025-03-14
sample Vagrantfile for qemu

kayiwa:


acozine:
The original goal of this work was to create local, full-blown VMs for testing new applications and other changes. However, we are using sandbox machines in vSphere for this now - also, Vagrant does not support all chip types, so this would have been a hack.

PR #5941, opened 2025-03-17
[Traefik Wall] Fix redirect for production Catalog.

tpendragon:
This makes sure we only rewrite `?f[]=` if it's right after the first slash in a path, so it no longer matches on `https://catalog.princeton.edu/catalog/?f[subject_facet][]=Catrysse%2C+Wim+1973-%E2%80%94Exhibitions`


christinach (approved):
Thanks @tpendragon ! 🐻‍❄️

PR #5950, opened 2025-03-18
[pas] add missing environment variable

sandbergja:
This addresses the following warning, which fills up the logs at /var/log/apache2/error.log:

```
PHP Warning:  Undefined array key ENVIRONMENT in /opt/pas_cap/releases/20250310191309/bootstrap.php on line 22
```


maxkadel (approved):
Thanks @sandbergja !

PR #5955, opened 2025-03-18
[postfix] Remove duplicate .princeton.edu domain when obtaining TLS certificate

sandbergja:
In this case, `domain_name` is set to the `inventory_hostname` (lib-ponyexpr-prod.princeton.edu), so appending .princeton.edu leaves us with lib-ponyexpr-prod.princeton.edu.princeton.edu.

Helps with https://github.com/PrincetonUniversityLibrary/pas-craft3/issues/147

### DIFF
--- a/roles/postfix/tasks/main.yml
+++ b/roles/postfix/tasks/main.yml
@@ -66,7 +66,7 @@
   notify: restart postfix
 
 - name: Postfix | update acme certificates for {{ domain_name }}
-  ansible.builtin.command: /usr/bin/certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme.sectigo.com/v2/InCommonRSAOV --eab-kid {{ vault_acme_eab_kid }} --eab-hmac-key {{ vault_acme_eab_hmac_key }} --domain {{ domain_name }}.princeton.edu --cert-name {{ domain_name }}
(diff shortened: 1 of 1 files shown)


roles/postfix/tasks/main.yml
```diff
@@ -66,7 +66,7 @@
   notify: restart postfix
 
 - name: Postfix | update acme certificates for {{ domain_name }}
-  ansible.builtin.command: /usr/bin/certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme.sectigo.com/v2/InCommonRSAOV --eab-kid {{ vault_acme_eab_kid }} --eab-hmac-key {{ vault_acme_eab_hmac_key }} --domain {{ domain_name }}.princeton.edu --cert-name {{ domain_name }}
+  ansible.builtin.command: /usr/bin/certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme.sectigo.com/v2/InCommonRSAOV --eab-kid {{ vault_acme_eab_kid }} --eab-hmac-key {{ vault_acme_eab_hmac_key }} --domain {{ domain_name }} --cert-name {{ domain_name }}
```
VickieKarasic, line 69: This looks fine, as long as {{ domain_name }}  is defined as the FQDN (it looks like that's the case in the vars for this role).

christinach (approved):
Thanks @sandbergja 🍄 !

kevinreiss:
After we deploy this change we need to revoke the old double princeton.edu cert via this procedure https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/incommon_certbot_revoke.yml.

PR #5958, opened 2025-03-19
[Traefik Wall] Add traffic tracking for /advanced route.

tpendragon:
The catalog's advanced search is very similar to /cat
alog, and is getting hit in the same way.


christinach (approved):
Thanks @tpendragon !

PR #5965, opened 2025-03-21
removes config for theses-dissertations, redirects thesis-central

acozine:
The new thesis and dissertation submission platform is hosted in AWS and managed by 4Science.

We have a CNAME record pointing theses-dissertations.princeton.edu to the new site (and away from our nginxplus load balancers). That is working, so the nginxplus config for theses-dissertations is obsolete. This PR removes it.

Many departments are still sharing the old thesis-central URL with their students. This PR redirects that URL appropriately for the new site. Eventually we hope to retire the thesis-central name, but we cannot do that yet.


VickieKarasic (approved):
Looks good to me @acozine - thanks!

PR #5368, opened 2024-09-20
Updates libweb2 config

acozine:
We deployed the changes from #5366 but the redirect did not go to the correct page.

Trying again.

PR #5968, opened 2025-03-21
Restore libweb5-redirects.conf, delete libweb2-redirects.conf

### LINKED ISSUE #5967: Missing redirect file causes nginx -t to fail
### Expected behavior
I can run the full nginxplus.yml playbook with `-e nginx_cleanup_config=true`, clear out all pre-existing config files, and restart nginx.

### Actual behavior
When I run the nginxplus.yml playbook with `-e nginx_cleanup_config=true`, the playbook fails on the task that checks `nginx -t`, complaining about a missing config file.

### Error messages
```
fatal: [adc-prod2.princeton.edu]: FAILED! => {"changed": false, "cmd": ["/usr/sbin/nginx", "-t"], "delta": "0:00:00.113205", "end": "2025-03-21 20:54:37.476062", "msg": "non-zero return code", "rc": 1, "start": "2025-03-21 20:54:37.362857", "stderr": "nginx: [emerg] open() \"/etc/nginx/conf.d/templates/libweb5-redirects.conf\" failed (2: No such file or directory) in /etc/nginx/conf.d/libweb5.conf:20\nnginx: configuration file /etc/nginx/nginx.conf test failed", "stderr_lines": ["nginx: [emerg] open() \"/etc/nginx/conf.d/templates/libweb5-redirects.conf\" failed (2: No such file or directory) in /etc/nginx/conf.d/libweb5.conf:20", "nginx: configuration file /etc/nginx/nginx.conf test failed"], "stdout": "", "stdout_lines": []}
```

This is fair enough, since `/etc/nginx/conf.d/libweb5.conf` includes (or tries to include) `/etc/nginx/conf.d/templates/libweb5-redirects.conf`.

We removed `libweb5-redirects.conf` as part of #5368. I think we may have removed the wrong file - perhaps we intended to remove `libweb2-redirects.conf`?

### Steps to replicate
Run the nginxplus.yml playbook with `-e nginx_cleanup_config=true`.
But please don't!

### Impact of this bug
We may be missing some redirects.
Also, our nginx playbook is not reliable.

Restore `libweb5-redirects.conf`. Check whether we still need `libweb2-redirects.conf` - if we don't, remove it.

### LINKED ISSUE #5976: [Traefik Wall] Add Common Crawl to the good bots list.
See https://commoncrawl.org/ccbot

As per a code4lib presentation, they're friendly and nice bots, with a mission we can support.

acozine:
Closes #5967.

In #5368 we deleted the wrong redirect file. We should have deleted the redirects for libweb2; instead we deleted the redirects for libweb5. This PR fixes our mistake.

### DIFF
--- a/roles/nginxplus/files/conf/http/templates/libweb2-redirects.conf
+++ b/roles/nginxplus/files/conf/http/templates/libweb2-redirects.conf
@@ -1,1 +1,0 @@
-rewrite http://libweb2.princeton.edu/rbsc2/(.*)$ https://static-prod.lib/princeton.edu/scsites/$request_uri permanent;
--- a/roles/nginxplus/files/conf/http/templates/libweb5-redirects.conf
+++ b/roles/nginxplus/files/conf/http/templates/libweb5-redirects.conf
@@ -1,0 +1,1 @@
+rewrite ^/visual_materials/(.*)$ https://library.princeton.edu$request_uri permanent;


sandbergja (approved):
Thanks, @acozine ! 🐴

PR #5969, opened 2025-03-24
[GLOBUS] Add the ability to specify the globus environment separte from the instance name

carolyncole:


VickieKarasic (approved):
Thanks @carolyncole and @leefaisonr - this makes sense to me. I'll let other Ops folks review as well.

acozine:
Changes in #5505 broke the PDC globus build capability. This PR restores it.

Specifically, the PDC projects have configuration for staging EC2 instances and configuration for prod EC2 instances, which is currently the same for pre-curation, post-curation, deposit, and embargo.  This PR supports having the same setup for all of PDC-related prod EC2 instances and all PDC-related staging EC2 instances, requiring only two vars files (instead of six). 

For other uses of globus, we can still have one var file per EC2 instance by passing the same value for both variables.

PR #5971, opened 2025-03-25
Updates for new checkmk

carolyncole:


acozine:
See also #6020 and #6013

PR #5973, opened 2025-03-25
Update ezproxy configuration templates. 

kevinreiss:
Made EZproxy config.txt consistent with working production configuration.


VickieKarasic (approved):
thanks @kevinreiss

PR #5974, opened 2025-03-26
Moved utility playbooks to the playbooks/utils directory

### LINKED ISSUE #4544: Move existing "utility" playbooks into the playbooks/utils directory
Follow-up to #4500.

Most of the playbooks in our `playbooks` directory are project-related, but we have a few that work across projects or are otherwise "utility" playbooks (things like certificate management, prompt updates, etc.).

Move these playbooks into the new `playbooks/utils` directory.

Beck-Davis:
Moved "utility" playbooks that are not project-specific or can be applied to any host to playbooks/utils directory. Closes #4544


kayiwa (approved):
OK! Thanks @Beck-Davis

PR #5980, opened 2025-03-26
[prancible] add a condition for Rapid7 in security_theater playbook

aruiz1789:
Unlike the other tools in the `security_theater` playbook, Rapid7 did not have an initial task to `Install, configure, and start [the service]` before proceeding to download the .deb file. This was causing the playbook to fail on the Rapid7 task to install the agent when Rapid7 was already on a machine, since it detected the deb file on the machine and failed to configure the agent. The task we added allows for a check of the Rapid7 service first before installing the file, and if it is already installed, it will skip the task to download the .deb file and will proceed to update to the latest version of the agent if it's already there.

### DIFF
--- a/playbooks/utils/security_theater.yml
+++ b/playbooks/utils/security_theater.yml
@@ -102,55 +102,58 @@
         ansible.builtin.command: /etc/init.d/besclient start
       #For future maintenance, BigFix block ends here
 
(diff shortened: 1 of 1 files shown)


VickieKarasic (approved):
thanks @aruiz1789

kayiwa (approved):
OK! Thanks @aruiz1789 and @VickieKarasic

PR #5990, opened 2025-03-28
Add a reminder about revoking old certs

acozine:
The mail servers host their own SSL certificates. They auto-renew, so we don't normally interact with them much. However, when we rebuild a mail server using the postfix role, we create fresh certificates. When we do that, we need to revoke the previous certificates, or they will propagate endlessly. This PR adds a reminder to slack to the tasks file for the postfix role.

### DIFF
--- a/playbooks/postfix.yml
+++ b/playbooks/postfix.yml
@@ -16,6 +16,12 @@
     - role: roles/postfix
 
   post_tasks:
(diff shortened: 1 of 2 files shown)


acozine:
D'oh, right, I need to put this in the post-tasks in the playbook. Will do this next week.

Issue #5988, opened 2025-03-28
Investigate and Resolve Frequent CIFS Mount Hanging Alerts


kayiwa:
The Ops team is experiencing a high volume of "CIFS mount hanging" alerts originating from CheckMK for various virtual machines. This is believed to be an underlying infrastructure problem rather than a monitoring artifact, potentially causing instability for applications relying on these mounts.

**Background & Context:**

- This issue has been persistent for some time. The application "Figgy" has implemented workarounds for dropped connections, but cannot handle cases where the mount disappears entirely, which sometimes occurs.
- Recent monitoring added to Figgy by Anna checks /proc/mounts for known mounts.
- An initial attempt to resolve the issue by writing timestamps to the storage (to keep connections alive) did not fix the problem.

**Affected Systems (Observed):**

- bibdata-qa1
- bibdata-qa2
- dpul-prod1
- figgy-web-prod1
- figgy-web-prod2
- figgy-web-prod3
- figgy-web-prod4
- figgy-worker-prod1
- figgy-worker-prod2
- figgy-worker-prod3
- figgy-worker-prod4
- figgy-worker-prod5
- figgy-worker-prod6
- lib-approvals-prod1
- lib-approvals-prod2
- lib-jobs-prod1
- lib-jobs-prod2
- maps-prod1
- slavery-prod1

**Proposed Solutions & Investigation Paths:**

1. Network Improvement:

- Action: Test moving affected VMs to the production private network, which offers greater throughput.

- [x] bibdata-qa1
- [x] bibdata-qa2
- [x] bibdata-worker-qa1
- [x] bibdata-worker-qa2
- [x] dpul-prod1
- [x] dpul-prod2
- [x] dpul-staging3
- [x] dpul-staging4
- [x] figgy-web-prod4
- [x] figgy-web-prod3
- [x] figgy-worker-prod1
- [x] figgy-worker-prod2
- [x] figgy-worker-prod3
- [x] figgy-worker-prod4
- [x] figgy-worker-prod5
- [x] figgy-worker-prod6
- [x] figgy-web-staging1
- [x] figgy-web-staging2
- [x] lib-approvals-prod1
- [x] lib-approvals-prod2
- [x] lib-jobs-prod1
- [x] lib-jobs-prod2
- [x] maps-prod1
- [x] slavery-prod1

**(Possible) Next Steps:**

2. **Protocol Migration (CIFS to NFS):**

[...]

PR #5991, opened 2025-03-31
move bibdata to the private network

kayiwa:
move the bibdata-qa1 to the private network. This allows us to see if the reduce the noise from failing cifs mounts from checkmk

related to #5988

### DIFF
--- a/group_vars/nfsserver/qa.yml
+++ b/group_vars/nfsserver/qa.yml
@@ -1,6 +1,6 @@
 ---
 # servers
-bibdata_qa1: "128.112.204.65"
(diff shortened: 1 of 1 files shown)


christinach (commented):
@kayiwa do we also need to move bibdata_qa2 ? I'm seeing a different ip only for [bibdata_qa1](https://github.com/pulibrary/princeton_ansible/pull/5991/files#diff-03c43e77df2f6dba29069e85debcf7ce94147c93ddb61f067c62cae57fac766aR3).

christinach (approved):
Thanks @kayiwa 🦒

PR #6010, opened 2025-04-02
Add a common.yml for figgy group vars

### LINKED ISSUE #4952: Create shared vars file for figgy
Currently we have vars files for each figgy environment. This involves some duplication of settings. To minimize maintenance and increase efficiency, we can create a `common.yml` file for the vars that are the same across all envs.

Follow-up to #4948

Beck-Davis:
Reformatted figgy production and staging group vars consolidating identical values into a single file.

Closes #4952

### DIFF
--- a/group_vars/figgy/common.yml
+++ b/group_vars/figgy/common.yml
@@ -1,0 +1,52 @@
+---
+application_db_name: '{{figgy_db_name}}'
+application_dbuser_name: '{{figgy_db_user}}'
+application_dbuser_password: '{{figgy_db_password}}'
+application_dbuser_ro_name: '{{figgy_db_ro_user}}'
+application_dbuser_ro_password: '{{figgy_db_ro_password}}'
+application_dbuser_role_attr_flags: 'SUPERUSER'
+application_host: '{{passenger_server_name}}'
+application_host_protocol: 'https'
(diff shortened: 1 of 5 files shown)


tpendragon:
@Beck-Davis This is very helpful, thank you! I think my main question is more about process, and maybe @acozine can help here. With these common files now I need to know what to do if I want to change staging out of step with production - for instance, if I want to update NodeJS on staging, but not prod, do I change common and not deploy the staging playbook? Or do I add the nodejs version to the staging section temporarily?

I'm just out of date here, I haven't used playbooks with a common vars file.

Beck-Davis:
My understanding is that if the value (nodejs) differs, we would specify the versions in the production and staging specific files. Do I have that right, @acozine ?

acozine:
Here's how I usually use common/shared vars files:

If there are values that happen to be identical now, but we expect to differ regularly, we can put them back into the staging and production files.

If it's more that we want to test an upgrade by updating only the staging servers, and if that works we want to upgrade prod, then I would make the change in the common file on a branch, try it out in staging from the branch, and when everything works and it's time to upgrade production, merge the change and run the playbook against prooduction.

Does that approach work here?

tpendragon (approved):

tpendragon:
Sounds good to me, thank you!

Issue #5959, opened 2025-03-19
[checkmk] utils playbooks do not run.


carolyncole:
### Expected behavior
I would expect to be able to run the playbooks/utils/checkmk_add_local_checks.yml without error

I would also expect to be able to run playbooks/utils/checkmk_agent.yml without error

### Actual behavior

both playbooks error.  I believe the variables that were setup are no longer valid.  both call fir checkmk to be running on http.  Even with tweaking the variables I couldn't get them to connect with checkmk 

### Error messages
before changing to https
```
fatal: [pdc-describe-redis-staging1.lib.princeton.edu]: FAILED! => {"changed": false, "msg": "404 - Not found Details: b'<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">\\n<html><head>\\n<title>404 Not Found</title>\\n</head><body>\\n<h1>Not Found</h1>\\n<p>The requested URL was not found on this server.</p>\\n<hr>\\n<address>Apache/2.4.52 (Ubuntu) Server at pulmonitor-staging1.princeton.edu Port 80</address>\\n</body></html>\\n'"}
```

After changing to https
```
TASK [Add new local services on hosts.] ***************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
[...]

PR #6013, opened 2025-04-03
Supports agent installs in the paid version of CheckMK

acozine:
Partially addresses #5959.

When we looked at the errors with running the agent playbook, we discovered that the playbook ran successfully if we first manually added the `host` record on the CheckMK site where we wanted the host to be monitored. The first commit on this branch recreates that fix by adding a variable value allowing hosts to be created. I confirmed this by adding my sandbox to the `production` CheckMK site using the `checkmk_agent.yml` playbook: 
`ansible-playbook -e runtime_env=sandbox -e checkmk_agent_site=production -e checkmk_folder=linux/ops playbooks/utils/checkmk_agent.yml --limit sandbox-acozine.lib.princeton.edu`

**Remaining work:**

- Originally we were using `runtime_env` to manage the host group(s) and the checkmk vars - for example, run the playbook against `bibdata_staging` and register the agent on those VMs with the `staging` CheckMK site. I'm not sure that works well in our new distributed environment. There's probably a better way to do this.
- The group_vars also need to be updated for the new distributed environment - the `checkmk_agent_site` (the last bit of the URL, as in pulmonitor-staging.princeton.edu/production) and the `checkmk_agent_server` (the VM where that 'site' is managed) need to match now that we have separate VMs for each site.
- Also, I think the playbook for adding rules is still broken. At least, I haven't tested it, and I can't see how this would fix it.

### DIFF
--- a/group_vars/checkmk/agent_common.yml
+++ b/group_vars/checkmk/agent_common.yml
@@ -1,6 +1,7 @@
 ---
 checkmk_agent_version: "2.3.0p15"
 checkmk_agent_edition: cee
+checkmk_agent_add_host: 'true'
 checkmk_agent_server_protocol: http
 checkmk_agent_server: pulmonitor-staging1.princeton.edu
 checkmk_agent_server_port: "{% if checkmk_agent_server_protocol == 'https' %}443{% else %}80{% endif %}"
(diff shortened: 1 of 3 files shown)


acozine:
As of today, both the agent playbook and the rules playbook work!

acozine:
Known remaining work is documented in #6018. But this unblocks us, at least for the staging site. We will test against production in the next iteration.

PR #6021, opened 2025-04-07
[bibdata] move to private network

kayiwa:
we reduce the noise from CIFS mounts dropping connections by moving them
to the private network

related to #5988


christinach (approved):
Thanks @kayiwa ! 🦒

PR #6024, opened 2025-04-08
Add clamav installation role

bess:
Ref https://github.com/pulibrary/pdc_describe/issues/2080


acozine:
The goal here is to scan files submitted by users. Will running ClamAV while also running the required OIT security tools (Cortex XDR, Rapid7) cause any problems? Will ClamAV's actions be seen as potentially malicious by those other tools? 

Also, can the required OIT tools accomplish the same goals? Do we need all three tools?

acozine:
We ran this from the branch against pdc-describe-staging1 today.

PR #6028, opened 2025-04-08
[Catalog] Add a test traefik wall that can be quickly adjusted just for catalog testing

tpendragon:
Created with @christinach 

We wanted a wall that could be changed without affecting all the other sites.


christinach (approved):
Thanks @tpendragon !

PR #6037, opened 2025-04-10
[Bibdata] Add Prometheus to Bibdata QA / DPUL Staging

tpendragon:
We need a service that connected to one of our NFS shares to compare
performance and rule out networking issues. This seemed lowest impact.


kayiwa (approved):
OK! Thanks @tpendragon

PR #6040, opened 2025-04-11
add canary host to inventory

kayiwa:
our canary host was not getting patched. This adds the host to the pulcheck group so it will get patched on Tuesdays


Beck-Davis (approved):
Thank you @kayiwa 🧙🏽

Issue #6042, opened 2025-04-12
[Checkmk Runner]  Limit Syslog Size via Logrotate Configuration


kayiwa:
The syslog file on catalog-qa{1,2} is growing excessively large, reaching 30GB. This is consuming a significant amount of disk space and could potentially impact system performance.

```bash
pulsys@catalog-qa2:/$ dust -rp
No such file or directory: ./proc/1104506, ./opt/install/ruby-3.4.1, ./opt/install/ruby-3.1.0, ./opt/install/ruby-3.3.6
Did not have permissions for all directories
  74G └─┬ .
  64G    ├─┬ ./var
  63G    │ ├─┬ ./var/log
  30G    │ │ ├── ./var/log/syslog.1
  28G    │ │ ├── ./var/log/syslog
4.0G    │ │ ├─┬ ./var/log/journal
4.0G    │ │ │ └── ./var/log/journal/fac779f3aad04be88ee0ff6d8d3573ac
705M    │ │ └── ./var/log/syslog.2.gz
1.0G    │ └─┬ ./var/lib
736M    │    └─┬ ./var/lib/snapd
510M    │      └── ./var/lib/snapd/snaps
4.8G    ├─┬ ./usr
3.1G    │ ├─┬ ./usr/lib
1.1G    │ │ ├── ./usr/lib/firmware
953M    │ │ └─┬ ./usr/lib/modules
456M    │ │    ├─┬ ./usr/lib/modules/5.15.0-135-generic
450M    │ │    │ └── ./usr/lib/modules/5.15.0-135-generic/kernel
456M    │ │    └── ./usr/lib/modules/5.15.0-136-generic
839M    │ └── ./usr/local
3.5G    ├─┬ ./opt
3.4G    │ └─┬ ./opt/orangelight
2.0G    │    ├─┬ ./opt/orangelight/shared
1.9G    │    │ └─┬ ./opt/orangelight/shared/bundle
1.9G    │    │    └─┬ ./opt/orangelight/shared/bundle/ruby
968M    │    │      ├─┬ ./opt/orangelight/shared/bundle/ruby/3.1.0
751M    │    │      │ └── ./opt/orangelight/shared/bundle/ruby/3.1.0/gems
570M    │    │      └─┬ ./opt/orangelight/shared/bundle/ruby/3.4.0
455M    │    │        └── ./opt/orangelight/shared/bundle/ruby/3.4.0/gems
1.3G    │    └── ./opt/orangelight/releases
[...]

PR #6046, opened 2025-04-15
[orangelight] remove deploy user file owner ship

kayiwa:
theres a discrepancy with file ownership for brownfield hosts with will sometimes have root own the files

followup work on #6042


VickieKarasic (approved):
Looks good @kayiwa - thanks!

PR #6048, opened 2025-04-15
[bibdata] add prometheus collection

### LINKED ISSUE #6047: [bibdata] Ansible Playbook Error: Role Not Found
## The Ansible playbook fails with a "role not found" error. Specifically, the error indicates that the role prometheus.prometheus.node_exporter cannot be found when run on ansible-tower

```bash
ERROR! the role 'prometheus.prometheus.node_exporter' was not found in /runner/project/playbooks/roles:/runner/requirements_roles:/runner/project/external_roles:/root/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/runner/project/playbooks
The error appears to be in '/runner/project/playbooks/bibdata.yml': line 45, column 7, but may
be elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
    - role: prometheus.prometheus.node_exporter
      ^ here
```

**Possible Causes**

- The role prometheus.prometheus.node_exporter is not installed.
- The role is not in one of the paths where Ansible searches for roles.
- There is a typo in the role name in the playbook.

## Impact of this error
I cannot run the bibdata playbook against any of the environments including production.

## Notes
This change was added https://github.com/pulibrary/princeton_ansible/commit/c3944a933697cc1f51ad2d35a71794285ab2783d as a test . Related [slack thread](https://pulibrary.slack.com/archives/CL8F616KV/p1744302620954039)

kayiwa:
we have roles that pull directly from ansible galaxy to install
prometheus. this change allows playbooks to successfully run from
Ansible Tower when the new execution environment is rebuilt

closes #6047

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -58,6 +58,7 @@ ansible-creator = "*"
 netaddr = "*"
 passlib = "*"
 bcrypt = "*"
+ansible-builder = "*"
 
 [dev-packages]
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "c2c1892baa24341294114ac003c5ee787180c1b05e9eaba5fdd6635daddd4168"
+            "sha256": "047200cd7ddace0cc7bd62923e1f3128ac0a80bd3a4ef8a22a09dd79c8cd7129"
         },
         "pipfile-spec": 6,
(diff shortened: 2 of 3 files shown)


christinach (approved):
Thank you @kayiwa @acozine 🐴 🦒

acozine:
We don't need prometheus any more (we can light our own fires).

PR #6055, opened 2025-04-16
[videoreserves] Install needed php modules for PHP 8.4

sandbergja:
Ansible thought we were using PHP 8.3, but the staging and prod servers were using 8.4 instead, and missing some needed php modules.

Closes https://github.com/PrincetonUniversityLibrary/video_reserves/issues/92


christinach (approved):
Thanks @sandbergja !

PR #6058, opened 2025-04-16
[Traefik Wall] Add health checks.

### LINKED ISSUE #6056: [Traefik Wall] Traefik services should check health endpoints to bring services up/down if endpoints are down
Figgy-web-prod4 just crashed and folks were getting gateway timeout errors - that's what happens when Traefik's down. We realized that nginx has health checks (good!), but Traefik didn't - so nginx would check once, twice, three times, and Traefik would round robin so not all three of those would fail - just the one that Traefik picked.

Traefik has health checks, we just need to add them and match our nginx config.

tpendragon:
Closes #6056


tpendragon:
I gotta watch this for a little bit.

kayiwa (approved):
OK! Thanks @tpendragon

PR #6060, opened 2025-04-16
[Repec] Don't install nodejs

sandbergja:
It is not needed for this application.  Helps with https://github.com/pulibrary/repecwp/issues/165


christinach (approved):
thanks @sandbergja !

PR #6061, opened 2025-04-16
[Traefik Wall] Export metrics to Prometheus.

### LINKED ISSUE #6057: [Traefik Wall] Add observability
We don't have a great way to track the logs for Traefik or see its status.

We need some form of visibility - either a Grafana dashboard or access to the Traefik dashboard. Maybe both? Also preferably somewhere to aggregate the logs. Maybe also our Grafana instance.

tpendragon:
We need to be able to tell when nodes are down.

There's some commits at the end of this, because once I had observability I was able to notice that DPUL staging was repeatedly crashing.

Closes #6057


kayiwa:
will need to resolve the conflicts @tpendragon

PR #6063, opened 2025-04-17
Remove prometheus. It was added do test metrics. This work is complete.

christinach:
Currently it does not work and fails the playbook to run against any environment.


maxkadel (approved):
Thanks @christinach!

Issue #6065, opened 2025-04-17
[cicognara] migrate cicognara to nomad infrastructure


kayiwa:
Migrate cicognara to Nomad infrastructure. Move the project away from the static site and use container orchestration from nomad for lower maintenance.

kayiwa:
heya @escowles what is the repository to clone for the updated cicognara static site. My assumptions here is that it is jekyll based.

escowles:
Yes, we have a parent Cicognara repository, with a Jekyll site, and we have checked in the `_site` directory to have a completely static version of the site to deploy. Here's that directory: https://github.com/pulibrary/digital-cicognara-library/tree/main/apps/cicognara-static/_site

In theory, it would be nice to have a playbook to build Jekyll sites without having to checkin the fully-static version. But so far it's been OK just checking the `_site` directory in.

acozine:
Closed by #6066

PR #6066, opened 2025-04-17
[cicognara] add replacement cicognara config

kayiwa:
we remove the lib-static configurations and add the container based configuration we also remove references to the old cicognara role

related to #6065


kayiwa:
run successfully on dev load balancer.

PR #6067, opened 2025-04-17
[Observability] Add Nomad observability infrastructure.

tpendragon:
This is the observability infrastructure we built to monitor DPUL-Collections in staging. Now that it's also monitoring a mount and other folks are using Nomad, it seemed prudent to move it here.


kayiwa (approved):
OK! Thanks @tpendragon

PR #6068, opened 2025-04-18
[lib-jobs] Don't install node

sandbergja:
As of https://github.com/pulibrary/lib_jobs/pull/907, it is no longer needed.

Helps with https://github.com/pulibrary/lib_jobs/issues/896

I ran this on staging, and also manually removed the node and yarn binaries from the staging boxes.


christinach (approved):
thanks @sandbergja 🍄

PR #6072, opened 2025-04-21
[DSS] Use latest node version 24.x

christinach:
closes https://github.com/pulibrary/DSS/issues/530

use latest node version 24.x
Release comes out on [4/22/2025](https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule)

See also https://github.com/pulibrary/DSS/pull/553


sandbergja (approved):
Thanks, @christinach  🐬

Issue #6075, opened 2025-04-23
[adc] Implement Log Rotation for /var/log/app_protect/security.log to Prevent Disk Overfill


kayiwa:
The file `/var/log/app_protect/security.log` is rapidly growing and contributing significantly to disk usage, currently occupying ~25GB. This growth rate has been observed to quickly fill up available disk space on the system, leading to potential stability issues.

**Observed Behavior:**

- As shown in the disk usage snapshot:

  - `/var` is consuming ~74GB
  - `var/log` is consuming ~73GB
  - `/var/log/app_protect/security.log` alone is consuming ~25GB, which is ~31% of /var/log.

**Impact:**

Excessive disk usage.

Potential risk of running out of disk space.

Service disruptions or logging failures if disk fills.

**Potential Resolution and first steps:**

1. Modify the logrotate configuration at /etc/logrotate.d/app_protect.conf:

```bash
"/var/log/app_protect/*.log" {
    compress
    copytruncate
    create 0640 nginx nginx
    daily
    delaycompress
    notifempty
    missingok
    maxsize 5G
    rotate 14
    postrotate
        /bin/systemctl restart nginx-app-protect.service > /dev/null 2>/dev/null || true
    endscript
}
```
2. Ensure proper ownership and permissions on the log directory:

```bash
chown -R nginx:nginx /var/log/app_protect
chmod 750 /var/log/app_protect
```

3. Test logrotate manually:

```bash
sudo logrotate -f /etc/logrotate.d/app_protect.conf
```

4. Monitor over the next few days to ensure space is reclaimed and log files rotate as expected.

**Acceptance Criteria:**

- [ ]  Log rotation for /var/log/app_protect/security.log is active.
- [ ]  Logs are compressed and rotated daily.
- [ ]  Disk usage from this log file remains under control (<2GB ideally).
- [ ]  Logrotate does not interfere with app_protect service stability.

kayiwa:
create new ticket that describes the contents of `/var/log/app_protect/security.log` and possibly add as a "wont fix" label.

kayiwa:
templatize this to a consistent thing?
path to log
then the generalized expectation
and the application

acozine:
A quick search turned up 4 examples of log rotation templates in this repo. They cover:
[PostgreSQL](https://github.com/pulibrary/princeton_ansible/blob/main/roles/postgresql/templates/logrotate.d.postgresql.j2)
[...]

PR #6079, opened 2025-04-24
[nginxplus] monitor logs

kayiwa:
we get a warning from datadog when the disk is about to fill up.
This playbook installs a checkmk local check and warns us with a
parametrized size of the security log is too large
we can then force a logrotate.
This will eventually be obsolete by better logroation

related to #6075


aruiz1789 (approved):
log size check its a good idea.

PR #6088, opened 2025-04-25
Remove default DNS servers from dnsmasq

tpendragon:
On Nomad, DNS was falling down to 8.8.8.8. We want it to pull from whatever DNS gave the machine, which is the PUL nameservers.

This was causing propagation of internal IPs to take ~8 hours.


kayiwa (approved):
Cool! thanks @tpendragon

PR #6101, opened 2025-04-28
[bibdata] add catalog_sync token for figgy integration

maxkadel:
Connected to https://github.com/pulibrary/bibdata/issues/2693


christinach (approved):
Thanks @maxkadel 🦄 !

PR #6103, opened 2025-04-28
Updates the replace-a-vm playbook for CrowdStrike usage

acozine:
When we are replacing a VM in response to an intrusion, we want to keep the old VM (powered off) instead of deleting it, so we can do a forensic investigation. The intrusion response will only run in Tower; the Tower Template will override the variable value for the folder. 

This PR:

- Sets a condition on the task that deletes the old VM after replacement, so we only delete VMs from the ToBeDeleted folders
- Renames the old VM using a timestamp instead of hardcoded text

We have added a folder in vSphere called `Forensics` to quarantine VMs that have experienced intrusions.


VickieKarasic (approved):
These changes make sense to me, @acozine! I'll let others review as well.

kayiwa (approved):
OK! Thanks @acozine

PR #6112, opened 2025-04-30
[bibdata] Install clang

sandbergja:
Our library_standard_numbers gem depends on a rust crate named rb-sys, which depends on clang.

See https://ansible-tower.princeton.edu/#/jobs/playbook/18192/output

This is a prerequisite to https://github.com/pulibrary/bibdata/pull/2698


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6113, opened 2025-04-30
[pulfalight] Use rails_app instead of blacklight_app

hackartisan:
refs pulibrary/pulfalight#1250

modeled on #4040


acozine:
I think `lae` is the only other project that uses the `blacklight_app` role - could we switch it too and get rid of the role entirely?

Issue #6083, opened 2025-04-24
[CheckMK] Upgrade to CheckMK to 2.3.p30


acozine:
CheckMK is giving us noisy warnings about VEEAM backup timestamps:

```
Service PROBLEM notification
Host: cdh-geniza2 (IP: 128.112.200.119)
Service: ESX Snapshots
State: WARNING
Additional Info
Snapshot with a creation time in future found. Please check your network time synchronisation.
Please take a look: @ansible
Check_MK notification: Thu Apr 24 01:41:14 EDT 2025
```

These alerts are caused by a bug in CheckMK.  Upgrading to at least 2.3p19 will resolve these warnings. The current 2.3x release is 2.3p30, and the release notes look fine, so let's upgrade to that. 

This is a blocker for migrating to production infrastructure for CheckMK.

## Acceptance Criteria

- [x] We select a version of CheckMK to install and update the CheckMK [group vars](https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/checkmk/shared.yml) to reflect the new version
- [x] All staging CheckMK sites are upgraded with the Ansible playbook and are running 2.3.p30
  - [x] /production
  - [x] /staging
  - [x] /forrestal
  - [x] /newsouth
  - [x] AWS/EC2
  - [x] GCP
- [x] All CheckMK agents are updated to the same version using the agent playbook
- [ ] Confirm that alerts about VEEAM backup timestamps disappear

aruiz1789:
Updating the Sites: 
- SCP the new checkmk 2.3.p30 installer to the checkmk servers.
- Stop the checkmk service
- Run the app install
- Run omd update {SITE_NAME}
- Start the checkmk service

'''
OMD[gcp]:~$ omd sites
SITE             VERSION          COMMENTS
gcp              2.3.0p30.cee     default version
'''

kayiwa:
Document how to do this on Windows. 

Possibly test with 2.3.p31(?)

acozine:
Partially addressed by #6114.

acozine:
If we test with p31, let's test using the checkmk_server.yml playbook. Also, should we move that playbook into the `utils` directory?

aruiz1789:
Forrestal and NewSouth servers have been updated to version p30.

aruiz1789:
Production has also been upgraded to version p30.

aruiz1789:
All production hosts have been updated to version 2.3.0p30

[...]

PR #6114, opened 2025-05-01
Enable automatic agent updates

acozine:
Related to #6083.

Now that we have updated the staging and production CheckMK sites to 2.3.0p30, baked a new agent with the matching agent version, and enabled secure automatic agent updates, we are ready to get resources registered for future agent updates.  We can also get rid of the local copy of the agent role that we were using while we figured out how to install the agent from the server on the paid version - that fix was #6013.

This PR:

- sets the agent and server versions and reverts to using the generic role from Ansible Galaxy
- removes the local copy of the agent role


kayiwa (approved):
OK! Thanks @acozine

PR #6120, opened 2025-05-02
Fix security playbook so the falcon configure role will run

### LINKED ISSUE #6117: Security playbook succeeds but falcon-sensor fails to start
### Expected behavior
I can run the security playbook and get a running install of CrowdStrike falcon-sensor.

### Actual behavior
I run the playbook and it installs `falcon-sensor`, but the service will not run.

### Error messages
Checking the status of the service gives us:

```
$ sudo systemctl status falcon-sensor
○ falcon-sensor.service - CrowdStrike Falcon Sensor
     Loaded: loaded (/lib/systemd/system/falcon-sensor.service; disabled; vendor preset: enabled)
     Active: inactive (dead)

xxxxxxx systemd[1]: Starting CrowdStrike Falcon Sensor...
xxxxxxx falconctl[625]: CID is not set. Use falconctl to set the CID
xxxxxxx falconctl[625]: ERROR: failed to process the option --cid
xxxxxxx falconctl[625]: Run 'falconctl -h' OR 'falconctl --help' for usage help
xxxxxxx systemd[1]: falcon-sensor.service: Control process exited, code=exited, status=1/FAILURE
xxxxxxx systemd[1]: falcon-sensor.service: Failed with result 'exit-code'.
xxxxxxx systemd[1]: Failed to start CrowdStrike Falcon Sensor.
```

### Steps to replicate
Spin up a new VM.
Run the security playbook against it.
SSH in and run `sudo systemctl status falcon-sensor`

The CID is available in `group_vars/crowdstrike/vault.yml`. You can manually set the CID with the command `sudo /opt/CrowdStrike/falconctl -s --cid=<our_CID>`.

### Impact of this bug
We are not gathering security information from falcon-sensor unless we manually update new servers to set the CID for falcon-sensor.

acozine:
Closes #6117.

Turned out that the service facts update in real time. So after the falcon-install role ran, the falcon-sensor service appeared in the list of service facts. That meant the condition "if falcon-sensor service doesn't exist" was no longer met, and the falcon-configure role wasn't running.

This PR solves the problem by checking for the falcon-sensor log file, which does not get created until the service is actually running. Then it runs the falcon-configure role if the log file does not exist. 

This should fix any machines that are currently broken.


kayiwa (approved):
OK! Thanks @acozine

PR #6107, opened 2025-04-29
updates Rapid7 package version

acozine:
OIT updated the version of Rapid7 we use.

This PR updates it on our end to fix the security playbook.

PR #6121, opened 2025-05-02
[Rapid7] Use 'latest' instead of hardcoded version

acozine:
Our security-install playbook started failing a few weeks ago when OIT upgraded the install package versions. They have now renamed the package files to use `latest` instead of including the version number. 

This works fine in our playbook except for the step that configures Rapid7 with the token - that needs the full path with the version number to run the script. However, we tested the playbook without that task on a new VM and OIT says it worked fine.

Leaving that task in the playbook for now, commented out, in case we see issues in the future. If we need to reinstate it, we will need to add tasks that search the path and return the version number.

### DIFF
--- a/playbooks/utils/security_theater.yml
+++ b/playbooks/utils/security_theater.yml
@@ -112,23 +112,23 @@
     block:
       - name: Download the Rapid7 deb file (Ubuntu)
         ansible.builtin.get_url:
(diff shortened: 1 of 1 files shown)


acozine:
Related to #6107.

kayiwa (approved):
OK. Thanks @acozine

PR #6125, opened 2025-05-06
Cicognara wall

kayiwa:
cicognara is getting bombarded with bots.


acozine:
Cicognara is running fine today. We think either the bots have passed it by, or the issues we saw had a different root cause. If we want to implement Traefik for Cicognara in the future, we would need to make the changes in this PR, update nomad/traefik-wall/deploy/bot-plugin-production.tpl.yml to list the paths we want to defend, and add the site to the cloudflare turnstile configuration.

PR #6132, opened 2025-05-06
[rust] Allow deploy user to run rust binaries

sandbergja:
@christinach and I did this work, but then I forgot to commit it as part of #6129 -- sorry!


christinach (approved):
Thank you @sandbergja

Issue #6126, opened 2025-05-06
Update Datadog usage


acozine:
### What maintenance needs to be done?
We want to make sure our usage of Datadog is giving us value. We can fine-tune what logs we send and which hosts we track in Datadog.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
We have added a lot of stuff to our Datadog platform. Not all of it is useful. Some was useful for a period of time, but can be removed now.

## Acceptance Criteria

We're doing option 1 to start.

### Host Removal

-   [x] **Remove Staging and Low Impact Hosts**
    * byzantine-prod1/2
    * lae-prod1/2
    * lib-approvals-prod1/2
    * lib-static-prod1
    * lockers-and-study-spaces-prod1/2
    * [orcid-staging1/2](https://github.com/pulibrary/orcid_princeton/issues/460)
    * [pdc-describe-redis-staging1](https://github.com/pulibrary/pdc_describe/issues/2127)
    * [pdc-describe-staging1/2](https://github.com/pulibrary/pdc_describe/issues/2127)
    * [pdc-discovery-staging1/2](https://github.com/pulibrary/pdc_discovery/issues/800) *(Note: These also have APM, resulting in an additional $72/mo cost)*
    * slavery-prod1/2
    * static-prod1/2
    * [tigerdata-staging1/2](https://github.com/pulibrary/tigerdata-app/issues/1460) *(Note: These also have APM, resulting in an additional $72/mo cost)*

-   [x] **Remove Non-Application, CheckMK Monitored Infrastructure Hosts**
    * adc-prod1/2
    * lib-postgres-prod1/2
    * lib-solr-prod1/2/3 *(Note: We would lose the Solr health dashboard)*
    * lib-solr-prod7/8/9 *(Note: We would lose the Solr health dashboard)*
    * lib-zk-prod1/2/3
    * lib-zk4/5/6
    * libserv14.princeton.edu
    * mysql-db-prod1
    * [pdc-describe-redis-prod1](https://github.com/pulibrary/pdc_describe/issues/2127)

### Logging

-   [x] Filter Solr requests from ADC logs.
-   [x] Stop aggregating Solr logs.
-   [ ] Condense PDC Describe logging.
-   [ ] Stop logging all `WARN` messages.
-   [ ] Stop logging all `DEBUG` messages.

tpendragon:
## Background

[...]

PR #6133, opened 2025-05-06
[Datadog] Disable load balancer logs for Solr.

tpendragon:
Work towards #6126


kayiwa (approved):
OK! Thanks @tpendragon

PR #6145, opened 2025-05-09
[common] add dua cli

### LINKED ISSUE #6144: Feature Request: Interactive Mode for Iterative Disk Cleanup
**User Story**

As a user with a nearly full hard drive, I want to use [dua-cli'](https://github.com/Byron/dua-cli)s interactive mode to quickly identify and remove large, unnecessary files and directories in an iterative process, so that I can efficiently free up disk space without resorting to slow and cumbersome manual commands like du -hS * or find ... | xargs du.

**Problem Description**

When my hard drive is getting uncomfortably full, freeing up space is a priority. My current workflow often involves manually running `du -hS *` in various directories, identifying large culprits, and then using rm to delete them. This is a slow, repetitive, and often frustrating process, requiring me to constantly re-run commands and lose context as I navigate the file system. Tools like `find ... | xargs du` can help, but they still present a flat output that makes hierarchical exploration and targeted deletion difficult.

`dua-cli'`s interactive mode (`dua i`) is a significant improvement, providing a fast and visual way to explore disk usage and mark items for deletion. However, in a typical disk cleanup scenario, this is often an iterative process. I might delete some large files, and then want to re-evaluate the remaining space in certain directories to find the next biggest space hogs. Currently, while `dua i` allows deletion, the workflow for subsequent rounds of cleanup within the same session or easily picking up where I left off for a more in-depth dive into a specific large directory could be smoother.

**Proposed Enhancement**

Enhance the interactive mode of `dua-cli` to better support an iterative disk cleanup workflow. This could involve one or more of the following:

- Improved Re-scanning/Refreshing: Provide a more seamless way to re-scan the currently viewed directory or a selected subdirectory within the interactive session to reflect the space freed by deletions and update the size rankings. While a full restart is possible, a targeted refresh would maintain context and speed up the iterative process.
- Persistent Selection/Navigation: Allow users to "bookmark" or easily return to specific directories they were investigating, even after performing deletions or briefly navigating away.
- Filtering and Sorting Enhancements: While sorting by size is crucial, additional filtering options (e.g., hide files below a certain size, filter by modification date) could help narrow down targets for deletion in subsequent passes.
- Cumulative Space Freed Counter: Displaying the total amount of space freed during the current interactive session would provide positive reinforcement and help users track their progress.

**Benefits**

Implementing these enhancements would significantly improve the user experience for anyone using `dua-cli` to clean up a full hard drive. It would make the iterative process of identifying and removing unnecessary data much faster, more intuitive, and less prone to the manual overhead required with traditional command-line tools. This aligns well with `dua-cli`'s goal of being a fast and convenient disk usage analyzer.

kayiwa:
closes #6144
add the dua cli is an interactive way to look for large files on a system.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -1,8 +1,8 @@
 ---
 # defaults file for roles/common
-ansible_distribution_release: bionic
+ansible_distribution_release: jammy
+dua_install_tag: "v2.30.0"
 duf_version: "0.8.1"
-dust_version: "0.9.0"
 fd_find_version: "9.0.0"
 configured_dependencies: []
 common_shared_packages:
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -72,12 +72,39 @@
     name: '{{ configured_dependencies }}'
     state: present
 
-- name: Common | install dust
+- name: Common | remove dust
   ansible.builtin.apt:
-    deb: "{{ dust_url }}/v{{ dust_version }}/du-dust_{{ dust_version }}-1_amd64.deb"
-  when:
-    - running_on_server
-    - ansible_os_family == "Debian"
+    name: dust
+    state: absent
+
+- name: Common | Ensure system installation directory exists (e.g., /usr/local/bin)
+  ansible.builtin.file:
+    path: "{{ system_install_dir }}"
+    state: directory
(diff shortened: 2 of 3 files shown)


kayiwa:
Can we tag this to a commit (if only for idempotency)

Issue #6119, opened 2025-05-02
Ansible Role: Install and Configure Grafana with Nginx and LDAP


kayiwa:
Create an Ansible role that automates the installation and configuration of Grafana, including setting up an Nginx reverse proxy for access and configuring LDAP authentication with automatic assignment of users in a specific group to Grafana administrator privileges.

### Detailed Requirements:

- **Grafana Installation:**

  - Install Grafana on the target host.
  - Ensure Grafana is configured to run as a service.

- **Loadbalancer:**

  - Install and configure Nginx to act as a reverse proxy for Grafana.
  - Configure Nginx for secure access (including SSL/TLS).
  - Configure appropriate Nginx settings (e.g., buffer sizes, timeouts).

- **LDAP Authentication:**

  - Configure Grafana to authenticate users via LDAP.
  - Make this configurable:
    - LDAP server URL.
    - LDAP bind DN and password.
    - LDAP base DN for user search.
    - LDAP user search filter.
    - LDAP group search filter.
    - LDAP group mapping to Grafana roles.

- **Automatic Administrator Assignment:**

  - Automatically assign users who are members of a specified LDAP group to the "Admin" role in Grafana.
  - This should be configurable so that the administrator group can be easily changed.

- **Role Configuration:**

  - The role should be highly configurable through Ansible variables (defined in defaults/main.yml and vars/main.yml).
  - Provide sensible defaults for all configuration options.
  - Include clear documentation within the role (README.md) for all variables and usage.

### Tasks:

- [ ] create a host-vm pulemetry
- [ ] Create the Ansible role structure.
- [ ] Implement tasks for Grafana installation.
- [ ] Implement tasks for Nginx installation and configuration.
- [ ] Implement tasks for Grafana LDAP configuration.
- [ ] Define all necessary variables in defaults/main.yml and vars/main.yml.
- [ ] Create a comprehensive README.md file.

### Acceptance Criteria:

- [ ] The Ansible role installs and configures Grafana as specified.
- [ ] Nginx is correctly set up as a reverse proxy.
- [ ] LDAP authentication functions as expected.
- [ ] Users in the specified LDAP group are automatically assigned Grafana administrator privileges.
- [ ] The role is well-documented and idempotent
- [ ] The role works on the targeted vm(s)

### Additional Notes:

Include example playbooks in the role documentation.
Consider Grafana provisioning for datasources and dashboards (?)

PR #6149, opened 2025-05-09
[prancible] Create new role for pulemetry

VickieKarasic:
Also set up molecule and tasks

Related to #6119


kayiwa:
superseded by #6315

PR #6155, opened 2025-05-12
[Nomad] Simplify pul nomad to make it much easier to contribute

tpendragon:
## Goal

Reduce the cognitive load and improve the maintainability of the `pul_nomad` role. This PR is successful if other members of IT feel comfortable navigating this role.

## Steps & Reasoning

- [x] Remove the imported `nomad` and `consul` roles. Implement them in our code with the same output (so they can be run on our machines.)
    * These roles are complex because they support so many different operating systems and configuration via variables. Installing Nomad/Consul isn't that much more complex than a normal package - the role will be easier to understand it if we just do it.
- [x] Add documentation about what each part of the cluster does, where to find information about Nomad/Consul/DNSMasq/etc, and references for the full configuration files.
    * It's hard to navigate the web to find the config reference for these.
- [x] Add documentation about common maintenance tasks like updating nomad/consul
- [x] Make this role idempotent
- [x] Use `community.general` modules when possible to simplify complex `register/check/act` loops.
    * It's complex to understand how running a command, registering a variable, and then checking that variable leads to idempotence. Collapsing that understanding into a module someone else maintains will make it easier to understand.
- [x] Add in code documentation for complex things like when using `import_tasks` vs `include_tasks.`
- [x] Separate `server` and `client` tasks
    * Although this will lead to some duplication, it's what we're doing everywhere else, so it should make it easier to understand this role.
- [x] Prefer hard-coded configurations rather than variables when possible and when things won't vary in our environment.
- [x] Reduce the number of ACL tokens we have to register.
- [x] Document how ACL works and why we do it the way we do, point to code in the role where it's happening.
- [x] Make it possible to run this role on only one node, rather than the whole cluster at a time.
- [x] Maybe adjust how we push application secrets. A different playbook? A tag? Unsure, could use feedback.
- [x] Maybe expand tests?
- [x] Add documentation for running a local cluster using `molecule`
- [x] ??

## Reviewing Notes

This PR has a lot going on, and there were only a few people who understood how this whole role came together in the first place - thus the goal here is to make it easier.

My recommendation for reviewing:

1. Read the documentation at https://github.com/pulibrary/princeton_ansible/blob/simplify_pul_nomad/roles/pul_nomad/README.md - is it clear and helpful?
2. Start at https://github.com/pulibrary/princeton_ansible/blob/simplify_pul_nomad/roles/pul_nomad/tasks/main.yml and process through it - does it make sense, given the context of the README, what's happening and when?

If you want to see the diff, it's here (without the role deletions): https://github.com/pulibrary/princeton_ansible/pull/6155/commits/f2d0d3747d9e460791a4fe1fb901ca6865a121fa. That being said, the changes are significant, it's probably easier to review this as if it's a new role.

This role has been proven to be backwards compatible - it's what's running on the cluster now.

### DIFF
--- a/roles/pul_nomad/README.md
+++ b/roles/pul_nomad/README.md
@@ -64,20 +64,19 @@ Variables have defaults in [`defaults/main.yml`](defaults/main.yml) and may be d
 
 Consul is a service registration system with a DNS API.
 
-What this means is that you can register services in it, like
-`128.4.1.2 on port 8216 is a postgres database`, and then other services can
-configure their DB host to `postgres-database.service.consul` and consul will
-route it to `128.4.1.2:8216`.
+Nomad uses it to register services. This means when you deploy an application like Figgy and it starts a container on port 94182, Nomad tells Consul "Figgy is running on port 94182 on machine nomad-client-staging1.lib.princeton.edu, call it figgy-staging", which lets other machines in the cluster contact any instance it has registered as "figgy-staging" by contacting "figgy-staging.consul.service" as if it was a URL.
 
 Multiple services can be registered to the same name and Consul will load
 balance them, and consul can health check the services to make sure it doesn't
-route anything to services that are broken.
+route anything to services that are broken. For example, if one figgy container broke, then Consul won't forward figgy-staging.consul.service to it.
+
+We use Consul with Nomad, because our nginx load balancer can query Consul for machines for a specific service and automatically adjust when Nomad scales a service without reconfiguring nginx.
 
 ### Server Types
 
 Consul has "server" software and "client" software.
 
-A cluster consists of 3 server instances, one on each `nomad-host-*.lib.princeton.edu` VM, and a client software install on every client VM (`nomad-client-*.lib.princeton.edu`)
+Our cluster consists of 3 server instances, one on each `nomad-host-*.lib.princeton.edu` VM, and a client software install on every client VM (`nomad-client-*.lib.princeton.edu`)
 
 ### Configuration
 


roles/pul_nomad/README.md
```diff
@@ -1,36 +1,224 @@
...
+
+### Consul Interaction
+
+Nomad uses Consul for clustering (the Nomad servers find each other by querying
+Consul) and service discovery (when you launch a job in Nomad it can register a
+service in Consul that we can point to from our load balancers.)
+
+### Host Volumes
+
+Host volumes are a way for containers to save data to the underlying client. It's a folder on the client which the container maps - if a job requires a certain host volume it will only ever get scheduled on a node which has that volume.
```
acozine, line 162:
```suggestion
Host volumes are a way for containers to save data to the underlying client VM. It's a folder on the client VM which the container maps - if a job requires a certain host volume it will only ever get scheduled on a node which has that volume.
```

acozine (commented):
Some of the stuff we talked about.

tpendragon:
@acozine Did a bunch of re-wording in the README, wonder if that's clearer now?

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+Nomad is a container orchestrator.
+
+What this means is you can write job files, tell Nomad to deploy them, and Nomad
+will automatically find a client VM in the cluster that can run them. If that
+client goes down it will launch the container on a different client.
+
+We install Podman on our Nomad client VMs so they can run containers without us
+having to rely on Docker.
+
+### Server Types
```
acozine, line 123:
```suggestion
### VM Types
```

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+
+### Consul Interaction
+
+Nomad uses Consul for clustering (the Nomad server installs find each other by querying
+Consul) and service discovery (when you launch a job in Nomad it can register a
+service in Consul that we can point to from our load balancers.)
+
+### Host Volumes
+
+Host volumes are a way for containers to save data to the underlying client. It's a folder on the client which the container maps - if a job requires a certain host volume it will only ever get scheduled on a node which has that volume. Most containers have no need for persistent data on disk - they store that data in a database or something similar, but if we run a database or something like that in Nomad it might need a host volume. In those cases, we must ensure high availability through some other means (e.g "cloud" mode on Solr, clusters in Postgres, HA Mode in Redis, etc.)
```
acozine, line 158:
```suggestion
Host volumes are a way for containers to save data to the underlying client VM. Most containers have no need for persistent data on disk - they store that data in a database or something similar, but if we run a database or something like that in Nomad it might need a host volume. A host volume is a folder on the client which the container maps - if a job (service) requires a certain host volume it will only ever get scheduled on a node which has that volume. In those cases, because we are tying the job (service) to a specific node, we must ensure high availability through some other means (e.g "cloud" mode on Solr, clusters in Postgres, HA Mode in Redis, etc.)
```

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+
+Nomad has a useful [tutorial](https://developer.hashicorp.com/nomad/tutorials/stateful-workloads/stateful-workloads-host-volumes) about how they work.
+
+More recent versions of Nomad have "dynamic host volumes" which would allow us to create these on the fly, not in Ansible, but we're not upgraded enough for that. Nomad has a [tutorial](https://developer.hashicorp.com/nomad/tutorials/stateful-workloads/stateful-workloads-dynamic-host-volumes) for those.
+
+### Application Configurations
+
+Containers deployed to the Nomad cluster may still need external systems provisioned and variables persisted to Nomad. For example an application may need a postgres database set up on our cluster, secrets installed into Nomad (which are currently in Ansible Vault), or other configuration done.
+
+Those things happen in [`tasks/application_configurations.yml`](tasks/application_configurations.yml).
```
acozine, line 168:
```suggestion
Those things happen in [`tasks/application_configurations.yml`](tasks/application_configurations.yml). Each application we deploy to Nomad has its own configuration tasks file included here.
```

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+
+This pattern may change in the future. If you want to run a playbook without running these application configurations you can do: `ansible-playbook playbooks/nomad.yml --skip-tags app_config`. You may want to do that if you're just hoping to upgrade a client or server install, change the Nomad/Consul configuration, or other tasks that are about the cluster and not the applications on it.
+
+Similarly, to update a single app's configuration you can do something like `ansible-playbook playbooks/nomad.yml --tags dpulc`
```
acozine, line 172:
```suggestion
Similarly, to update a single app's configuration you can do something like `ansible-playbook playbooks/nomad.yml --tags dpulc`, because the include tasks all have tags on them.
```

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+
+1. **DNS Token (global)**: Consul's [DNS service](https://developer.hashicorp.com/consul/docs/discover/dns) uses this token to get service information from the cluster. If this is wrong or unset, `dig @127.0.0.1 -p 8600 consul.service.consul ANY` will return no answers.
+1. **Agent Token (local)**: Consul's agent (the process run via `sudo service consul start`, uses this to communicate with the rest of the cluster and synchronize services. This uses a "[node identity](https://developer.hashicorp.com/consul/docs/secure/acl/role#node-identities)" token, which has a default policy.
+1. **Nomad Server Token (global)**: Allows Nomad's server installs to mesh, manage ACLs, and update Consul services. Every Nomad Server VM has this configured. Defined in [nomad/configure_server.yml](tasks/nomad/configure_server.yml)
+1. **Nomad Client Token (global)**: Allows Nomad's client installs to update Consul services as they're deployed into Nomad. Every Nomad Client VM install has this configured. Defined in [nomad/configure_client.yml](tasks/nomad/configure_client.yml)
```
acozine, line 196:
```suggestion
1. **Nomad Server Token (global)**: Nomad's server uses this token to talk to Consul to learn about each other's existence, manage ACLs, and update Consul services. Every Nomad Server VM has this configured. Defined in [nomad/configure_server.yml](tasks/nomad/configure_server.yml)
1. **Nomad Client Token (global)**: Nomad's client uses this token to update Consul services as they're deployed into Nomad. Every Nomad Client VM install has this configured. Defined in [nomad/configure_client.yml](tasks/nomad/configure_client.yml)
```

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,220 @@
...
+
+### Nomad Tokens
+
+We don't currently have any special nomad tokens, just the management token. We create that token when we [bootstrap](https://developer.hashicorp.com/nomad/tutorials/access-control/access-control-bootstrap) Nomad's ACL system in [nomad/server.yml](tasks/nomad/server.yml)
+
+# Generating Variables
+
+Some variables need to be generated and stored in the vault. To generate new
+ones for consul (`consul_gossip_encryption_key`) you can run:
```
acozine, line 205:
```suggestion
# Refreshing the Gossip Encryption Key

To refresh the gossip encryption key for consul (`consul_gossip_encryption_key`), generate a new key and save it to the vault. To generate the key, run:
```

acozine (commented):
Stage two achieved.

roles/pul_nomad/README.md
```diff
@@ -1,36 +1,219 @@
...
 
-Author Information
-------------------
+## Consul
 
-An optional section for the role authors to include contact information, or a website (HTML is not allowed).
+Consul is a service registration system with a DNS API.
+
+What this means is that you can register services in it, like
+`128.4.1.2 on port 8216 is a postgres database`, and then other services can
```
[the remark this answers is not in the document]
hackartisan: You could, for example, register the ip `128.4.1.2` and port `8216` as a postgres database, and then...

PR #6157, opened 2025-05-13
Remove DataDog from Orcid and PDC Discovery

bess:
Ref https://github.com/pulibrary/pdc_discovery/issues/800
Ref https://github.com/pulibrary/orcid_princeton/issues/460


kayiwa (approved):
OK! Thanks @bess

PR #6160, opened 2025-05-13
Francis' work to allow Solr8 and Solr9

acozine:
I reorganized the branch. @kayiwa did the actual work.

Supersedes #5896. 

This PR cleans up our Solr Cloud role, makes it easier to understand, and supports backups for Solr9 as well as Solr8.

Related to #5740

PR #6163, opened 2025-05-14
I5740 solr backup separate

kayiwa:
create a new solr9cloud role. As currently done this https://github.com/pulibrary/princeton_ansible/pull/6160 will not be able to run on solr8 without breaking them. 

we have successful backups on solr8 and solr9 

https://github.com/pulibrary/pul_solr/pull/472


.github/workflows/molecule_tests.yml
```diff
@@ -83,6 +83,7 @@ jobs:
           - sidekiq_worker
           - sneakers_worker
           # - solrcloud
```
tpendragon, line 85: Not testing solr8 anymore?

roles/solr9cloud/tasks/main.yml
```diff
@@ -0,0 +1,265 @@
...
+# Create required directories
+- name: Solrcloud | Create Solr directories
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    mode: "0755"
+    owner: "{{ solr_user }}"
+    group: "{{ solr_group }}"
+  loop:
+    - "{{ solr_data_home }}"         # /solr
```
tpendragon, line 50: Is there a strength to these being configurable? Can we just hardcode those?

roles/solr9cloud/tasks/main.yml
```diff
@@ -0,0 +1,265 @@
...
+  ansible.builtin.include_tasks: backup_mounts.yml
+
+- name: Solrcloud | Check if Solr is already installed
+  ansible.builtin.stat:
+    path: "{{ solr_versioned_dir }}/README.txt"
+  register: solr_install_check
+  tags: [download, install]
+
+# Download SolrCloud package
+- name: Solrcloud | Download SolrCloud archive
```
tpendragon, line 79: I wonder if we can extract this to another file? Like `install.yml`

roles/solr9cloud/tasks/main.yml
```diff
@@ -0,0 +1,265 @@
...
+
+# Configure SolrCloud environment with idempotence check
+- name: Solrcloud | Check if init script exists
+  ansible.builtin.stat:
+    path: /etc/default/solr.in.sh
+  register: init_file_check
+  tags: [configure]
+
+# Configure SolrCloud environment
+- name: Solrcloud | Configure SolrCloud init script
```
tpendragon, line 177: Maybe extract this stuff to a `configure.yml`?

roles/solr9cloud/tasks/service.yml
```diff
@@ -0,0 +1,99 @@
...
+  tags: [service]
+
+# If the service file doesn't exist, create it
+- name: Solrcloud | Create initial Solr systemd service file
+  ansible.builtin.template:
+    src: solr.service.j2
+    dest: /etc/systemd/system/solr.service
+    mode: "0644"
+  notify: Reload systemd and restart solr
+  when: not service_file_check.stat.exists
```
tpendragon, line 16: Question: Why are these checks necessary? Isn't `template` already idempotent?

roles/solr9cloud/templates/solr.in.sh.j2
```diff
@@ -0,0 +1,39 @@
+# roles/solrcloud/templates/solr.in.sh.j2
+{{ ansible_managed | comment }}
+# Local customizations for Solr
+
+# Memory settings
```
tpendragon, line 5: I wonder if we can just hardcode a bunch of these? Do we have significant differences for these variables in our cluster?

tpendragon (approved):

.github/workflows/molecule_tests.yml
kayiwa: we never have. :-(

PR #6171, opened 2025-05-15
Make Checkmk vars scalable

acozine:
Originally our CheckMK architecture incorporated cloud resources in our `staging` and `production` sites. We now have CheckMK sites running in the cloud specifically for cloud infrastructure. This meant that the checkMK automation could not re-use our `runtime_env` variable the way we re-use it elsewhere.

This PR:

- adds files to `group_vars/checkmk` for AWS and GCP
- updates the `staging`, `shared` and `production` group vars files for values that are no longer shared
- updates the vault file to include credentials for the AWS and GCP sites
- updates the playbook to load vars files based on a new variable called `checkmk_service`
- fixes a typo in inventory that was making `prds-dataspace-staging-endpoint1` unreachable


acozine:
Note: this branch will not work fully in Tower - the changes to inventory need to be merged to main and imported into Tower before the Tower jobs will pick up all the variables set in inventory.

aruiz1789 (approved):

PR #6176, opened 2025-05-16
Use OpenJpeg for Cantaloupe.

tpendragon:
We're getting rid of Cantaloupe soon, and so switching this out during
the migration will mean we don't have to pay the Kakadu license.


kayiwa (approved):
OK! Thanks @tpendragon

PR #6178, opened 2025-05-19
Add reservoir staging machines to inventory

Beck-Davis:
Add reservoir-staging{1,2}.lib.princeton.edu to princeton_ansible inventory

Closes #6173

PR #6179, opened 2025-05-19
Correct FQDN for reservoir-staging VM

Beck-Davis:
Corrects the FQDN for reservoir-staging{1,2} vms added with #6178


VickieKarasic (approved):
Thank you @Beck-Davis  :)

Issue #6084, opened 2025-04-24
[CheckMK] Migrate to production infrastructure


acozine:
Follows up on #5725.

Now that our distributed CheckMK architecture in the staging vSphere environment is working well, we should migrate to the production environment.

## Acceptance Criteria
- [ ] We create `pulmonitor-prod1` and pulmonitor-prod2` VMs in the production vSphere with the same resources as the `pulmonitor-staging1/2` VMs
- [ ] We build CheckMK on those VMs to match the current staging architecture and versions - `prod1` will host the `/production` site and the main integrated view; `prod2` will host the `/staging` site
- [ ] We create `pulmonitor.princeton.edu` as the production URL (DNS, certs, nginxplus config)
- [ ] We integrate all CheckMK sites (/staging, /forrestal, /newsouth, AWS, GCP) into the main site
- [ ] We back up the content of the staging VMs and restore to the production VMs
- [ ] We tidy up any leftover artifacts (DNS for `pulcheck*`, the old `pulmonitor` VM, firewall rules, nginxplus config, etc.)

acozine:
Related to (and blocked by) #6083

PR #6196, opened 2025-05-27
[checkmk] not have production and staging paths

kayiwa:
add new server block for staging and production with location additions on the TLS server blocks

related #6084


acozine:
Per today's discussion, the added `-service` is necessary because we still have upstreams for `pulmonitor-staging` and `pulmonitor-production` defined in the checkmk staging configuration for nginxplus.

acozine (approved):

Issue #6199, opened 2025-05-29
Rotate all Configured Alma API Keys


kevinreiss:
### What maintenance needs to be done?
In order to comply with the Alma Internal Audit Recommendations we need to rotate all Alma API Keys in use by PUL applications at least once a year. 

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Relevant deadlines
- [ ] Audit Compliance Deadline Date 06/27/2025

### Why is this maintenance needed?
- We need to comply with audit recommendations.
 
### Acceptance criteria
 - [x] Orangelight staging key rotated
 - [x] Orangelight production key rotated
 - [x] Bibdata staging key rotated
 - [x] Bibdata production key rotated
 - [x] Lib-jobs staging key rotated
 - [x] Lib-jobs production key rotated
 - [x] Abid staging key rotated
 - [x] Abid production key rotated

### Notes
Follow this [procedure](https://pul-confluence.atlassian.net/wiki/spaces/ALMA/pages/149979159/Alma+API+Key+Management).

kevinreiss:
@mzelesky I believe all these have been now rotated and deployed live. Thanks @sandbergja, @regineheberlein, and @christinach.

PR #6203, opened 2025-05-30
Rotate alma staging api key

sandbergja:
This PR removes an old Alma API Key named "Blacklight Sandbox View", which was used in lib-jobs staging and bibdata staging.
It replaces it with two new keys (one for lib-jobs staging, a separate key for bibdata staging) [according to this procedure](https://pul-confluence.atlassian.net/wiki/spaces/ALMA/pages/149979159/Alma+API+Key+Management).

Part of #6199


christinach (approved):
Thanks @mzelesky and @sandbergja !

PR #6209, opened 2025-06-02
[Datadog] Skip logging load balancer 302s

tpendragon:
MOST of the 302s are for bots redirecting to Traefik.

This is a visibility loss for sure, but we're ingesting way too many logs right now. Hopefully we can switch to Loki soon. This should cut our log ingestion in half.


kevinreiss (approved):

carolyncole (approved):
Thanks @tpendragon

PR #6221, opened 2025-06-04
Add ci in the by_environment directory

Beck-Davis:
Since we have ci machines that are separate from production, staging, and qa, adding a new file under by_environment for ci machines.


kayiwa (approved):
OK! Thanks @Beck-Davis

PR #6239, opened 2025-06-16
unbreak main branch.

kayiwa:
the dpul role depends on openjdk. we need a container that can allows java to run. we also lower the version of requests from dependabot which allows us to run molecule locally


jrgriffiniii (approved):

hackartisan (approved):
There is a declared dependency on openjdk, but I'm not sure why. I can't think of a reason dpul would need java. Regardless, I'm in favor of a passing build.

Issue #6255, opened 2025-06-18
[Nginxplus] Deploy New NGINX Plus License & Certificate to Production Load Balancers


kayiwa:
A new NGINX Plus license key and SSL certificate must be deployed immediately to all production load balancers to prevent service disruption. The current license expired June 16th.

**Required Artifacts**

- **New NGINX Plus License Key:** 
  - File: nginx-plus-license-2025.key (Found at F5.com)

**- New SSL Certificate:**
  - Files: wildcard_cert_2025.crt
  
**- New JSON Web-Token:**
  - Files: wildcard_cert_2025.crt

**Confirm license/cert validity:**

```bash
nginx -t -c /etc/nginx/nginx.conf  # Validate config syntax  
sudo nginx -V | grep -E "(license|cert)" # Verify current versions 
```

kayiwa:
While this ticket is open we will see this error

```bash
pulsys@adc-prod2:~$ sudo apt -y update
Hit:1 http://security.ubuntu.com/ubuntu jammy-security InRelease
Ign:2 https://apt.datadoghq.com stable InRelease
Ign:3 https://apt.vector.dev stable InRelease
Hit:4 https://apt.datadoghq.com stable Release
Hit:5 https://apt.vector.dev stable Release
Hit:6 http://us.archive.ubuntu.com/ubuntu jammy InRelease
Hit:7 http://us.archive.ubuntu.com/ubuntu jammy-updates InRelease
Hit:8 http://us.archive.ubuntu.com/ubuntu jammy-backports InRelease
Get:9 http://nginx.org/packages/ubuntu jammy InRelease [3,581 B]
Err:12 https://pkgs.nginx.com/app-protect/ubuntu jammy InRelease
  400  Bad Request [IP: 18.198.112.44 443]
Get:13 http://nginx.org/packages/ubuntu jammy/nginx amd64 Packages [30.4 kB]
Err:14 https://pkgs.nginx.com/app-protect-security-updates/ubuntu jammy InRelease
  400  Bad Request [IP: 18.198.112.44 443]
Err:15 https://pkgs.nginx.com/plus/ubuntu jammy InRelease
  400  Bad Request [IP: 18.198.112.44 443]
Reading package lists... Done
W: http://nginx.org/packages/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
[...]

PR #6257, opened 2025-06-18
[nginxplus] add new replacement keys

kayiwa:
we need a new crt, key and jwt file
this enables nginxplus to update

related to #6255


acozine:
These were not the keys we were looking for.

Issue #6231, opened 2025-06-12
[Log Rotation] templatize the logrotation of our VMs


kayiwa:
Currently, log rotation settings are managed inconsistently across different roles. This leads to duplication of effort, potential for configuration drift, and difficulty in maintaining a standardized logging policy across the infrastructure.

Create logrotation tasks in the common role that will be responsible for managing the configuration of the `logrotate` utility on target systems. The core work should be a Jinja2 template for for individual configuration files in `/etc/logrotate.d/`.

The role should be highly configurable using variables, with sensible defaults. These variables should be definable at different levels of the Ansible inventory hierarchy (e.g., `group_vars/all`, `group_vars/<group_name>`, `host_vars/<hostname>`) to allow for maximum flexibility.

related to #6230

PR #6260, opened 2025-06-18
[Common]  logrotate

kayiwa:
use the common role for determining how logrotation will happen throughout PUL
related to #6231


roles/common/README.md
```diff
@@ -1,41 +1,151 @@
...
-The are packages that are generally expected on a server that Ubuntu doesn't
-always have on by default.
+Install all the tools our VMS have in Common
 
-Requirements
-------------
+## Overview
 
-None
+Since we include logrotate in the role, we want other roles to have access to a highly flexible, hierarchical approach to managing log rotation across your infrastructure. It allows you to define logrotate rules at multiple levels of your Ansible inventory, from global defaults to host-specific configurations.
```
acozine, line 7: This file is still the README for the common role, which does a lot more than log rotation. This Overview feels very tightly focused on one topic.

roles/common/README.md
```diff
@@ -1,41 +1,151 @@
...
-Dependencies
-------------
+Variables can be defined at multiple levels with the following precedence (highest to lowest):
 
-None
+1. **Host Variables** (`host_vars/hostname.yml`)
+2. **Group Variables** (`group_vars/group_name.yml`)
+3. **Role Variables** (`roles/role_name/vars/main.yml`)
+4. **Global Variables** (`group_vars/all.yml`)
+5. **Role Defaults** (`roles/role_name/defaults/main.yml`)
```
acozine, line 19: This may not be accurate, since we are not actually using group vars in the way Ansible designed them to be used. For example, we do not have a file called `group_vars/figgy_production.yml` - instead we have a directory called `group_vars/figgy` and within that we have various files . . . We typically include those variables as `vars_files`, so I'm not sure how Ansible would choose among multiple values in our case.

roles/common/templates/logrotate_rules.j2
```diff
@@ -0,0 +1,120 @@
...
+{{ path }}
+{% endfor %} {
+{% if item.options.daily is defined and item.options.daily %}
+    daily
+{% elif item.options.weekly is defined and item.options.weekly %}
+    weekly
+{% elif item.options.monthly is defined and item.options.monthly %}
+    monthly
+{% elif item.options.yearly is defined and item.options.yearly %}
+    yearly
```
acozine, line 14: I think we want to standardize much more than this. This approach allows too much variability - I don't want us to have any logs that rotate monthly or yearly.

roles/common/templates/logrotate_rules.j2
```diff
@@ -0,0 +1,120 @@
...
+{% endif %}
+{% if item.options.rotate is defined %}
+    rotate {{ item.options.rotate }}
+{% endif %}
+{% if item.options.size is defined %}
+    size {{ item.options.size }}
+{% endif %}
```
acozine, line 21: Here, too, we should standardize more. We want all logs to define a maximum size for rotation.

roles/common/templates/logrotate_rules.j2
```diff
@@ -0,0 +1,120 @@
...
+{% if item.options.nodateext is defined and item.options.nodateext %}
+    nodateext
+{% endif %}
+{% if item.options.dateformat is defined %}
+    dateformat {{ item.options.dateformat }}
+{% endif %}
+{% if item.options.extension is defined %}
+    extension {{ item.options.extension }}
+{% endif %}
```
acozine, line 75: Same thing with all of this stuff - we should define a standard format that we want all our rotated logs to use. Otherwise we will be back in the position we've been in, where some logs do one thing and others do something different.

acozine (changes requested):
Nice work building out the skeleton of a more coordinated approach to log rotation. This set of changes won't achieve our goals, though - it allows, even encourages, too much variation. 

We want to standardize our log rotation, which requires a much more opinionated (and shorter!) template. I put a few specific comments on bits of the PR - let's touch base about reworking the approach.

roles/common/templates/logrotate_rules.j2
kayiwa: but this is how logging works. If for example we define for the median... then suddenly our nginx load balancer is reloading loads more because we've decided this is our tolerance level to balance it out with a rails app that is less taxed. Or even the same nginx (passenger) configs will defer on a application web server vs a load balancer.

roles/common/molecule/default/group_vars/all.yml
```diff
@@ -0,0 +1,33 @@
+---
```
acozine, line 1: What is the difference between setting variables in the defaults for hte common role and setting them here?

roles/common/tasks/main.yml
```diff
@@ -67,6 +67,20 @@
...
     - reload logrotate timer settings
 
+- name: Common | Configure custom logrotate rules
+  ansible.builtin.template:
+    src: "logrotate_rules.j2"
+    dest: "/etc/logrotate.d/{{ item.name }}"
+    mode: "0644"
+    owner: root
+    group: root
+  loop: "{{ logrotate_rules }}"
```
acozine, line 77: Is this task going to be used to set log rotation for the logs that appear on every machine? Do we want the custom rules to happen in the specific roles (rails, solr, etc.) rather than here?

roles/common/templates/logrotate_rules.j2
```diff
@@ -0,0 +1,43 @@
...
+    dateformat -%Y%m%d-%s
+{% if item.options.sharedscripts is defined and item.options.sharedscripts %}
+    
+    # Shared scripts (only when explicitly needed)
+    sharedscripts
+{% endif %}
+{% if item.options.postrotate is defined and item.options.postrotate != '' %}
+    
+    # Post-rotation commands
+    postrotate
```
acozine, line 39: This probably needs to be inside the `if` statement, so we don't end up with stray `postrotate` lines in rotation files that don't need them.

roles/common/handlers/main.yml
```diff
@@ -8,3 +9,11 @@
     name: vector
     state: restarted
   when: running_on_server
+
+- name: test logrotate configuration
+  ansible.builtin.command: logrotate -d /etc/logrotate.conf
+  register: logrotate_test
+  failed_when: logrotate_test.rc != 0
+  changed_when: false
```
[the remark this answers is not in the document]
kayiwa: I am pretty sure idempotency failed but that may have been my linter

Issue #6254, opened 2025-06-18
[Log Rotation] Add CheckMK monitoring for abnormal growth of logs / size of logs


acozine:
If a log that usually writes 20 mb per day starts writing 30 mb per hour, we want to know about it before log rotation kicks in.

Set up monitoring for the logs we rotate so we will now if/when a log starts growing at an unusual pace.

PR #6273, opened 2025-06-20
add a logmonitor check

kayiwa:
we are creating a role because this check requires variables. As currently implemented there is no place to pass parameters to monitor on local scripts.

related to #6254

### DIFF
--- a/playbooks/utils/checkmk_add_logmonitor.yml
+++ b/playbooks/utils/checkmk_add_logmonitor.yml
@@ -1,0 +1,34 @@
+---
+# To run the playbook:
+#  will run with the staging runtime environment and the rails rule group on pdc-describe-staging2
+
+- name: Install CheckMk local check scripts on host
+  hosts: "{{ runtime_env | default ('staging') }}"
+  remote_user: pulsys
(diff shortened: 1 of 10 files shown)


kayiwa:
I considered but since we are documenting inline... it because hard(er) for me to make this useful. In fact this is the reason I imagined the value of making this a role. 🤷🏿‍♂️

acozine:
Hmmmm. Thinking out loud here:

The outcome we want is that for each log we have two files on the server: a log rotation definition file and a check in CheckMK that looks to see if it's growing at a rate we're not happy about.

Can we have (or at least start with) static values for the interval, the warning level, and the critical level? If so, then the only two things we need to pass variables for to make your script work are the name of the log file and the name of the statfile, both with absolute paths. We should be able to build the name of the statfile from the name of the log file (something like this: `log_growth_statefile: "/var/lib/check_mk_agent/state/{{ log_growth_logfile }}.size" - we probably have to massage the file name to eliminate the path), so it's basically only one variable, really.

So how do we pull values for "the files we care about" . . . .?

PR #6259, opened 2025-06-18
[tigerdata] remove maintenance window message from error page

leefaisonr:
ref https://github.com/pulibrary/tigerdata-app/issues/1555

PR #6278, opened 2025-06-23
[tigerdata] remove maintenance window message from error page

leefaisonr:
ref https://github.com/pulibrary/tigerdata-app/issues/1555


leefaisonr:
supercedes #6259

PR #6283, opened 2025-06-25
[prancible] add docsible

kayiwa:
allow autogeneration of documentation using [docsible](https://docsible.github.io/)


acozine:
This is for role documentation. Not sure how we will use this - a follow-up PR here or in GitLab to document how to document would be great, but working code wins.

PR #6289, opened 2025-06-27
[lib_svn] use common logrotate template

### LINKED ISSUE #6265: [log rotation] update lib_svn logrotation
update the lib_svn role to depend on the new [common role](https://github.com/pulibrary/princeton_ansible/tree/i6231_logrotate/roles/common) for logrotation

kayiwa:
remove the current svnserve template. Replace it with a template on staging and production servers
adjust the template to using the opinionated defaults with customized permissions

closes #6265


regineheberlein (approved):
Thanks @kayiwa

PR #6291, opened 2025-06-30
[offboarding max] - remove Max

maxkadel:
This hasn't been run on anything yet
The only one where it really matters is probably the deploy_user / pulsys playbook.

For the admin ones, I won't be able to log in without CAS, so shouldn't be a rush to run them.

Connected to https://github.com/pulibrary/dacs_handbook/issues/311


carolyncole:
Boo!

kayiwa:
more BOO!

PR #6294, opened 2025-06-30
[nginxplus] use common role template

### LINKED ISSUE #6267: [log rotation] update nginxplus role logrotation
update the nginxplus role to depend on the new [common role](https://github.com/pulibrary/princeton_ansible/tree/i6231_logrotate/roles/common) for logrotation

Additional criteria to look into. 
Make sure the number access.log. doesn't run over the selected configurations in the template.

kayiwa:
- **remove logrotate config template**
- **use the common role logrotate template**

closes #6267


VickieKarasic (approved):
Thanks @kayiwa

PR #6299, opened 2025-07-01
update logrotate for postgresql

aruiz1789:


kayiwa (approved):
OK! Thanks @aruiz1789

PR #6303, opened 2025-07-02
[nginxplus] update logrotate

kayiwa:
bypass the nginxplus license key to update logrotation PR <https://github.com/pulibrary/princeton_ansible/pull/6294>

```bash
-rw-r-----  1 nginx adm    165K Jul  1 12:11 hc.log.1
-rw-r-----  1 nginx adm    271K Jul  1 12:47 access.log.2.gz
-rw-r-----  1 nginx adm     12K Jul  1 13:04 access.log.1
-rw-r-----  1 nginx adm       0 Jul  2 00:00 hc.log
-rw-r-----  1 nginx adm       0 Jul  2 00:00 access.log
-rw-r-----  1 nginx adm    6.4M Jul  2 00:00 error.log.1
-rw-r-----  1 nginx adm     82K Jul  2 13:23 error.log-20250702-1751462631.gz
-rw-r--r--  1 nginx adm     12K Jul  2 13:25 error.log-20250702-1751462723
drwxr-xr-x  2 root  root   4.0K Jul  2 13:25 .
drwxrwxr-x 13 root  syslog 4.0K Jul  2 13:25 ..
-rw-r--r--  1 nginx adm    4.8K Jul  2 13:26 error.log
```

sample output of `/var/log/nginx/` shows that it works

### DIFF
--- a/playbooks/utils/common_only.yml
+++ b/playbooks/utils/common_only.yml
@@ -6,6 +6,9 @@
   # hosts: staging:qa:production
   remote_user: pulsys
   become: true
+  vars_files:
+    - ../group_vars/nginxplus/main.yml
+    - ../group_vars/nginxplus/production.yml
   vars:
     running_on_server: true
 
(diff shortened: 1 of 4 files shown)


playbooks/utils/common_only.yml
```diff
@@ -6,6 +6,9 @@
   # hosts: staging:qa:production
   remote_user: pulsys
   become: true
+  vars_files:
+    - ../group_vars/nginxplus/main.yml
```
acozine, line 10: I believe this would mean that the app-protect log rotation directive would be put on all machines this playbook runs against.

acozine: Per discussion, the above comment is true. However, we are never going to merge this change. It is only being used as a stopgap while we wait for our new nginxplus license.

PR #6312, opened 2025-07-04
[cdh] docraptor API

### LINKED ISSUE #6279: Allow DocRaptor IPs to access to CDH staging webapps (prosody, shakespeareandco) to test PDF generation
We are adding functionality to use a [DocRaptor](https://docraptor.com/) API service to generate PDFs from html content; the API needs access to project URLs so it can get the html as well as all the associated assets (images, fonts, javascript, etc) needed to render the page.

We need to allow access on the prosody and shakespeareandco staging web applications so that stakeholders can test the new functionality.

The DocRaptor documentation recommends using basic auth or proxy (which I don't think works in our case), but they do also provide an IP list at a URL https://docraptor.com/documentation/api/ip_listing

These are the two staging NGINX+ config files:
- https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/cdh_test_prosody.conf
- https://github.com/pulibrary/princeton_ansible/blob/main/roles/nginxplus/files/conf/http/cdh_test_shakespeareandco.conf

We can provide login access to the PPA interface where the DocRaptor API request will be triggered, if that would be helpful when testing the new configuration.  (Or possibly a script that does the same thing, if that would be easier for testing.)

kayiwa:
- **add a script that checks for ips at docraptor**
- **enable prosody and shakespeare to use docraptor**

closes #6279


roles/nginxplus/files/conf/http/cdh_test_shakespeareandco.conf
```diff
@@ -67,6 +68,9 @@ server {
        proxy_send_timeout         2h;
        proxy_read_timeout         2h;
        proxy_intercept_errors on;
+# either block docraptor or our allow list
+        satisy any;
```
acozine: Typo:
```suggestion
        satisfy any;
```

PR #6314, opened 2025-07-07
[recapgfa] reboot exception

kayiwa:
the libserv44 host is heavily used during the maintenance window
we add it to the list of hosts that will not be rebooted automatically
we will get a slack alert for Ops that can allow a reboot to happen at Central Time

related to <https://github.com/PrincetonUniversityLibrary/security/issues/78>


acozine (commented):
Nice work finding the way we handle the manual reboots for the ansible servers. Just one change to the logic (see below).

PR #6318, opened 2025-07-08
Update bib API Key for Lib Jobs Prod, 

kevinreiss:
Also removes retired staff from notifications for issue https://github.com/pulibrary/lib_jobs/issues/933.


sandbergja (approved):
Thanks, @kevinreiss ! 🚲

PR #6328, opened 2025-07-09
[lib_jobs] logrotate with correct user

kayiwa:
we want su to be root so logrotate does not complain.
put back the tast in lib-jobs role that created the logrotation


VickieKarasic (approved):
Ok - thanks @kayiwa!

Issue #5718, opened 2025-01-09
Create a Tower template for adding/updating the CheckMK agent


acozine:
### User story
As an engineer, I want CheckMK to stay up-to-date with any changes (new services installed, replacement VMs, etc.).

### Acceptance criteria

- [x] I can run the `utils/checkmk_agent.yml` playbook from a template in Tower and install the CheckMK agent on a single VM with `--limit <vmname>`
- [x] I can run the `utils/checkmk_agent.yml` playbook from a template in Tower, update all the monitored services on existing VMs, and ensure that the CheckMK agent is installed on all VMs in an environment with `--limit <envname>`

### Concrete example
If I run the Tower template with `--limit sandbox-xk2843.princeton.edu`

### Implementation notes, if any
- [x] create the template
- [x] install the [checkmk.general collection](https://galaxy.ansible.com/ui/repo/published/checkmk/general/) on an EE (either update an existing EE or build a new, special-purpose one)
- [x] add a documentation line to the playbook noting that to run it locally, you need to install the checkmk.general collection
- [x] test the template on a new VM to confirm that the playbook installs the agent 
- [x] test the template on a replacement VM to confirm that the playbook updates data collection
- [x] test the template on a VM with recent changes to confirm that the playbook updates monitored services

acozine:
The CheckMK collection was added to our standard/base EE by https://github.com/pulibrary/princeton_ansible/pull/4783. However, the EE was not rebuilt at the time - rebuilt it now, and we're on to the next problem.

acozine:
Playbook currently fails - both locally and on Tower - on the `Linux - Download Vanilla CRE agent` task with this error:

```
fatal: [lib-postgres-qa1.princeton.edu]: FAILED! => {"attempts": 3, "changed": false, "dest": "/tmp/check-mk-agent_2.2.0p22-vanilla.deb", "elapsed": 0, "msg": "Request failed: <urlopen error [Errno 111] Connection refused>", "url": "http://localhost:80/my_site/check_mk/agents/check-mk-agent_2.2.0p22-1_all.deb"}
```

acozine:
[...]

PR #6344, opened 2025-07-16
[CheckMK] expands playbook documentation, requires 'limit'

acozine:
Partially addresses #5718.

This PR adds documentation to the playbook about what is required to run it locally.

It also adds a pre-task that stops the playbook if you do not use 'limit'. Since we must define which CheckMK service and which folder to add VMs to, it seems smart to force everyone to use 'limit' - otherwise we risk having all our VMs added to a single CheckMK service and folder.


kayiwa (approved):
OK. Thanks @acozine !

PR #6349, opened 2025-07-21
[checkmk] monitor nginxplus config

### LINKED ISSUE #5364: [Check_MK] Monitor nginxplus config regularly
### User story
As an operations person, I want to know when the inactive load balancer is failing its nginx config check, so I can fix it before the active load balancer bumps traffic over.

### Acceptance criteria
- [ ] A check in CheckMK runs `nginx -t` regularly, knows what failure looks like, and alerts us. 
- [ ] Test by pushing a config file with an error in it to the inactive load balancer and confirming that we get an alert to Slack and/or elsewhere saying that `nginx -t` has failed.

Start with the dev load balancers, then expand to production.

kayiwa:
- **add checkmk to sudoers group**
- **add a checkmk shell script**
- **add a template task that is ONLY installed on adc-* hosts**

closes #5364


acozine:
To avoid adding this check to all servers, we put the script and the task that installs it into the nginxplus role. We plan to follow this pattern for other local checks, like the rails health check and the passenger status check.

PR #6352, opened 2025-07-21
[lib_jobs] Set gitlab host to production

regineheberlein:


sandbergja (approved):
Thanks, @regineheberlein ! 🐝

PR #6356, opened 2025-07-22
[svn] remove lib-svn-staging1 from inventory

kayiwa:
we have migrated the staging to use git

related to https://github.com/pulibrary/ops-catchall/issues/84


VickieKarasic (approved):
Thanks @kayiwa

PR #6361, opened 2025-07-22
removes another unused sandbox

acozine:
A bit more tidying on our staging inventory.

This host has already been removed from CheckMK and from vSphere.


kayiwa (approved):
OK! Thanks @acozine

PR #6366, opened 2025-07-24
I4788 ezproxy test shibuser

### LINKED ISSUE #4788: EZproxy Test Can't Use Shibboleth

kayiwa:
closes #4788


kevinreiss (approved):
Thanks @VickieKarasic and @kayiwa.

PR #6368, opened 2025-07-25
Updating ORCID vars for the new settings and the Rails vars to include quotes

carolyncole:


roles/rails_app/templates/rails_app_config
```diff
@@ -1,4 +1,4 @@
 {% for variable in rails_app_vars %}
-export {{variable.name}}={{variable.value}}
+export {{variable.name}}="{{variable.value}}"
```
carolyncole, line 2: @kayiwa I think this should be benign.  What do you think?

PR #6369, opened 2025-07-25
updated the curl url

aruiz1789:


kayiwa (approved):
Nice catch @aruiz1789

PR #6371, opened 2025-07-28
Use only one pair of double quotes for Bibdata allow list

sandbergja:
Prior to this commit, running the bibdata playbook led to the following line in ~deploy/app_configs/bibdata:

```
export BIBDATA_IP_ALLOWLIST=""128.112.200.203 128.112.204.114""
```

And when you ran `env | grep -i allowlist`, you got:

```
BIBDATA_IP_ALLOWLIST=128.112.200.203
```

And when you log into a new SSH session, you got this distracting error:

```
-bash: export: `128.112.204.114': not a valid identifier
```

Since #6368 was merged, we no longer need to supply double quotes for this environment variable, so let's remove them so we only have one pair of quotes.

I have run this on staging and qa.

### DIFF
--- a/group_vars/bibdata/production.yml
+++ b/group_vars/bibdata/production.yml
@@ -57,7 +57,7 @@ rails_app_vars:
   - name: BIBDATA_DB_HOST
     value: '{{ postgres_host }}'
   - name: BIBDATA_IP_ALLOWLIST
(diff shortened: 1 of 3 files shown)


regineheberlein (approved):
thanks @sandbergja

PR #6380, opened 2025-07-30
[TOWER] Update tower to allow for newer version of capistrano

carolyncole:
The new Hanami ORCID application utilizes a newer version of Capistrano.  This should allow that application and older ones to deploy.


kayiwa (approved):
OK! Thanks @carolyncole

PR #6388, opened 2025-07-31
Seeking the right combination of quotes and escaping for GCP inventory vars

acozine:
More follow-up on #6385 and #6379.

As currently written in the main branch, the ssh options for gcp resources behind the bastion hosts are not correctly pulled into Tower.

We need:
`ansible_ssh_common_args: -o ProxyCommand="ssh -o 'ForwardAgent yes' -o 'StrictHostKeyChecking=accept-new' -q pulsys@bastion-prod.pulcloud.io -W %h:%p"`

Tower has:
`"ansible_ssh_common_args": "'-o ProxyCommand=\"ssh '-o ForwardAgent yes' '-o StrictHostKeyChecking=accept-new' -q pulsys@bastion-prod.pulcloud.io -W %h:%p\"'"`

This PR tries three different approaches to quotations and escaping for the gcp inventory. Again, we need to merge these to main so we can test them in Tower. If one of them works correctly, I'll follow up with a PR to fix the other two.

Fingers crossed.

PR #6391, opened 2025-08-01
Allow Tower to connect to GCP resources 

acozine:
Final piece of GCP connectivity through our bastion hosts.

Related to #6388, #6385, #6379, #6378, and #6376.

In #6388 we tried three different options for the ProxyCommand. The vars for the GCP prod group worked. This PR updates the other two groups to use the same syntax.


kayiwa (approved):
OK! thanks @acozine

Issue #6357, opened 2025-07-22
Common role maintenance failures in staging


acozine:
(Edited to alphabetize the list of failures)
### What maintenance needs to be done?
Remove inventory entries for obsolete/decommissioned VMs.
Update inventory to allow SSH connections to unreachable VMs.
Update common role to accommodate other failures.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Relevant deadlines
N/A

### Why is this maintenance needed?
We want to be able to run the common role as a maintenance task and have it installed necessary tools, update configuration, and generally prevent drift across our infrastructure. 

### Acceptance criteria
 - [ ] I can run the [Common Maintenance template](https://ansible-tower.princeton.edu/#/templates/job_template/90/details) with `--limit staging`, all hosts are reachable, and no tasks fail
 
### Failures from July 22, 2025:

- [x] bastion-staging.pulcloud.io - Can't find tmux
- [x] cdh-test-derrida-crawl1.princeton.edu - Module failure on the task 'Install common shared packages' - decommission, see #6358
- [x] dpul-collections-staging1.lib.princeton.edu - /etc/chrony does not exist
- [x] dpul-collections-staging2.lib.princeton.edu - No package htop available
- [x] gcp_dataspace_staging1 - unreachable
- [x] gcp_oar_staging1 - unreachable
- [x] iiif-staging1.princeton.edu - unreachable
- [x] lib-sftp-staging1.princeton.edu - /etc/chrony does not exist
- [x] lib-svn-staging1.princeton.edu - unreachable - removal in progress
- [x] mflux-staging1.lib.princeton.edu - /etc/chrony does not exist
- [x] nomad-client-staging1.lib.princeton.edu - /etc/chrony does not exist
- [x] nomad-client-staging2.lib.princeton.edu - /etc/chrony does not exist
- [x] nomad-client-staging3.lib.princeton.edu - /etc/chrony does not exist
- [x] pdc-globus-staging-embargo - No package matching 'python-setuptools' 
- [x] recapgfadev2.princeton.edu - /etc/chrony does not exist
- [x] recapgfatest.princeton.edu - /etc/chrony does not exist
- [x] sandbox-rl36671.lib.princeton.edu - Unable to start service vector
- [x] sandbox-tw8766.lib.princeton.edu - unreachable - can be removed, long since decommissioned
- [x] slavery-staging-upgrade1.lib.princeton.edu - unreachable - decommissioned, see #6306
[...]

Issue #6306, opened 2025-07-02
[pas] decommission slavery-staging-upgrade{1,2} machines


VickieKarasic:
### What maintenance needs to be done?

The following machines can be decommissioned, as they are no longer needed: 

- [ ] slavery-staging-upgrade1.lib.princeton.edu
- [ ] slavery-staging-upgrade2.lib.princeton.edu

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### Why is this maintenance needed?
- These machines are no longer needed 

### Acceptance criteria
 - [x] Machines are powered off
 - [x] After 30 days, machines are deleted 
 - [x] Delete DNS records
 - [x] Remove from [Ansible files](https://github.com/search?q=repo%3Apulibrary%2Fprinceton_ansible%20slavery-staging-upgrade1.lib.princeton.edu&type=code)

PR #6392, opened 2025-08-01
removes pas upgrade VMs from inventory

acozine:
Partial fix for #6306.

Partial fix for #6357.

We no longer need staging-upgrade VMs for the PAS project. They've already been shut off and will be deleted soon. 

This PR removes them from our Ansible inventory.


sandbergja (approved):
Thanks, @acozine ! 🐴

PR #6407, opened 2025-08-05
Renamed the role from rails_app to ruby_app

carolyncole:
Made small changes to aknowledge the role is now tested for more than just rails Did not rename all the variables, becuase I did not want to change everyone's config files


hackartisan (commented):
If you wanted to use `ruby_` variables in your app, could you create them and set the rails ones to use those values if they're there? Just an idea.

carolyncole:
> If you wanted to use `ruby_` variables in your app, could you create them and set the rails ones to use those values if they're there? Just an idea.

Yes, I was thinking about allowing the role to accept both.  I thought this would be one step and then I would worry about the var names next...

hackartisan (approved):

PR #6416, opened 2025-08-06
[ruby_app] Updating the ruby_app role to allow either rails_ or ruby_ vars to be passed to the role

carolyncole:


carolyncole:
@hackartisan here is my thought for renaming the vars after the role rename.

PR #6417, opened 2025-08-06
[LIB-JOBS] Remove the lib-jobs specific app config sfx_wroldcat

carolyncole:
Rely on the app config created by the ruby app role


christinach (approved):
thanks @carolyncole

PR #6418, opened 2025-08-06
[orangelight] Remove unnecessary duplicate quotes from environment variable

sandbergja:
With the duplicate quotes, only the first username in the array actually gets added to the `ORANGELIGHT_ADMIN_NETIDS` environment variable, and everybody else does not get access to the flipflop features.


christinach (approved):
thanks @sandbergja !

PR #6427, opened 2025-08-08
[common] check for checkmk

kayiwa:
Our common role is setup assuming our virtual machines have checkmk
we add a task that checks for the checkmk binary. If it is present the
logrotate file is added. Otherwise it skips


VickieKarasic (approved):
Looks good - thanks @kayiwa

PR #6430, opened 2025-08-12
Add a bulk queue to Figgy.

tpendragon:
This will let us queue things that are below "default", but above "low",
so that bulk ingests won't prevent other things.

Work towards pulibrary/figgy#6765


hackartisan (commented):
Are these configured in ansible right now or something? And we can move them here this way?

hackartisan:
Oh, ha. Okay those probably aren't the right defaults, they seem sort of figgy-specific 😂 Oh, well!

PR #6437, opened 2025-08-14
Add issues from ReCAP LAS GitLab repo to the Ops team board

acozine:
Closes Ops Team Handbook  ticket 149.

Pulls issues from two repos in GitLab into the ops team board.


acozine:
Currently drafted because I'm still hoping to find a way to loop over the repos instead of repeating all the tasks. No luck so far, though.

PR #6446, opened 2025-08-19
[pul_solr] Upload a copy of new jars for solr9

rladdusaw:
Related to pulibrary/pul_solr#484


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #6449, opened 2025-08-21
[CircleCI Runner] Clean up CircleCI Deployer Working Directory

tpendragon:
The nomad boxes were running out of disk space because the deployer
wasn't cleaning up after itself.


christinach (approved):
thanks @tpendragon !

PR #6453, opened 2025-08-22
Playbook to revoke SSL certs should check `nginx -t` for safety

acozine:
When we changed the name of the EAL site from eal-apps.princeton.edu to eal.lib.princeton.edu, we changed the config file name and generated new SSL certificates, but the old certs and the old config file were still on the load balancers. When the certs came up for renewal, we revoked the obsolete ones . . . which meant the obsolete config file was looking for cert files that no longer existed.

This PR adds a safety check to the revoke-a-cert playbook, with a message if the `nginx -t` config check fails.

It also removes the duplicate entry for `vars_files` and hard-codes the slack alerts channel.

### DIFF
--- a/playbooks/utils/incommon_certbot_revoke.yml
+++ b/playbooks/utils/incommon_certbot_revoke.yml
@@ -3,7 +3,7 @@
 # to run in production, you must add '-e runtime_env=production'
 # Run the playbook on the inactive load balancer first
 # if the 'nginx -t' check fails, we probably didn't delete an obsolete config file
-# run 'playbooks/nginx.yml' following the comments for removing obsolete config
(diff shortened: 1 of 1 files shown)


playbooks/utils/incommon_certbot_revoke.yml
```diff
@@ -15,8 +18,6 @@
       msg: "you must use -l or --limit"
     when: ansible_limit is not defined
     run_once: true
-  vars_files:
-    - ../group_vars/all/vault.yml
```
acozine, line 19: This section is duplicated above (old lines 9 and 10, new lines 12 and 13).

Issue #6463, opened 2025-08-27
Create an Ansible Role for Apache Solr and Node.js Mirror Synchronization


kayiwa:
Our internal mirror for Apache Solr and Node.js are critical for our build processes and dependency management. The current update process involves a cron rsync command that fails and has to be actively monitored by an administrator. This is a time-consuming and error-prone process that lacks consistency and a clear audit trail.

Create a single, reusable Ansible role that can be used to synchronize any mirror. The role should be generic enough to handle both the Solr and Node.js mirrors by accepting variables. The core logic should use the ansible.builtin.rsync module to ensure efficient and reliable file synchronization.

**Acceptance Criteria**

- The role is created in the ansible/roles directory with an appropriate name (e.g., mirror_sync).

- The role includes a defaults/main.yml file to define configurable variables such as source, destination, and user.

- The role's main task uses the ansible.builtin.rsync module.

- The role can be successfully applied to both the Apache Solr and Node.js mirrors

- The role includes a README.md file with clear instructions on how to use it, including an example playbook snippet.

- Necessary file permissions are handled correctly within the role to ensure the mirror is accessible after synchronization.

- The role handles the correct user context for the rsync operation.

**Tasks**

- [ ] Create the new Ansible role directory structure.

- [ ] Define variables in defaults/main.yml.

- [ ] Test the role on a test environment for both Solr and Node.js mirrors.

- [ ] Add comprehensive documentation to the README.md file.

acozine:
Can this be closed? We merged #6464 - is there more work that needs to be done on this ticket?

PR #6464, opened 2025-08-27
add a role that manages our mirror

kayiwa:
the pulmirror needs a hand at the wheel. this role allows us to manage our mirrors.
related to #6463


sandbergja (approved):
Thanks, @kayiwa !

PR #6466, opened 2025-08-28
Fix ansible-galaxy installation instructions

bess:
Ref https://github.com/pulibrary/tigerdata-app/issues/1771


carolyncole (approved):
Thanks @bess

Issue #6467, opened 2025-08-28
nginx config for tiger-data CI should only define one proxy pass for `/cable`


acozine:
The config file for tigerdata-ci forwards `/` on port 80 to port 443. It also defines proxy passes for `/cable` [under port 80/HTTP](https://github.com/pulibrary/princeton_ansible/blob/0f05f4ef136184e443e8d0029debb3681fbb3d9f/roles/nginxplus/files/conf/http/dev/tigerdata_ci.conf#L23) as well as [under port 443/HTTPS](https://github.com/pulibrary/princeton_ansible/blob/0f05f4ef136184e443e8d0029debb3681fbb3d9f/roles/nginxplus/files/conf/http/dev/tigerdata_ci.conf#L49). 

We should forward all traffic to port 443 and define the proxy pass under the port 443/HTTPS section so users can only access `/cable` with HTTPS.

acozine:
Closed by #6468.

PR #6468, opened 2025-08-28
[tigerdata] update tigerdata-ci config to define only one proxy pass

leefaisonr:
ref #6467


acozine (approved):
Looks good, thanks @leefaisonr!

PR #6474, opened 2025-09-02
[nginxplus] App protect config

kayiwa:
we can no longer install app-protect. Remove all references to it on the prod loadbalancers.


carolyncole (approved):
Thanks @kayiwa!

PR #6480, opened 2025-09-02
Set Solr9 so it will write to the usual backup directory

### LINKED ISSUE #6476: [Solr] Solr boxes that perform backups fail to reboot.
### Summary or User Story

We had to add a bind mount for Solr 9 backups to work correctly:  https://github.com/pulibrary/princeton_ansible/commit/b08dcff4d95c37220a2ce9002082e677fa37b054#diff-20a87ee4197a2c7db096b35402382f4f0e4aad5888f01743114139ef2720d16cR47-R65

Unfortunately, if the network mount it's pointed to isn't bound yet (which is the case on reboot), then that mount fails and prevents Ubuntu starting up. An example of someone else hitting that: https://askubuntu.com/questions/991068/ubuntu-doesnt-boot-when-mount-bind-not-available-during-startup

We found this issue as part of an incident where patch monday for staging servers resulting in lib-solr-staging4d and lib-solr-staging1 being down from that point forward:

https://docs.google.com/document/d/1KJH4EUob6rJfP4a1927TM_tLrUZJLLYT4If5a-Mgj-o/edit?tab=t.0

The only thing preventing this from happening to production is that Francis runs those processes manually, and part of his process is uncommenting that mount from `fstab` manually, restarting, then putting it back and remounting.

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [ ] lib-solr-staging1.princeton.edu and lib-solr-staging4d.princeton.edu can be safely rebooted.

### First step

Brainstorm some solutions, do some searching, try some things.

acozine:
Closes #6476 . . . we think/hope?

The Solr9 servers have not been writing to `/solr/data/backup`. We thought this was a restriction related to CIFS mounts. As a workaround, we had a double-mount on the Solr servers - first we mounted the Isilon share to `/mnt/solr_backup` and then we mounted `/mnt/solr_backup` to `/solr/data/backup`. 

This [SO post](https://stackoverflow.com/questions/76169675/how-to-set-solr-index-data-path-version-9-2-0) suggested that a change to the Solr9 security policy would let us write backups directly to the CIFS mount at `/solr/data/backup`.

This PR adds a task to update the security policy, removes the tasks that create the double mount, and updates the allowed paths for simplicity.

We made these changes manually on `lib-solr-staging1`, unmounted everything except the simple CIFS mount of the Isilon share to `/solr/data/backup`, restarted Solr, and confirmed that we could create backups.

Changed files:

- `/etc/fstab`
- `/etc/default/solr.in.sh`
- `/opt/solr-9.2.0/server/etc/security.policy`

We did not test the automation. There may be bugs.


acozine:
Failing CI with what looks like a legit failure - the blockinfile task adds the two lines again on the second run:
```
   CRITICAL Idempotence test failed because of the following tasks:
  *  => solr9cloud : Configure security to allow Solr 9.2 to write to backup dir
```

roles/solr9cloud/tasks/config.yml
```diff
@@ -1,91 +1,72 @@
 ---
 # Configure SolrCloud
+# Current not used - we are using the solr8_config and solr9_config files instead
```
christinach, line 3: Can we remove the commented part? Reviewing with @sandbergja

kayiwa:
many moons ago we only had one "not managed by ansible" zookeeper ensemble. We were uncomfortable touching this and decided to use znodes for `solr7`, `solr8` and we are stopping this at `solr9`. We are slightly more comfortable with zookeeper and at least have it managed via ansible. The consequence is that all our solr clusters have a dedicated zookeeper ensemble to the cluster.

PR #6481, opened 2025-09-03
[solr] create cloud solr mount

kayiwa:
- **add tasks to create a gcsfuse mount**
- **create a mount that uses gcsfuse**
- ** uses [gcsfuse](https://github.com/googlecloudplatform/gcsfuse)

related to #6476


roles/solr9cloud/tasks/backup_gcs_mount.yml
```diff
@@ -0,0 +1,173 @@
+---
+# Configure GCS backup mount for SolrCloud using gcsfuse
+- name: Solrcloud | Install gcsfuse dependencies
```
sandbergja, line 3: Would the first few tasks here make sense in their own gcsfuse role?  Definitely not a blocker.

roles/solr9cloud/files/solr-backup-account-key.json
```diff
@@ -0,0 +1,122 @@
+$ANSIBLE_VAULT;1.1;AES256
```
sandbergja, line 1: Would it make sense to have separate keys for staging vs prod?

sandbergja (approved):
Thanks, @kayiwa !  This looks great, a nice way to start moving away from the isilon for solr backups.  Two questions, neither is a blocker from my perspective.

roles/solr9cloud/tasks/backup_gcs_mount.yml
kayiwa: they would. I think we should do it "when" (and if we remember) we have another use for gcsfuse

PR #6486, opened 2025-09-04
simplify the backup mounts in solr8

acozine:
Partial fix for #6476.

We mount the Isilon share directly to `/solr/data/backup` instead of mounting it to `/mnt/solrbackup` and then adding a bind mount from `/mnt/solrbackup` to `/solr/data/backup`.

Once we confirm that this allows the lead Solr box to write backups and reboot with no drama, the next step is to send backups to GCE.

### DIFF
--- a/roles/solrcloud/tasks/main.yml
+++ b/roles/solrcloud/tasks/main.yml
@@ -11,8 +11,8 @@
     mode: 0755
     group: deploy
     owner: deploy
-  loop:
-    - solr_backup
+  tags:
+    - fix_backups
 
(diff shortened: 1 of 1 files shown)


acozine:
Tested today - we replaced the lib-solr-staging1 VM, rebuilt it with the solr9 playbook from this branch (plus the checkmk agent playbook), then deployed `pul_solr`. Everything worked. We then tested backups to the Isilon as mounted at `/solr/data/backups`. That worked. We then tested rebooting the machine. That worked.

PR #6497, opened 2025-09-09
[solr] Use version 9.4.2 of the lucene-analysis-icu jar

sandbergja:
This version number needs to match our lucene version, rather than our solr version.

Hopefully this will resolve the following error when attempting to analyze the cjk_* dynamic field in the catalog config:

```
Cannot invoke "org.apache.lucene.analysis.icu.segmentation.ICUTokenizerConfig.combineCJ()" because "config" is null
```

Helps with https://github.com/pulibrary/bibdata/issues/2911


sandbergja:
Closing in favor of #6499

Issue #6477, opened 2025-09-02
Solr boxes should report when they're down


tpendragon:
### Summary or User Story

Created as a part of this incident: https://docs.google.com/document/d/1KJH4EUob6rJfP4a1927TM_tLrUZJLLYT4If5a-Mgj-o/edit?tab=t.0

lib-solr-staging1.princeton.edu and lib-solr-staging4d.princeton.edu were both down for over a day before we noticed. CheckMK had a failing service that it couldn't reach that box, but there were no alerts in `#sensu_alerts` (which we probably would have missed), and no alerts somewhere more visible.

### Acceptance Criteria

- [ ] When a Solr box is down for > 1 hour, we should get an alert.

### First step

Look into ways to adjust the CheckMK alert.

aruiz1789:
Hey @tpendragon, 
I am working on setting alerting for the stating solr boxes.   Do you want this alert to go to a particular Slack channel?

aruiz1789:
Checkmk solr staging services monitoring.

<img width="1304" height="208" alt="Image" src="https://github.com/user-attachments/assets/17239908-753f-4127-9e9b-f8bd052363e0" />

 Working on two issues:

1.  Zookeeper service.

- **lib-solr-staging4d-6d** Can we disable the zookeeper service
```
pulsys@lib-solr-staging4d:~$ sudo systemctl is-active zookeeper.service
failed
pulsys@lib-solr-staging4d:~$ sudo systemctl is-enabled zookeeper.service
enabled
```

2. Checkmk local check for Solr backup.
 
Are we still managing backups for staging solr 8 via this script? 

- Mount point is different from what's in the local check script.
```
pulsys@lib-solr-staging4d:~$ df -t cifs
Filesystem                                 1K-blocks          Used    Available Use% Mounted on
//diglibdata1.princeton.edu/solrbackup 1548317583360 1356102129432 192215453928  88% /mnt/solr_backup

pulsys@lib-solr-staging4d:~$ ls -ald /solr/data/backup/solr8/staging
ls: cannot access '/solr/data/backup/solr8/staging': No such file or directory
```

- The latest backup is from 20250901.
```
pulsys@lib-solr-staging4d:~$ ls -lt /mnt/solr_backup/solr8/staging/
total 0
drwxr-xr-x 2 deploy deploy 0 Sep  1 00:00 20250901
```

- For cloud backups no backup set are present.
```
[...]

PR #6501, opened 2025-09-10
I6477 notify solr

kayiwa:
related to #6477


kayiwa:
> I'm confused, how is this related to solr downtime checking?

 Couple things, this was supposed to be a draft PR but my brain wasn't switched on, the checkmk community forum seemed to recommend this as a way to alert where one remains fully in control. We are using slack happily and I was testing to this last week. It got to hello world and I believe @aruiz1789 is further along.

kayiwa:
[Notify](https://exchange.checkmk.com/p/notify-via-ntfy) goes to your phone via pub/sub

Issue #6487, opened 2025-09-05
Get ready for applications to migrate to Solr9


acozine:
### Solr9 backups and upgrade

- Fix 9.2 backups on staging and production
  - fix for security
  - send backups to GCE
  - update checkmk alerting
- DACS to run local tests against version 9.9, work through whatever breaks
- upgrade staging Solr9 to version 9.9
  - see if the upgrade breaks anything for CDH
  - most likely breakages are if specific versions are called on cdh-ansible
  
### Provisioning and performance testing
- ops and DACS to work together on reducing memory provisioning and then testing performance in staging Solr9
  - Solr9 provisioning currently identical to Solr8
  - VMs are likely overprovisioned for memory - how much can we reduce this?
-  this all works, we can replicate the work in production Solr9

### Acceptance criteria

- [ ] Catalog-staging runs happily in Solr 9.9
- [ ] Solr9 backups happen regularly
- [ ] Solr9 backups are monitored
- [ ] Solr9 backups go to GCE
- [ ] Backup files in GCE are pruned
- [ ] We can restore from a GCE backup

acozine:
Adding an acceptance criterion of:
- [ ] we can rebuild a Solr9 cluster using Prancible and it works

That work requires closing #5496.

Issue #6472, opened 2025-08-29
Update Catalog Staging Group Vars to Point to Solr 9


kevinreiss:
- [x] Group_var values for the solr version that catalog staging is using points to solr 9.

PR #6504, opened 2025-09-12
Upgrade solr to version 9.9.0

christinach:
Upgrade solr to version 9.9.0
To include the fix https://issues.apache.org/jira/browse/SOLR-17515 that was included in solr 9.8.0 

related to #6487
related to #6472

see also [v9.9.0.dependency_upgrades](https://solr.apache.org/docs/9_9_0/changes/Changes.html#v9.9.0.dependency_upgrades)

### DIFF
--- a/roles/solr9cloud/templates/solr.in.sh.j2
+++ b/roles/solr9cloud/templates/solr.in.sh.j2
@@ -35,7 +35,7 @@ ENABLE_REMOTE_JMX_OPTS="true"
 RMI_PORT="1099"
 
 # All system properties in a single line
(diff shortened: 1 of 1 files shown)


christinach:
@sandbergja found one of the references in [cdh solrconfig.xml](https://github.com/Princeton-CDH/ppa-django/blob/6e75c500968e74b0198611b91da4a50f6c28ddb2/solr_conf/conf/solrconfig.xml#L1156-L1157)

christinach:
We ran the playbook only against lib-solr-staging2

PR #6509, opened 2025-09-16
upgrade to zookeeper 3.9.4

kayiwa:
we are seeing errors on the Solr UI saying `Unimplemented` when writing collection state.
move from Ubuntu package zookeeper 3.4
make use of systemd (start-foreground) simple service management

### DIFF
--- a/roles/zookeeper/defaults/main.yml
+++ b/roles/zookeeper/defaults/main.yml
@@ -11,8 +11,8 @@ zk_client_port: 2181
 zk_leader_port: 2888
 zk_election_port: 3888
 
-# 4lw whitelist for health checks you already use (ruok/mntr/srvr)
-zk_4lw_whitelist: "ruok,srvr,mntr,conf,isro"
+# 4lw allowlist for health checks you already use (ruok/mntr/srvr)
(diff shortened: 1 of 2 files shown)


christinach (changes requested):
@kayiwa thank you! We've reviewed with @sandbergja and had a question.

kayiwa:
Zookeeper TimeOuts [documentation here](https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html)

and [more here](https://zookeeper.apache.org/doc/r3.5.9/zookeeperProgrammers.html)

kayiwa:
useful four letter word examples

```bash
# health check (expects 'imok')
echo ruok | nc -w 2 <host> 2181

# server details (mode, version, latencies)
echo srvr | nc -w 2 <host> 2181

# metrics (tab-separated key/value lines)
echo mntr | nc -w 2 <host> 2181

# active configuration
echo conf | nc -w 2 <host> 2181

# read-only state (ro or rw)
echo isro | nc -w 2 <host> 2181

```

christinach (approved):
Thank you @kayiwa !

Issue #2305, opened 2021-06-09
Set tcp_keepalives params in postgres configuration


hackartisan:
We may want to consider setting these to something like

tcp_keepalives_count = 10
tcp_keepalives_idle = 30
tcp_keepalives_interval = 30

The system default values appear to be:
```
$ sysctl -A | grep net.ipv4.tcp_keepali
net.ipv4.tcp_keepalive_intvl = 75
net.ipv4.tcp_keepalive_probes = 9
net.ipv4.tcp_keepalive_time = 7200
```

meaning it takes more than 2 hours to drop a lost connection.

More input is needed here regarding potential downsides to this change.

postgres docs on these settings:
- https://postgresqlco.nf/doc/en/param/tcp_keepalives_count/
- https://postgresqlco.nf/doc/en/param/tcp_keepalives_idle/
- https://postgresqlco.nf/doc/en/param/tcp_keepalives_interval/

I investigated this option in response to a repeated / intermittent PG::UnableToSend: no connection to the server from figgy3. Log message:
```
{
	"id": "AQAAAXnxGoZwVtYU4AAAAABBWG54R3BVUEFBQVg2QXVoeUdKSUh3QUI",
	"content": {
		"timestamp": "2021-06-09T14:08:38.000Z",
		"tags": [
			"source:ruby",
			"service:figgy",
			"filename:production.log",
			"sourcecategory:sourcecode",
			"application:figgy",
			"environment:production",
			"type:webserver"
		],
		"host": "figgy3",
		"service": "figgy",
		"message": "[500] GET / (CatalogController#index)",
		"attributes": {
			"duration": 4850000,
			"action": "index",
			"controller": "CatalogController",
			"http": {
				"status_code": 500,
				"url_details": {
					"path": "/"
				},
				"method": "GET"
			},
			"format": "html",
			"sql_queries": "'User Load (0.36) SELECT  \"users\".* FROM \"users\" WHERE \"users\".\"id\" = $1 ORDER BY \"users\".\"id\" ASC LIMIT $2'",
			"ddsource": [
				"ruby"
			],
			"@version": "1",
			"@timestamp": "2021-06-09T14:08:38Z",
			"db": 0.36,
			"error": "ActiveRecord::StatementInvalid: PG::UnableToSend: no connection to the server\n: SELECT  \"users\".* FROM \"users\" WHERE \"users\".\"id\" = $1 ORDER BY \"users\".\"id\" ASC LIMIT $2",
[...]

PR #6510, opened 2025-09-16
[Figgy + Postgresql] Set keepalive timeout lower.

tpendragon:
We were having postgres timeouts which was resulting in some items not
being able to be ingested in Figgy. This value allowed those items to
ingest. The relevant honeybadger error:
https://app.honeybadger.io/projects/53391/faults/119695976

Closes https://github.com/pulibrary/figgy/issues/6800
Related to https://github.com/pulibrary/princeton_ansible/issues/2305

This will only change the setting for Figgy, and keep the default for
the other postgres cluster.


kayiwa (approved):
OK Thanks @tpendragon !

PR #6515, opened 2025-09-17
Lapsed certs

kayiwa:
add cicognara and pcdm certificates


acozine:
By default, certs now last 90 days, so we will need to update these more often.

PR #6521, opened 2025-09-19
add a checkmk check for solr9 backups

kayiwa:
we add a checkmk localcheck for backups
related to #6487


sandbergja (approved):
This is so cool!  Thanks, @kayiwa ! 🦒

Issue #6541, opened 2025-09-25
[Bibdata] Bibdata staging solr vars use solr9 staging


christinach:
### What maintenance needs to be done?
Update group_vars in bibdata staging to use solr9 staging

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- We have run a full reindex on solr9 staging and is pointing to catalog-staging. 
- We want to point bibdata staging to solr9 staging

### Acceptance criteria
 - [x] Bibdata staging is using solr9 staging
 - [x] An incremental publishing job from alma sandbox is indexed successfully on staging started at `1:13pm Sep 25th, 2025`

christinach:
resolved by https://github.com/pulibrary/princeton_ansible/pull/6545

christinach:
waiting for the incremental from the alma sandbox

PR #6545, opened 2025-09-25
[Bibdata-staging] Point bibdata staging group vars to solr staging

christinach:
related to #6541


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #6550, opened 2025-09-26
the use of the package module does not actually remove packages

kayiwa:
running this uses the local package manager instead of the package module.

### DIFF
--- a/playbooks/utils/uninstall_package.yml
+++ b/playbooks/utils/uninstall_package.yml
@@ -1,21 +1,69 @@
 ---
-# you MUST run this playbook on a host or group with '--limit' for example `ansible-playbook -v ---limit pdc_describe_staging playbooks/utils/uninstall_package.yml -e package_to_remove=vector`
-#
+# Uninstall a package/service on RHEL and Ubuntu systems
+# Usage: ansible-playbook -v --limit <host_or_group> playbooks/utils/uninstall_package.yml -e package_to_remove=<package_name>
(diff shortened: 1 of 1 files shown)


acozine:
One other thought on this - I'd hesitate to add a standalone playbook that adds packages. If we want to add a package to all our systems, I'd vote to add it to the `common` role vars and run that. If we want to add it to specific systems (all ruby boxes, or all solr boxes), I'd vote to add it to the relevant role and run that. 

The only reason I added a playbook to remove packages is that re-running `common` after we remove a package from the `common` role vars means we don't install it any more, but we're not removing it either.

kayiwa:
not married to the install a package. It was in my brain.

kayiwa:
> What behavior did you see that led to this change? Did the `package` module not select the correct package manager in some cases? Or did you SSH to a machine that the original playbook had run against and find a package still installed? A lot of the new tasks here seem to do things that Ansible generally does automatically - could adding `autoremove: true` do most if not all of what this PR is trying to achieve?

I ran this https://ansible-tower.princeton.edu/#/jobs/playbook/30305/output

and 

https://ansible-tower.princeton.edu/#/jobs/playbook/30307/output

and 

https://ansible-tower.princeton.edu/#/jobs/playbook/30308/output

When I logged to lib-solr-prod1 (which was complaining about the vector service) I found the package to still be in there. I also logged into allsearch-prod1.princeton.edu and it was still in there.

PR #6554, opened 2025-09-26
updates version of checkmk to match what we are running

acozine:
I hope this will fix the Tower error in the checkmk playbook about "a later version is already installed". 

Also, our automation should reflect the current reality.


sandbergja (approved):
Thanks, @acozine ! 🐴

PR #6573, opened 2025-10-08
[lib_jobs] Add staging gitlab repo for testing

rladdusaw:
Clone the gitlabs staging libsvn repo for use in testing on production servers.

### DIFF
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -31,7 +31,6 @@ deploy_user_github_keys:
   - https://github.com/sandbergja.keys
   - https://github.com/sdellis.keys
   - https://github.com/tpendragon.keys
-  - https://github.com/tventimi.keys
   - https://github.com/VickieKarasic.keys
 deploy_user_local_keys:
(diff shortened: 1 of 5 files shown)


group_vars/lib_jobs/production.yml
```diff
@@ -54,6 +54,7 @@ transaction_error_feed_recipients: '{{ vault_lib_jobs_prod_transaction_error_fee
 voucher_feed_recipients: '{{ vault_lib_jobs_prod_voucher_feed_recipients }}'
 # Git
 git_lab_host: 'gitlab-prod-vm.lib.princeton.edu'
+git_lab_staging_host: 'gitlab-staging-vm.lib.princeton.edu'
```
sandbergja: if this value is the same for staging and prod, could we put it in common.yml instead?

sandbergja:
Also, just to check @rladdusaw : do all staging and prod boxes have sufficient disk space for another copy of the repo (since it is so large!)

rladdusaw:
> Also, just to check @rladdusaw : do all staging and prod boxes have sufficient disk space for another copy of the repo (since it is so large!)

Staging is good, I've already deployed to it. I'll check on prod.

PR #6597, opened 2025-10-21
rm eal keys from pulsys user playbook

aruiz1789:
removing eal keys from pulsys user playbook as part of the decommissioning.


acozine (commented):
One last bit - can you also get rid of [these two lines](https://github.com/pulibrary/princeton_ansible/blob/c5040b301bb22d5e828fbf186c734a9b6e61f4ab/group_vars/all/vars.yml#L118-L119)?

PR #6600, opened 2025-10-22
I6526 [common ] journald

### LINKED ISSUE #6526: [log rotation] Set journald limits for logging
### User story
As an operations person, I want the journalctl logs to be respectful about how much disk they use.

On the towerdeploy1 machine the journal logs were using 1.4GB of space and the VM was alerting for disk usage today. I set `SystemMaxUse` to `800M` and restarted journald, and disk usage declined by 5%. 

### Acceptance criteria
Across our infrastructure, the output of `sudo journalctl --disk-usage` is reasonable related to the use of the machine and the available disk space.

- [ ] We confirm how `journalctl` logs are managed on all Linux OS flavors we use
- [ ] The common role inserts appropriate limits on the space the `journalctl` logs use
- [ ] Any time the `journalctl` log management settings are changed, we restart `journald`

Log rotation for journalctl on Ubuntu is controlled by `/etc/systemd/journald.conf`. There are four options:

- `SystemMaxUse` limits the total disk space that journal logs can use (this seems the most appropriate)
- `SystemKeepFree` ensures a specified amount of disk space remains free (this seems risky if some other files start using a lot of disk space)
- `SystemMaxFileSize` sets the maximum size of individual journal files
- `SystemMaxFiles` limits the number of individual journal files

Once changed, the settings must be activated by restarting `journald` with `sudo systemctl restart systemd-journald`.

kayiwa:
Setting limits on journald 

closes #6526

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -24,5 +24,3 @@
   ansible.builtin.systemd:
     name: systemd-journald
     state: restarted
-  when:
-    - running_on_server
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -93,7 +93,6 @@
     owner: root
     group: root
   when:
-    - running_on_server
     - ansible_service_mgr=="systemd"
 
 - name: Common | configure journald log limits
@@ -104,7 +103,6 @@
     owner: root
     group: root
   when:
-    - running_on_server
     - ansible_service_mgr=="systemd"
   notify:
     - restart journald


roles/common/README.md
```diff
@@ -68,6 +97,7 @@ Customize logrotate per service or environment by overriding `logrotate_rules` i
 - hosts: all
   become: true
   vars:
+    journald_system_max_use: "1G"
```
acozine: Here we say 1G, below we default to 800M.

roles/common/handlers/main.yml
```diff
@@ -16,3 +13,16 @@
...
+    daemon_reload: true
+    name: logrotate.timer
+    enabled: true
+
+- name: restart journald
+  ansible.builtin.systemd:
+    name: systemd-journald
+    state: restarted
+  when:
+    - running_on_server
```
acozine: We shouldn't need this in a handler - it only runs if some other task returns `changed`, so by definition if you're running a handler you're already failing idempotency.

roles/common/tasks/main.yml
```diff
@@ -85,6 +85,30 @@
...
+
+- name: Common | configure journald log limits
+  ansible.builtin.template:
+    src: "journald.conf.j2"
+    dest: "/etc/systemd/journald.conf.d/00-journal-size.conf"
+    mode: "0644"
+    owner: root
+    group: root
+  when:
+    - running_on_server
```
acozine: Same question here - do we need this?

roles/drupal10/molecule/default/converge.yml
```diff
@@ -11,10 +11,22 @@
...
+        name: locales
         update_cache: true
         cache_valid_time: 600
+
+    - name: Ensure en_US.UTF-8 locale is uncommented in /etc/locale.gen
+      ansible.builtin.lineinfile:
+        path: /etc/locale.gen
+        regexp: '^# en_US.UTF-8 UTF-8'
+        line: 'en_US.UTF-8 UTF-8'
+        state: present
```
acozine: Oh, I see, these changes are not related to the syslog work, just incidental test fixes?

acozine (commented):
Looks good - the README changes and template are especially nice. I was confused by the drupal10 changes - I assume they were added to make CI pass overall? I had a couple of other minor suggestions/questions.

roles/common/tasks/main.yml
acozine: Addressed in 49c28ef657a4d2b86ee8d03ec54352752b2aa91c

PR #6604, opened 2025-10-22
rm sidekiq role from lib_jobs playbook

aruiz1789:
Removed sidekiq-workers role from lib_jobs playbook.
Ran playbook on staging VMs


kayiwa (approved):
OK! Thanks @aruiz1789

PR #6623, opened 2025-10-28
Fixing the ruby health check to allow for urls other than localhost/health.json

carolyncole:
Running the fixed template on pdc discovery


kayiwa (approved):
OK! Thanks @carolyncole

PR #6642, opened 2025-10-31
migrate allsearch, approvals, datacommons to staging lb

aruiz1789:
Migrated nginx config from prod LB to dev LB for:
- allsearch-api_staging
- allsearch_staging
- approvals_staging
- datacommons_staging
ran playbook on dev lb to create configs and on production lb to clean config, also tested site availability.


kayiwa (approved):
OK! Not in your ticket but we will need to make sure the network registration is moved to adc-dev instead of lib-adc

PR #6649, opened 2025-11-04
migrated dacs staging sites to dev lb

aruiz1789:
Migrated dss, lib-jobs, lockers and study, slavery, static tables and repec sites from prod lb to staging lb.
Confirmed staging sites are loading as expecting.


acozine (commented):
Hooray for migrating these sites over to the dev LBs!

Can we delete the proxy-pass files from `roles/nginxplus/files/conf/http/templates` as we add them in `roles/nginxplus/files/conf/http/dev/templates`?

VickieKarasic (approved):
Looks good - thanks, @aruiz1789

aruiz1789:
Yes, I will delete the proxy-pass files if they are no longer needed by any other config file.
This one "dss-proxy-pass.conf" is still needed in the "dss-prod.conf" file.

PR #6653, opened 2025-11-06
[gitlab] remove shell script used to configure repo

kayiwa:
use ansible modules. This will unbreak main branch


acozine (commented):
For future reference, can you post the error we were seeing that this PR corrects? That way if we see a similar error in future, we have some pointers.

PR #6656, opened 2025-11-06
Update dnf repos to use local mirrors

aruiz1789:
Updated the common role to include the addition of a local mirror for Rocky dnf repos.


acozine (commented):
This is great progress. We need to decide whether we want to manage multiple defined mirrors or not. Can we pair up to test this out and see if we can get it to handle multiple mirrors?

acozine (approved):
LGTM, thanks @aruiz1789

PR #6660, opened 2025-11-07
Make sure CircleCI runners never stop trying to restart.

tpendragon:
They should spin up very fast when they crash and never stop trying,
because they fail on purpose.


christinach (approved):
Thanks @tpendragon !

PR #6664, opened 2025-11-11
modify the fits role to use a dif source

aruiz1789:
Updated the fits role to a new source to download the fits files.

This addresses a CI failure that was returning this error:
`request failed (403) for http://projects.iq.harvard.edu/files/fits/files/fits-0.8.5.zip`

### DIFF
--- a/roles/fits/defaults/main.yml
+++ b/roles/fits/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
-fits_version: "{{ '0.8.5' }}"
-fits_download_url: http://projects.iq.harvard.edu/files/fits/files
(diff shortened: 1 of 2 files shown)


roles/fits/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-fits_version: "{{ '0.8.5' }}"
-fits_download_url: http://projects.iq.harvard.edu/files/fits/files
+fits_version: "{{ '1.6.0' }}"
```
acozine: I think there was a reason we had pinned the version of FITS to 0.8.5. I don't know what it was, but if we still need FITS, we should ask about version constraints in the dev channel.

PR #6686, opened 2025-11-20
[tigerdata] adding tigerdata-ci1 to staging group

kayiwa:
we found that tigerdata-ci1 had not been patched for a while while it is in the inventory it was not part of any of the environments


carolyncole (approved):
Thanks @kayiwa @leefaisonr @precillap

PR #6690, opened 2025-11-24
include all rdss groups in rdss team group

acozine:
I ran a playbook with `--limit rdss` today and noticed that the mflux VMs were not included.

This PR adds some group entries to inventory so that all RDSS VMs roll up into the `rdss` group.


kayiwa (approved):
Ok! Thanks @acozine

PR #6691, opened 2025-11-25
Updated playbook to run as remote user pulsys

aruiz1789:
Updated the remote user and set gather facts to false on third playbook


VickieKarasic (approved):
Thanks @aruiz1789

Issue #6689, opened 2025-11-24
Redirect Reitred Veridian Domain historicperiodicals.princeton.edu


kevinreiss:
### What maintenance needs to be done?
The domain https://historicperiodicals.princeton.edu should be redirected to https://bluemountain.princeton.edu. The domain is already sunset so we should do this as soon as possible. See the [existing domain configuration](https://github.com/pulibrary/princeton_ansible/blob/0e3f9cb42a3ce054df6a5327881ec802a9a6a574/roles/nginxplus/files/conf/http/veridian-sites.conf#L82-L102).

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- The content from the veridian application formerly at https://historicperiodicals.princeton.edu is now part of https://bluemountain.princeton.edu. 

### Acceptance criteria
 - [ ] A request for the root domain https://historicperiodicals.princeton.edu sends you https://bluemountain.princeton.edu
 - [ ] Notify @cwulfman when this is completed.

PR #6692, opened 2025-11-25
Redirect deprecated domain to blue mountain.

kevinreiss:
For #6689.


acozine (commented):
Can we also delete lines 13-16? Do we need to define an upstream for veridian14 once the bluemountain config is changed?

PR #6698, opened 2025-11-26
Update vars.yml

VickieKarasic:
Add correct keys. Follow-up to #6687


christinach (approved):
Thanks @VickieKarasic !

PR #6711, opened 2025-12-05
[iiif] libimages

kayiwa:
related to this https://gitlab.lib.princeton.edu/ops/team-handbook/-/issues/212


roles/iiif/README.md
```diff
@@ -1,22 +1,108 @@
-Role Name
-=========
+# Role Name: iiif
 
-Installs [Samvera Labs](https://github.com/samvera-labs/serverless-iiif)
+Deploys [Samvera Labs serverless-iiif](https://github.com/samvera-labs/serverless-iiif) - a IIIF Image API 2.1/3.0 server running on AWS Lambda.
```
acozine: I'd love to see a description here of what serverless-iiif does in layperson's terms - mostly because I don't entirely understand it. Best I can come up with right now is: 
"Serverless-iiif supports serving images and derivatives from cloud buckets to URLs within our applications using serverless computing platforms in the cloud. For example, images from libimages2 are now stored in an S3 bucket and served by an AWS Lambda."

PR #6745, opened 2025-12-17
[pulfalight] Fix quotations in env variables

hackartisan:


kayiwa (approved):
OK! Thanks @hackartisan

PR #6754, opened 2025-12-18
[pulfalight] Update ruby version

hackartisan:
refs pulibrary/pulfalight#1625

also move vars from role defaults to shared group_vars file, where it's
more explicit and easier to find


kayiwa (approved):
OK! Thanks @hackartisan

PR #6758, opened 2025-12-18
[Deploy_user] accommodate new VM changes

### LINKED ISSUE #6757: SSH Authentication Failure: Deploy user prompting for password during local Capistrano deploy
Local deployments via Capistrano are currently failing. The process hangs or fails because the deploy user on the remote host is prompting for a password instead of authenticating via SSH keys.

### Current Observations

- **User Access:** I am able to log in as pulsys without issue.
- **Configuration Status:** The deploy user keys task (Ansible) is reporting "green" (successful state).
- **Key Verification:** Confirmed that the SSH key handshake between my local machine and GitHub is functioning correctly.

### Steps to Reproduce

1. Initiate a deployment locally using Capistrano (`cap <environment> deploy`). 
2. Observe the process halting at the SSH authentication step with a password prompt.

kayiwa:
our new vms place the ssh_config AllowUsers in a different location. The deploy_user used to be appended to `/etc/ssh/sshd_config` when running this role. The new `AllowUsers` is in a different location. This PR allows both the old format and new format

closes #6757

### DIFF
--- a/roles/deploy_user/tasks/main.yml
+++ b/roles/deploy_user/tasks/main.yml
@@ -1,20 +1,20 @@
 ---
-- name: deploy_user | create system user group
+- name: Deploy_user | create system user group
   ansible.builtin.group:
     name: "{{ deploy_user }}"
     gid: "{{ deploy_user_uid }}"
 
-- name: deploy_user | create system user
+- name: Deploy_user | create system user
   ansible.builtin.user:
     name: "{{ deploy_user }}"
     uid: "{{ deploy_user_uid }}"
     group: "{{ deploy_user }}"
     home: "/home/{{ deploy_user }}"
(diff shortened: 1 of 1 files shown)


hackartisan (changes requested):
This seems to read an old file, scrape it for its contents, and then write those contents to the new file location.

We created the old file initially, right? Can we just create the new file in whatever way we created the old file? If we're not expecting to have this old file moving forward at some point, then how will the value get put in the new file if it's not in the old file to begin with?

Other, less important question, why are we now capitalizing the role under "name"? It seems awkward to have both capitalization and an underscore, and it's nice for it to match the role precisely.

kayiwa:
superceded by #6763

Issue #2739, opened 2022-02-09
[CDH] Check Isilon backups of user uploaded content


kayiwa:
We need to backup user uploaded content on the CDH production websites

PR #6764, opened 2025-12-19
[golang] add golang installer

kayiwa:
for #2739 we need a newer version of Golang than is provided by Ubuntu
  Jammy. This role installs Golang from source.

### DIFF
--- a/roles/golang/molecule/default/verify.yml
+++ b/roles/golang/molecule/default/verify.yml
@@ -16,7 +16,7 @@
 
     - name: Check if Go binary exists at install location
       ansible.builtin.stat:
-        path: "{{ expected_install_dir }}/bin/go"
+        path: "{{ expected_install_dir }}/go/bin/go"
       register: go_binary
 
     - name: Assert Go binary is present and executable
@@ -24,7 +24,7 @@
         that:
           - go_binary.stat.exists
           - go_binary.stat.executable
-        fail_msg: "Go binary not found or not executable at {{ expected_install_dir }}/bin/go"
+        fail_msg: "Go binary not found or not executable at {{ expected_install_dir }}/go/bin/go"
 
     - name: Check if the symlink exists in /usr/local/bin
       ansible.builtin.stat:
@@ -36,8 +36,8 @@
         that:
           - go_symlink.stat.exists
           - go_symlink.stat.islnk
-          - go_symlink.stat.lnk_source == expected_install_dir + '/bin/go'
-        fail_msg: "/usr/local/bin/go is missing or does not point to {{ expected_install_dir }}/bin/go"
+          - go_symlink.stat.lnk_source == expected_install_dir + '/go/bin/go'
+        fail_msg: "/usr/local/bin/go is missing or does not point to {{ expected_install_dir }}/go/bin/go"
 
     - name: Get installed Go version
       ansible.builtin.command: /usr/local/bin/go version


roles/golang/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+---
+
+## What it does
+
+On each run, the role:
+
+1. Ensures the download directory (`{{ golang_download_dir }}`) exists.
+2. Derives the correct architecture string (`amd64` / `arm64`) from
+   `ansible_architecture` if you didn’t override `golang_arch`.
```
acozine: Why would I want to override `golang_arch`? Shouldn't the discovery of the architecture always work?

roles/golang/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  ansible.builtin.unarchive:
+    src: "{{ golang_download_dir }}/go{{ golang_version }}.linux-{{ golang_arch }}.tar.gz"
+    dest: "/usr/local"
+    remote_src: true
+  when: golang_tarball.changed
+
+- name: Golang | Ensure /usr/local/bin/go symlink exists
+  ansible.builtin.file:
+    src: "{{ golang_install_dir }}/bin/go"
+    dest: "/usr/local/bin/go"
```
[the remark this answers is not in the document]
kayiwa: I was wondering what I didn't spot yesterday, then I ran against my sandbox and it works like I expect. The short version is: we do want `dest: /usr/local` here, not `{{ golang_install_dir }}`, and the symlink is not pointing to itself.

A bit more detail:

* The official Go tarballs are laid out like this inside the archive:

    ```text
    go/
      bin/go
      src/...
      pkg/...
    ```
   
* The Go install docs use:
  ```bash
  tar -C /usr/local -xzf go1.xx.x.linux-amd64.tar.gz
  ```
  which results in `/usr/local/go/...

* Our defaults match that:
  ```yaml
  golang_install_dir: "/usr/local/go"
  ```

* In the role:
  ```yaml
  - name: Golang | Install Go {{ golang_version }}
  ansible.builtin.unarchive:
    src: "{{ golang_download_dir }}/go{{ golang_version }}.linux-{{ golang_arch }}.tar.gz"
    dest: "/usr/local"
    remote_src: true
  ```
  This creates `/usr/local/go/bin/go`, `/usr/local/go/src`, etc.

* And the symlink task is:
  ```yaml
  - name: Golang | Ensure /usr/local/bin/go symlink exists
  ansible.builtin.file:
    src: "{{ golang_install_dir }}/bin/go"   # /usr/local/go/bin/go
    dest: "/usr/local/bin/go"
    state: link
   ```
    So we end up with:
   ```text
   /usr/local/go/bin/go        # real binary
   /usr/local/bin/go           # symlink pointing to the above
   ```

If we changed `dest to {{ golang_install_dir }}` (i.e. `/usr/local/go`) while still unpacking a tarball that already contains a top-level `go/` directory, we’d wind up with `/usr/local/go/go/bin/go`, which would be incorrect.

I added a small comment above the unarchive task to make this clearer, e.g. “Go tarball contains a top-level go/ directory; extract into /usr/local to get /usr/local/go per upstream docs.”

roles/golang/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+- name: Golang | Install Go {{ golang_version }}
+  ansible.builtin.unarchive:
+    src: "{{ golang_download_dir }}/go{{ golang_version }}.linux-{{ golang_arch }}.tar.gz"
+    dest: "/usr/local"
+    remote_src: true
+  when: golang_tarball.changed
+
+- name: Golang | Ensure /usr/local/bin/go symlink exists
+  ansible.builtin.file:
+    src: "{{ golang_install_dir }}/bin/go"
```
acozine:
```suggestion
    src: "{{ golang_install_dir }}/go/bin/go"
```

roles/golang/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    mode: "0644"
+    force: false
+    checksum: "{{ golang_tarball_checksum | default(omit) }}"
+  register: golang_tarball
+
+# Go tarball contains a top-level go/ directory; extract into /usr/local to get /usr/local/go per upstream docs.
+- name: Golang | Install Go {{ golang_version }}
+  ansible.builtin.unarchive:
+    src: "{{ golang_download_dir }}/go{{ golang_version }}.linux-{{ golang_arch }}.tar.gz"
+    dest: "/usr/local"
```
acozine:
```suggestion
    dest: "{{ golang_install_dir }}"
```

roles/golang/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+golang_version: "1.25.5"
+
+# Architecture string for Go tarball. Override if needed. (for local molecule? building)
+golang_arch: "amd64"
+
+# Where to cache downloaded tarballs
+golang_download_dir: "/usr/local/src"
+
+# Where Go will be installed
+golang_install_dir: "/usr/local/go"
```
acozine:
```suggestion
golang_install_dir: "/usr/local"
```

PR #6765, opened 2025-12-19
[plakar] install and configure plakar

### LINKED ISSUE #2739: [CDH] Check Isilon backups of user uploaded content

kayiwa:
This installs and configures plakar. 

This role can go in ONLY after https://github.com/pulibrary/princeton_ansible/pull/6764 is merged. So it will fail on CI and a rebase from main will fix that. 

closes #2739


group_vars/nfsserver/staging.yml
```diff
@@ -1,4 +1,13 @@
 ---
+# plakar backups
+plakar_store_name: "nfs_aws"
+plakar_repo_passphrase: "{{ vault_plakar_repo_passphrase }}"
+plakar_repo_bucket: "pul-nfs-backup"
```
acozine: For var values that do not change from one env to another, can we use a `group_vars/nfsserver/shared.yml` file so we would not have to change them in three places if they ever needed to change?

roles/plakar/tasks/configure_backup.yml
```diff
@@ -0,0 +1,124 @@
...
+
+# Make sure the snapshot root exists
+- name: Plakar | Ensure snapshot root directory exists
+  ansible.builtin.file:
+    path: "{{ plakar_snapshot_root }}"
+    state: directory
+    mode: "0755"
+
+# Check if the AWS store is already configured
+- name: Plakar | Check if AWS store exists
```
acozine: We talked about making the role smart enough to know if a machine is already being backed up or not, so we could run it regularly for maintenance. So we need logic for "is this thing already being backed up" - if yes, then keep the default `false` settings for configure backups and run initial backup; if no, set those to `true` and run the rest of the role. This task might be a good spot to introduce that logic.

acozine (commented):
First day back, so my brain is slow. But here are a few thoughts.

acozine (commented):
A few minor questions/updates.

kayiwa:
this branch has diverged significantly.

PR #6797, opened 2026-01-06
add staging and production vms

### LINKED ISSUE #6794: [Repec] add a second VM in each environment
As an operations engineer, I want to know I can reboot one VM without taking an entire service offline. Having two VMs for each service, in each environment, supports this goal.

The repec service currently only has one production VM (repec-prod2) and one staging VM (repec-staging1). Add a second VM in each environment.

## Acceptance criteria

- [ ] we have two VMs in each environment in vSphere
- [ ] all four VMs have matching resources (currently the staging VM has 16GB of memory; prod has 8GB of memory, only 1GB is in use)
- [ ] the VMs are built out using the repec playbook
- [ ] the inventory includes all four VMs
- [ ] the nginxplus configuration files include all VMs 
- [ ] the `passenger_server_name` variable points to a URL, not an individual VM

### Relevant deadlines
N/A

### Describe the VM(s) you need

If you do not know the answers to some of these questions, that's okay. Please add the information you do have, and we will work together to gather any other information we need. You can also look at the [documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/new_vms_for_devs.md) in pul-it-handbook.

Which project are these VMs for?
- [x] Project: repec-prod, repec-staging

How many VMs do you need?
- [x] Number of VMs: 2

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [x] FQDN: repec-prod1.lib.princeton.edu, repec-staging2.lib.princeton.edu

What virtual resources does each VM need?
- [x] Number of CPUs per VM: 2
- [x] Amount of memory per VM (in GB): 8
- [x] Amount of storage/disk space per VM (in GB): 30

What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [x] Port number(s): 22, 80, 443
- [x] Service(s): SSH, HTTP, HTTPS

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [x] Additional resources: none

It would be great to migrate the pre-existing repec VMs to the private network as part of this ticket.

kayiwa:
we add a staging vm in the private network and a production VM

closes #6794

### DIFF
--- a/inventory/all_projects/repec
+++ b/inventory/all_projects/repec
@@ -1,5 +1,6 @@
 [repec_staging]
 repec-staging1.princeton.edu
+repec-staging2.lib.princeton.edu
 [repec_production]
-# repec-prod1.princeton.edu
+repec-prod1.princeton.edu
 repec-prod2.princeton.edu
--- a/roles/nginxplus/files/conf/http/dev/repec-staging.conf
+++ b/roles/nginxplus/files/conf/http/dev/repec-staging.conf
@@ -12,6 +12,7 @@ limit_req_zone $external_traffic zone=repec-staging-ratelimit:10m rate=10r/s;
 upstream repec-staging {
     zone repec-staging 64k;
     server repec-staging1.princeton.edu resolve;
+    server repec-staging2.lib.princeton.edu resolve;
     sticky learn
           create=$upstream_cookie_repecstagingcookie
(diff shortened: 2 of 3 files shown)


christinach (approved):
Thanks @kayiwa !

Issue #6812, opened 2026-01-12
Fix Mailcatcher Installation: `Gem::UnknownCommandError: Unknown command query`


kayiwa:
The Ansible playbook for `pdc_describe` is failing during the mailcatcher installation task. It appears that the `gem` command being executed (`gem query`) is no longer supported or is behaving unexpectedly on the version of Ruby/RubyGems installed on the staging servers (Ruby 3.3.0).

Error log

```text
TASK [mailcatcher : mailcatcher | install global mail catcher] ********************
fatal: [pdc-describe-staging1.princeton.edu]: FAILED! => {
    "changed": false, 
    "cmd": "/usr/local/bin/gem query --norc -n '^mailcatcher$'", 
    "msg": "ERROR: While executing gem ... (Gem::UnknownCommandError)\n Unknown command query"
}
```

### Steps to Reproduce

  1. Run the pdc_describe playbook against the staging environment:

    ```bash
    ansible-playbook playbooks/pdc_describe.yml -l pdc_describe_staging
    ```

  2. The task `mailcatcher | install global mail catcher` will fail.

#### Relevant File
`roles/mailcatcher/tasks/main.yml:11`

```yaml
- name: mailcatcher | install global mail catcher
  gem:
    name: mailcatcher
    state: present
    user_install: false
```

#### Proposed Fix
We may need to:

1. Update the Ansible version to ensuring the `gem` module is compatible with Ruby 3.3.0.

2. Or, bypass the Ansible gem module and use a command task with gem install mailcatcher if the module remains incompatible.

3. Ensure rubygems-invoker or similar dependencies are updated on the target hosts.

#### Acceptance Criteria

  - [ ] The pdc_describe playbook runs to completion without failure.

  - [ ] Mailcatcher is successfully installed and available globally on staging hosts.

PR #6815, opened 2026-01-12
[mailcatcher] avoid break systemd unit file

kayiwa:
ensure our mailcatcher role remains idempotent and ensure that our systemd unit file will continue to work

related to #6812


kayiwa:
superseded by #6832

Issue #6768, opened 2025-12-19
[Tower] Expects private network (.lib) in Replace a VM playbook


VickieKarasic:
### What maintenance needs to be done?

In the Replace a VM (Originally in Staging) playbook, this task fails:

```
TASK [Wait for VM to be reachable via SSH on derived FQDN] *********************
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": 300, "msg": "Timeout when waiting for ezproxy-test.lib.princeton.edu:22"}
```
because it expects the machine to be on the private network (.lib). We need to fix the template so that this isn't the default. 

The playbook then fails on 
```
TASK [Detect OS type on new VM]
Failed to connect to the host via ssh: ssh: Could not resolve hostname ezproxy-
test.lib.princeton.edu: Name or service not known
fatal: [localhost -> ezproxy-test.lib.princeton.edu]: UNREACHABLE! => {"changed": false, "msg": "Data could not be sent to remote host \"ezproxy-test.lib.princeton.edu\". Make sure this host can be reached over ssh: ssh: Could not resolve hostname ezproxy-test.lib.princeton.edu: Name or service not known\r\n", "unreachable": true}
```

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
We shouldn't have the default as the private network so that the playbook succeeds. 

### Acceptance criteria
 - [ ] VM being replaced should be reachable when not on the private network 
 - [ ] Playbook should succeed

VickieKarasic:
Thanks @acozine. I think making the DNS extension a field in the template is worth a try. If we haven't changed the template in the past week, I had made a [copy](https://ansible-tower.princeton.edu/#/templates/job_template/99/details) of it last week that we can use to experiment.

PR #6816, opened 2026-01-12
[ruby_app] make our ruby_apps idempotent

kayiwa:
part of #6768 that makes sure all our roles are idempotent


VickieKarasic (approved):
thanks @kayiwa

PR #6823, opened 2026-01-14
[Analytics] Migrate Analytics back to Umami + Upgrade

tpendragon:
Rybbit was nice, but there's not a ton of support, and a new major version of Umami is something we want to evaluate.


kayiwa (approved):
OK! Thanks @tpendragon

Issue #6789, opened 2025-12-24
Standardize Molecule Test Sequence (Add Idempotence & Cleanup)


kayiwa:
Currently, our Ansible Molecule test scenarios are running a simplified lifecycle. While this verifies that the roles run successfully, we are missing critical checks for idempotence (ensuring the role can run multiple times without unintended changes) and side effects.

We need to update our molecule.yml scenarios to use the standard, comprehensive test sequence.

Current Behavior Our current sequence skips the idempotence, side_effect, and initial cleanup/destroy phases:

```yaml
scenario:
  name: default
  test_sequence:
    - dependency
    - syntax
    - create
    - prepare
    - converge
    - verify
    - destroy
```

Update the test_sequence in molecule.yml to the standard lifecycle. This ensures:

Robustness: Cleaning up previous runs before starting (cleanup, destroy).

Idempotency: Re-running the converge step to ensure no changes are reported on the second run (idempotence).

```yaml
scenario:
  name: default
  test_sequence:
    - dependency
    - cleanup
    - destroy    
    - syntax
    - create
    - prepare
    - converge
    - idempotence   # Added (Critical)
    - side_effect   # Added
    - verify
    - cleanup 
    - destroy
```

PR #6827, opened 2026-01-15
remove side_effect from our molecule

kayiwa:
we do not typically use this in our testing. 

related to #6789


VickieKarasic (approved):
Looks good @kayiwa - thanks!

PR #6832, opened 2026-01-16
[mailcatcher ] make mailcatcher role idempotent

kayiwa:
related to #6789

### DIFF
--- a/roles/mailcatcher/templates/mailcatcher.service.j2
+++ b/roles/mailcatcher/templates/mailcatcher.service.j2
@@ -6,8 +6,8 @@ After=syslog.target network.target
 Type=simple
 WorkingDirectory=/home/{{ mailcatcher_user }}
 ExecStart=/bin/bash --login -c "{{ mailcatcher_install_location }} --foreground"
-User={{ mailcatcher_user | default('root') }}
-Group={{ mailcatcher_group | default('root') }}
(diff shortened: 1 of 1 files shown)


roles/mailcatcher/templates/mailcatcher.service.j2
```diff
@@ -4,10 +4,10 @@ After=syslog.target network.target
...
 [Service]
 Type=simple
-WorkingDirectory=/home/{{mailcatcher_user}}
-ExecStart=/bin/bash --login -c "{{mailcatcher_install_location}} --foreground"
-User={{mailcatcher_user}}
-Group={{mailcatcher_group}}
+WorkingDirectory=/home/{{ mailcatcher_user }}
+ExecStart=/bin/bash --login -c "{{ mailcatcher_install_location }} --foreground"
+User={{ mailcatcher_user | default('root') }}
+Group={{ mailcatcher_group | default('root') }}
```
acozine: We already set `root` as the default in our `defaults/main.yml` file. Let's define it only once - my vote would be in the defaults file, so my suggestion is:

```suggestion
User={{ mailcatcher_user }}
Group={{ mailcatcher_group }}
```

acozine (commented):
Thanks for bringing our molecule tests into the modern era!

One minor suggestion on this PR.

PR #6839, opened 2026-01-22
adds all sandbox VMs in vsphere

acozine:
Partially addresses #3897.

Puts sandbox VMs in their own group.


acozine:
As of today, gpmenos-sandbox, sandbox-ad, and sandbox-rl36771 are unreachable.

Issue #6848, opened 2026-01-26
Reinstall `lib-solr-prod{1,2,3}` with Reduced Root Disk Size


kayiwa:
### Environment:

    Host: `lib-solr-prod{1,2,3}`

### Issue Description

The root filesystem on lib-solr-prod{1,2,3} is currently provisioned with a 1TB logical volume, of which only ~357G is in use. This over-allocation of disk space is inefficient and consumes unnecessary storage resources. Additionally, the system may benefit from a clean reinstallation to ensure configuration consistency and remove any potential technical debt.
Requested Action

  - Reinstall the server with conservative OpenZFS /solr mount.
  - Reduce the root disk size to a more appropriate capacity — recommended: 400G (provides ample room -when sunset collections are removed for growth while reducing waste).
  - Ensure all required services (Solr, backup mounts, etc.) are reconfigured and validated post-reinstall.
  - Confirm /solr/data/cloud_backup (GCS FUSE mount) is preserved or re-established as needed.
  - Perform full post-install validation (service health, indexing, replication, monitoring).

### Justification

Cost & Resource Efficiency: Reduces underlying storage footprint.
Maintainability: Clean install improves system hygiene and reproducibility.
Future-Proofing: 500G provides >3x current free space with room for logs, updates, and temporary spikes.

> Notes

  -  Coordinate during a maintenance window to minimize impact.
  -  Confirm backups are current before proceeding.
  -  Revalidate monitoring and alerting post-reinstall.

PR #6849, opened 2026-01-26
[solr9cloud] ruby version

kayiwa:
related to #6848
there is a discrepancy with the ruby version on pul_solr


christinach (approved):
Thanks @kayiwa 🦒

PR #6851, opened 2026-01-27
monitor jvm on solr8 via checkmk with jolokia agent

aruiz1789:
This PR is part of the JVM monitoring with checkmk.. It will add the jar file to the files directory, copy it to the webapp jar directory and update the solr app opts to include the agent at startup.


acozine (approved):
Thanks for doing this work!

How often has the jolokia-agent been released, historically, and how long do they keep old releases available? If it's constantly changing and they don't maintain old versions, we might want to make the version a variable. However, we can make that change in a separate PR.

I'll merge this now, we can revisit if the jolokia release cadence turns out to be quick.

Issue #1774, opened 2020-10-16
Create a role fro libruby-prod


carolyncole:
Multiple ruby projects are hosted on libruby-prod.  At this point I am uncertain how we maintain the box and the code that runs on it...

One of the code repos is very very out of date...

https://github.com/pulibrary/repecwp/blob/main/config/deploy.rb

acozine:
@carolyncole is this still relevant?

acozine:
Ah, @kayiwa confirms that libruby-prod has been deleted. Closing as obsolete.

PR #6861, opened 2026-01-30
Remove the obsolete libruby playbook

acozine:
We don't need the libruby playbook any more. We do not currently have inventory groups for `libruby_<env>`, so the playbook would not find any hosts to run against. And even if it did run, it would only add the deploy user . . .

Per https://github.com/pulibrary/princeton_ansible/issues/1774#issuecomment-1015667627, the libruby-prod project was decommissioned.

This PR removes the playbook entirely.


kayiwa (approved):
OK! Thanks @acozine

Issue #6855, opened 2026-01-30
PostgreSQL Connection Exhaustion (`Too many clients already`)


kayiwa:
Related to #2305.

The production database is currently hitting its max_connections limit, resulting in `ActiveRecord::ConnectionNotEstablished` errors. This is causing intermittent downtime for the Rails application and background workers.

### Error logs

```plaintext
ActiveRecord::ConnectionNotEstablished: connection to server at "128.112.203.142", port 5432 failed: 
FATAL: remaining connection slots are reserved for non-replication superuser connections

ActiveRecord::ConnectionNotEstablished: connection to server at "128.112.203.142", port 5432 failed: 
FATAL: sorry, too many clients already
```

#### Current Configuration Analysis

- Current max_connections: 400
- Current shared_buffers: 16066MB (~16GB)
- Current work_mem: 40MB

kayiwa:
### Proposed Changes (Tuning)

1. Database Configuration (postgresql.conf via Ansible)
To provide immediate headroom and prevent administrator lockout:
  - [ ] Increase `max_connections` from `400` to `800`
  - [ ] Set `superuser_reserved_connections = 5` (currently commented/default).
  - [ ] Implement TCP keepalive settings to reap "zombie" connections from crashed application processes:
    -  [ ] `tcp_keepalives_idle = 60`
    -  [ ] `tcp_keepalives_interval = 10`
    -  [ ] `tcp_keepalives_count = 6`


Until this is implemented. The internet suggests this. 

```sql
SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE state = 'idle';
```

PR #6864, opened 2026-02-02
[postgresql] adjust the tcp connections for postgresql

kayiwa:
we are adjust the connection to stop relying on the underlying OS setting for TCP connections. This (we hope) will stop zombie connections from using up the postgresql connection

related to #6855


kayiwa:
ran this against the production databases.

PR #6870, opened 2026-02-03
[Nomad] Get molecule setup able to run jobs again.

tpendragon:
I wanted to start trying upgrading Nomad in Molecule, but found that the setup couldn't run jobs anymore. This gets it so that after `molecule converge`, you can go to localhost:4646 and submit that redis job.

It also doesn't end up with a working `dnsmasq` setup, I think because `nmcli` isn't installed in those docker images, but this at least moves things forward.


kayiwa (approved):
OK! Thanks @tpendragon

PR #6877, opened 2026-02-06
clean figgy libimages2 mount reference

aruiz1789:
removed smb file creds and role reference, also removed groupvars for "figgy_fallback_derivatives_mount".


tpendragon (changes requested):
Blocked by https://github.com/pulibrary/figgy/issues/7036.

I'm not convinced things would keep working on a newly generated box if we got rid of these, please let us remove references to that fallback directory first.

tpendragon (approved):

PR #6889, opened 2026-02-11
[postgresql] adjust postgresql max connections

kayiwa:
related to #6855 we are getting tons of max_connections errors we double the maximum connections while explicitly reserving the maximum super user required connections


christinach (approved):
Thanks @kayiwa 🦒

Issue #4135, opened 2023-08-10
[BIBDATA] cleanup Remove SCSB_S3_BUCKET_NAME from bibdata


christinach:
Remove the following env, It is not used.
- name: SCSB_S3_BUCKET_NAME
    value: 'alma-recap-production'

aruiz1789:
@christinach Can we also delete the AWS S3 Bucket?

```
ruiza@LIB-FMC6NRWR1L ~ % aws s3 ls --summarize --human-readable --recursive s3://alma-recap-production

Total Objects: 0
   Total Size: 0 Bytes
```

PR #6908, opened 2026-02-18
rm var from production.yml

aruiz1789:
Var has been deleted and empty bucket will also be deleted from AWS.

(devbox) ruiza@LIB-FMC6NRWR1L princeton_ansible % aws s3 ls --summarize --human-readable --recursive s3://alma-recap-production

Total Objects: 0
   Total Size: 0 Bytes


kayiwa (approved):
OK! Thanks @aruiz1789

aruiz1789:
Related to this issue: https://github.com/pulibrary/princeton_ansible/issues/4135

PR #6909, opened 2026-02-18
[cdh] move prosody to dev load balancer

kayiwa:
we are moving most non-production properties to our staging environment

related to this ticket https://gitlab.lib.princeton.edu/ops/team-handbook/-/issues/259


kayiwa:
removed certs from production and run on the dev load balancers.

PR #6911, opened 2026-02-19
[cdh] move cdh-web to staging load balancer

kayiwa:
related to https://gitlab.lib.princeton.edu/ops/team-handbook/-/issues/259

moving non-production web apps to staging loadbalancer


VickieKarasic (approved):
Ok - thanks @kayiwa!

PR #6916, opened 2026-02-23
Rename EZproxy vaulted files

acozine:
Follow up to #6906.

The vaulted files in the EZproxy role were vulnerable to getting committed without being encrypted.

This PR:
- adds `vault` to the names of two vaulted files
- updates the task that calls those files with the new filename pattern
- updates the README with the new filename pattern
- removes three other vaulted files that we don't seem to be using:
  - we swtiched from `files/princeton_allow.txt` to `templates/princeton_allow.txt.j2`
  - we switched from `files/user.txt` to `templates/user.txt.j2`
  - we weren't using `files/id_rsa` at all that I can tell


rladdusaw (approved):
Thanks @acozine!

PR #6923, opened 2026-02-25
[cdh] add keys should be on lib-solr-staging2

kayiwa:
we moved the [lead solr server](https://github.com/pulibrary/pul_solr/commit/b5b4cbbde0c39cf0138442fcd18e945aa729052f) to lib-solr-staging2 and failed to add this to the inventory. As a result the cdh keys are missing on the server.

This fixes that.

related to https://github.com/Princeton-CDH/cdh-ansible/pull/305

### DIFF
--- a/inventory/all_projects/cdh
+++ b/inventory/all_projects/cdh
@@ -38,11 +38,13 @@ lib-ponyexpr-prod.princeton.edu
 lib-postgres-prod1.princeton.edu
 lib-postgres-prod3.princeton.edu
 lib-solr-prod1.princeton.edu
-lib-solr-prod7.princeton.edu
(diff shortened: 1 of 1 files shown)


inventory/all_projects/cdh
```diff
@@ -44,5 +44,5 @@ adc-dev1.lib.princeton.edu
 adc-dev2.lib.princeton.edu
 lib-fs-staging.princeton.edu
 lib-postgres-staging1.princeton.edu
-lib-solr-staging1.princeton.edu
+lib-solr-staging2.princeton.edu
```
rlskoeser: It looks like this is replacing solr-staging1 with solr-staging2, should they both be listed?

PR #6937, opened 2026-03-04
[allsearch-frontend] Make sure it can run on a fresh VM

sandbergja:
If you run this playbook on a recently built machine, you get the error:

```
Destination directory /opt/allsearch_frontend/shared does not exist
```

See [this failed job on tower](https://ansible-tower.princeton.edu/#/jobs/playbook/42144/output)

This makes sure that the path exists before attempting to create the file.


regineheberlein (approved):
thanks @sandbergja !

PR #6939, opened 2026-03-04
[Nomad] Update Nomad to 1.11.1

tpendragon:
The big breaking change here is that Nomad now uses "workload identities" to register services in consul: https://developer.hashicorp.com/nomad/tutorials/integrate-consul/consul-acl

* Creates workload identity policies
* Sets a default identity
* Configures consul to allow Nomad as an auth provider
* Associates policies appropriately to those workload identities
* Configures a host_volumes_dir to add support for dynamic host volumes.

This also switches podman's internal DB to use sqlite instead of BoltDB, because we were having locking problems and BoltDB was deprecated anyways.

### DIFF
--- a/roles/pul_nomad/templates/nomad/client.hcl.j2
+++ b/roles/pul_nomad/templates/nomad/client.hcl.j2
@@ -20,6 +20,7 @@ client {
     {% endfor -%}
   }
   {% endif %}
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK! Thanks @tpendragon

PR #7000, opened 2026-03-27
removes oar-dev artifacts

acozine:
Paritally addresses #6783.

Removes the key for the `oar-dev` site.
Removes the configuration for the `oar-dev` site.


acozine:
Closing as obsolete. The oar-dev keys are gone and we don't have an obsd_httpd role any more.

PR #7028, opened 2026-04-10
[pdc-describe] Upgrade node and yarn for pdc_describe

bess:
Ref https://github.com/pulibrary/pdc_describe/issues/2234


leefaisonr (approved):
Thanks @bess

PR #7045, opened 2026-04-21
[pul_nomad] update cache

kayiwa:
the task that installs iproute2 fails without an update cache this unbreak main branch


christinach (approved):
thanks @kayiwa ! 🦒

PR #7048, opened 2026-04-22
[unbreak main] drop remote_tmp and /tmp from molecule

kayiwa:
ansible-core 2.19 surfaces a tmpdir race in the Docker connection plugin when remote_tmp points at /tmp and /tmp is also mounted as tmpfs. Gathering Facts fails with "dd: failed to open ... No such file or directory" because systemd-tmpfiles and Ansible's mkdir race on container boot, leaving the ansible-tmp-* subdirectory missing when dd runs.

Removing the remote_tmp override lets Ansible fall back to /root/.ansible/tmp, which the image already provisions at mode 777 on the overlay FS. Dropping /tmp from the tmpfs list keeps /run and /run/lock as tmpfs (systemd still needs those) while letting /tmp live on the overlay FS where it is not touched by systemd on boot.

no behavioral change for the roles themselves and only affects the molecule (currently failing)

ref: https://github.com/pulibrary/vm-builds/pull/59

### DIFF
--- a/roles/ad_join/molecule/default/molecule.yml
+++ b/roles/ad_join/molecule/default/molecule.yml
@@ -26,7 +26,6 @@ platforms:
     tmpfs:
       - /run
       - /run/lock
-      - /tmp
     volumes:
       - /sys/fs/cgroup:/sys/fs/cgroup:rw
     cgroupns_mode: host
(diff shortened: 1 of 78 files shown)


VickieKarasic (approved):
Thanks @kayiwa

Issue #7047, opened 2026-04-22
PostgreSQL outage caused by invalid pg_hba.conf entries using hostname with CIDR mask


kayiwa:
PostgreSQL failed to start because pg_hba.conf contained invalid host entries. The generated address field combined DNS hostnames with CIDR suffixes such as:

```conf
nomad-client-staging3.lib.princeton.edu/32
nomad-client-prod1.lib.princeton.edu/32
nomad-client-prod2.lib.princeton.edu/32
nomad-client-prod3.lib.princeton.edu/32
nomad-client-prod4.lib.princeton.edu/32
```
This is not valid PostgreSQL pg_hba.conf syntax. In the address column, PostgreSQL accepts either:

- a hostname by itself, for example:
  `nomad-client-staging3.lib.princeton.edu`
- or an IP/network in CIDR notation, for example:
  `192.0.2.15/32`

It does not accept a hostname with a CIDR mask appended.

Because of these invalid entries, PostgreSQL could not parse pg_hba.conf, failed startup, and shut down.

### Impact

PostgreSQL was unavailable after startup/reload attempted to read the invalid pg_hba.conf. This caused a database outage until the bad entries were corrected.

Observed errors

```sh
specifying both host name and CIDR mask is invalid: "nomad-client-staging3.lib.princeton.edu/32"
specifying both host name and CIDR mask is invalid: "nomad-client-prod1.lib.princeton.edu/32"
specifying both host name and CIDR mask is invalid: "nomad-client-prod2.lib.princeton.edu/32"
specifying both host name and CIDR mask is invalid: "nomad-client-prod3.lib.princeton.edu/32"
specifying both host name and CIDR mask is invalid: "nomad-client-prod4.lib.princeton.edu/32"
FATAL: could not load pg_hba.conf
database system is shut down
```

PR #7049, opened 2026-04-22
Fix pg_hba rules for Nomad playbooks.

tpendragon:
Work towards #7047


kayiwa (approved):
🎉  thanks @tpendragon

PR #7058, opened 2026-04-27
Proquest drop

kayiwa:
adds ansible management of the proquest server.


bess (approved):
Enthusiastically approved! Thank you so much for doing this!

PR #7066, opened 2026-04-28
[lib-jobs] Add an environment variable for the email addresses that should receive TMAS error emails

sandbergja:
Helps with https://github.com/pulibrary/lib_jobs/issues/1009


christinach (approved):
Thanks @sandbergja 🍄

PR #7072, opened 2026-04-29
[Analytics] Further improve security, disable filesystem write

tpendragon:
I think this application only takes in requests and forwards to postgres. If so, it doesn't need to write to its file system.

This is a bigger change than #7071, but it's deployed now to staging and I'd like to let it cook for a bit.

When connecting to the container, now you get this:

/app $ touch bla
touch: bla: Read-only file system
/app $

This also applies a number of recommendations from https://dev.to/teguh_coding/docker-security-hardening-10-practices-that-will-protect-your-containers-in-production-4ahe to prevent privilege escalation from within the container.

### DIFF
--- a/nomad/analytics/deploy/staging.hcl
+++ b/nomad/analytics/deploy/staging.hcl
@@ -59,6 +59,11 @@ job "analytics-staging" {
         security_opt = [
           "no-new-privileges"
         ]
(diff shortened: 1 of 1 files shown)


kayiwa (approved):
OK! thanks @tpendragon

PR #7078, opened 2026-04-30
[Allsearch-API] Add allsearch api servers to the allowed list to access the staff report

rladdusaw:
The staff index in Allsearch has gone stale since it can no longer query Lib-Jobs for the staff report. Adding staging and production allsearch-api server IPs to the list of allowed IPs.


sandbergja (commented):
Thanks, @rladdusaw ! 🐈‍⬛   Is there any way to do this by hostname rather than IP address?  It seems potentially annoying to need to update this list any time we rebuild an allsearch box.

If there is no way to do that, could we at least document that requirement somewhere?

rladdusaw:
> Thanks, @rladdusaw ! 🐈‍⬛ Is there any way to do this by hostname rather than IP address? It seems potentially annoying to need to update this list any time we rebuild an allsearch box.
> 
> If there is no way to do that, could we at least document that requirement somewhere?

The allow directive I'm using only supports IP. Where would be a good place to document this?

sandbergja:
> Where would be a good place to document this?

Maybe as a `post_install` in the allsearch_api playbook?

sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

Issue #7025, opened 2026-04-09
Implement Certificate-Based Authentication + MFA for SSH/Server Access


kayiwa:
The traditional security model assumes that if an attacker gains control of a developer’s machine, the blast radius can be contained. However, with the rise of AI‑assisted coding and terminal usage, we can no longer safely assume that every command executed on a server is intentional and understood by the developer.

AI agents, auto‑completion tools, or compromised IDE plugins could inject or suggest malicious commands without the user’s explicit awareness. This changes the threat model: we need to move beyond “trusting the human at the keyboard” and toward stronger, enforceable authentication for every sensitive action.

## Proposed Solution
We want to adopt a two‑pronged approach:

1. **Certificate‑based authentication** (e.g., SSH short‑lived certificates via native SSH CA, opkssh, Hashicorp Vault?) – replaces static SSH keys, enables per‑login scoping and expiry.
2. **MFA (Multi‑Factor Authentication)** – required at login time (e.g., TOTP, EntraID/Duo) to ensure that even if credentials/certificates are stolen from an AI‑compromised machine, the attacker cannot authenticate without the second factor. EntraID?

## Goals

- Every interactive login or command execution (e.g., sudo, ssh) is bound to a personal, non‑shared identity.
- MFA is enforced before a session is established.
- Certificate validity is short (e.g., ≤ 4 hours) to reduce replay risk.
- Audit logs clearly link every action to both a user identity and MFA challenge event.

## Non‑Goals (for this ticket)

- Replacing all existing key‑based auth overnight – we can phase this in.
- MFA for every single command (that could break automation); focus is on interactive human sessions.

kayiwa:
Notes on [RHEL Domain Services](https://learn.microsoft.com/en-us/entra/identity/domain-services/join-rhel-linux-vm?tabs=rhel)?

kayiwa:
Notes on [Ubuntu Domain Services](https://learn.microsoft.com/en-us/entra/identity/domain-services/join-ubuntu-linux-vm?utm_source=chatgpt.com)

tpendragon:
I saw a post once where Vault can act as a CA, so there was a command to have vault sign your public key with a short ttl and then allow the ssh connection, and then vault had Entra auth, but I can't find it now, all the examples I find are much more complicated than that or don't lead with what the end result is

If OIT has a vaulty thing maybe that's an option

kayiwa:
perhaps this is what we want.

[...]

PR #7079, opened 2026-04-30
[openssh] add entraid login role

kayiwa:
Create an entraid_join role for headless Microsoft Entra ID authentication on Ubuntu and Rocky Linux using Himmelblau. The role installs the appropriate Himmelblau packages and repositories, configures PAM/NSS, enables keyboard-interactive SSH authentication, and writes the Himmelblau configuration for the Princeton Entra domain.

This is needed so Linux users can log in over SSH with their Entra credentials without requiring a GUI workflow or Microsoft Entra Domain Services. The role also handles Rocky-specific SSH behavior by disabling the Red Hat ChallengeResponseAuthentication override, and it comments out OpenSSH AllowUsers entries during the pilot because they block UPN-style Himmelblau logins.

Side effects: SSH password and keyboard-interactive authentication are enabled for the pilot, and existing AllowUsers restrictions are commented out so Entra access should be restricted through Himmelblau/PAM group controls instead.

ref: #7025

### DIFF
--- a/playbooks/entraid_join.yml
+++ b/playbooks/entraid_join.yml
@@ -1,0 +1,7 @@
+---
+- name: enable ssh via entraid
+  hosts: sandbox-{{ runtime_env | default('ad') }}.lib.princeton.edu
+  remote_user: pulsys
+  become: true
+  roles:
+    - role: entraid_join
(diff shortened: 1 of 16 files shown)


kayiwa:
Tests here https://pulibrary.slack.com/archives/C9ZM9RX4J/p1777491507104269

VickieKarasic (approved):
This is great work - thanks @kayiwa!

PR #7086, opened 2026-05-05
I7085 update common

### LINKED ISSUE #7085: add FreeBSD in maintenance
our Common Role and os updates expect OpenBSD which we have since sunset.

kayiwa:
Ensures our FreeBSD hosts are updated.

closes #7085


VickieKarasic (approved):
Thanks @kayiwa

PR #7090, opened 2026-05-05
[common] make mirrors.yml idempotent

kayiwa:
- Use fixed path for sources.list backup (was timestamped, always changed)
- Remove timestamp comment from 99-ubuntu-archive.list content
- Bump cache_valid_time from 0 to 3600
- Remove duplicate mirrors.yml include from tasks/main.yml

i7085: redis - rename pul_redis filename and simplify cache flow

- Rename apt_repository filename from pul_redis to redis to avoid collision with common role's legacy-PUL-prefixed file scrubbing
- Replace manual apt-get clean/update block with apt_repository's built-in update_cache: true (idempotent, surfaces real errors)
- Add legacy-cleanup task to migrate hosts with old filenames
- Remove changed_when: false from redis.conf template task so config changes properly notify the restart handler

### DIFF
--- a/roles/common/tasks/mirrors.yml
+++ b/roles/common/tasks/mirrors.yml
@@ -32,7 +32,6 @@
   loop:
     - /etc/apt/sources.list.d/00-princeton.list
     - /etc/apt/sources.list.d/pul_postgresql.list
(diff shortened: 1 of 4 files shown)


VickieKarasic (approved):
thanks @kayiwa

PR #7093, opened 2026-05-05
[DPUL-C] Improve prometheus metrics

tpendragon:
We use these to keep track of some DPUL-C metrics right now. Overriding instance was making all the requests/second graphs break.


kayiwa (approved):
OK! thanks @tpendragon

PR #7101, opened 2026-05-08
[pulmirror] rewrite /mirror/nodejs/release/ to top-level versions

kayiwa:
dotsrc's nodejs mirror has fallen behind upstream — v24.3.0 and later are missing. We're switching the sync upstream to wget-from-nodejs.org, which uses nodejs.org's actual layout (versions at top level: /v24.3.0/, not /release/v24.3.0/).

Multiple roles consume nodejs_release_url, which points at https://pulmirror.princeton.edu/mirror/nodejs/release/<version>/. To keep that contract stable while the on-disk layout changes underneath, add an nginx rewrite that serves /mirror/nodejs/release/<ver>/<file> from /mirror/nodejs/<ver>/<file>, with a fallback to the legacy /release/ path during the transition.

The rewrite tries the new top-level path first and falls back to the existing release/ subtree, so both layouts work simultaneously. Once wget-from-nodejs.org has fully populated the top-level versions, the release/ subtree can be removed and the fallback dropped.

### DIFF
--- a/roles/pulmirror/defaults/main.yml
+++ b/roles/pulmirror/defaults/main.yml
@@ -45,7 +45,8 @@ pulmirror_dnsimple_ttl: 300
 
 # nodejs
 nodejs_mirror_root: /usr/local/www/nginx-dist/mirror/nodejs
-nodejs_mirror_source: rsync://mirrors.dotsrc.org/nodejs/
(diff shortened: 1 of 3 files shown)


VickieKarasic (approved):
Thanks @kayiwa!

PR #7104, opened 2026-05-11
[signoz] set up a Microk8s signoz

kayiwa:
setup our signoz environment using microk8s. There's less cognitive dissonance with upstream who expect a kubernetes cluster.


playbooks/signoz.yml
```diff
@@ -0,0 +1,28 @@
+---
```
tpendragon, line 1: Is the goal to spin up a k8s cluster and also deploy signoz on it, or is the goal to deploy signoz via creating a k8s cluster? By which I mean, should the boxes be named signoz-xxx? If not, then maybe deploying signoz and setting up k8s should be two separate playbooks?

roles/k8s_nfs_provisioner/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
+#SPDX-License-Identifier: MIT-0
+---
+# tasks file for roles/k8s_nfs
+- name: Add NFS external provisioner Helm repo
```
tpendragon, line 4: This is used as the storage point for signoz? Why not local disk? I expect this'll be pretty bouncy, but happy to be wrong.

roles/nginxplus/files/conf/http/dev/staging_signoz.conf
```diff
@@ -0,0 +1,47 @@
+upstream signoz-k8s {
+    least_conn;
+    zone signoz-k8s 64k;
+
+    server k8s-staging1.lib.princeton.edu:32080 max_fails=3 fail_timeout=10s;
```
tpendragon, line 5: Is 32080...signoz? Should we have an ingress gateway?

roles/signoz_k8s/README.md
```diff
@@ -0,0 +1,58 @@
+# SigNoz Kubernetes (microk8s)
```
tpendragon, line 1: Should this be a role or a playbook? Is there a helm role or something we can utilize?

playbooks/signoz.yml
kayiwa: it is admittedly a large PR. There is a k8s.yml playbook.

roles/nginxplus/files/conf/http/dev/staging_signoz.conf
kayiwa: yes it is signoz and I am not following your second question. sure?

roles/signoz_k8s/README.md
kayiwa: we are pulling https://charts.signoz.io (in the defaults)

kayiwa: It can likely be simplified but the PR is me needing to understand.

What does signoz need. (upstream says a K8s cluster) So we did that. We need storage (also did) that. If I'm not at the helm I am hoping someone else will see what is going on in a clearly delineated way and hold where all the things can be broken into understandable chunks.

roles/nginxplus/files/conf/http/dev/staging_signoz.conf
tpendragon: My usual understanding of k8s setups is that it dynamically allocates ports for services and uses an ingress gateway (basically an in-k8s load balancer) for forwarding requests to those dynamic ports, since it knows where they are.

I think in this case it sets a static port, my question was do we want that. If the goal here is to only run signoz here then it probably doesn't matter much, if the goal is to run a variety of helm things on these microk8s nodes and run a k8s cluster, then we might wanna think about it.

kayiwa: I see.

roles/k8s_nfs_provisioner/tasks/main.yml
kayiwa: we have no idea how much this will grow. The upstream recommendation is google fuse + bucket (but we don't have those level problems -yet?) So I erred on the side of a networked storage that local.

bess (approved):
Thank you, @kayiwa ! 
I used the system prototype and it looks great!

roles/nginxplus/files/conf/http/dev/staging_signoz.conf
kayiwa: I ticketed this here. https://github.com/pulibrary/princeton_ansible/issues/7113

PR #7105, opened 2026-05-11
Roll OS updates through safer reboots

### LINKED ISSUE #7103: Implement rolling updates with serial reboots and Solr wait time
Currently, our OS update playbook does not support rolling reboots. We need to add the ability to perform updates serially across hosts during maintenance windows, with a configurable wait period after rebooting. We will start with the Solr nodes.

kayiwa:
Update the OS maintenance playbook so package upgrades and reboots are handled as separate phases. Package updates can still run in small batches, but reboot handling now defaults to one host at a time to avoid restarting too many service nodes together.

Add explicit reboot tracking for Debian, RedHat, and FreeBSD hosts instead of rebooting inline during the package update tasks. Debian hosts continue to use /var/run/reboot-required, RedHat hosts are marked when yum or dnf changes packages, and FreeBSD hosts can be marked after pkg upgrades.

Add a configurable stabilization delay for sensitive inventory groups such as Solr. This gives Solr nodes time to rejoin and settle before the next host in the group is rebooted.

Excluded hosts such as ansible and recap servers are still not rebooted automatically. They continue to notify ops in Slack when a reboot is needed.

closes #7103

### DIFF
--- a/playbooks/utils/os_updates.yml
+++ b/playbooks/utils/os_updates.yml
@@ -179,18 +179,6 @@
         - os_update_reboot_required | default(false)
         - os_update_reboot_excluded
 
-    - name: Reboot if required
-      ansible.builtin.reboot:
-        msg: "Reboot initiated by Ansible because of {{ os_update_reboot_reason | default('OS package updates') }}."
-        connect_timeout: 5
-        reboot_timeout: 900
-        pre_reboot_delay: 0
-        post_reboot_delay: 15
-        test_command: whoami
-      when:
-        - os_update_reboot_required | default(false)
-        - not os_update_reboot_excluded
-
     - name: Determine matching stabilization delays
       ansible.builtin.set_fact:
         matching_reboot_stabilization_delays: >-
@@ -209,7 +197,7 @@
       ansible.builtin.set_fact:
         os_update_post_reboot_delay: >-
           {{
-            matching_reboot_stabilization_delays | default([]) | length > 0
+            (matching_reboot_stabilization_delays | default([]) | length > 0)
             | ternary(matching_reboot_stabilization_delays | max, 15)
           }}
       when:
@@ -224,20 +212,6 @@
         pre_reboot_delay: 0
         post_reboot_delay: "{{ os_update_post_reboot_delay | default(15) }}"
         test_command: whoami
-      when:
-        - os_update_reboot_required | default(false)
-        - not os_update_reboot_excluded
-
-    - name: Determine matching stabilization delays
-      ansible.builtin.set_fact:
-        matching_reboot_stabilization_delays: >-
-          {{
-            reboot_stabilization_delays
-            | dict2items
-            | selectattr('key', 'in', group_names)
-            | map(attribute='value')
-            | list
-          }}
       when:
         - os_update_reboot_required | default(false)
         - not os_update_reboot_excluded
--- a/playbooks/utils/os_updates.yml
+++ b/playbooks/utils/os_updates.yml
@@ -1,10 +1,10 @@
-
 ---
 # by default this playbook runs in the staging environment
 # to run in qa, pass '-e runtime_env=qa'
 # to run in production, pass '-e runtime_env=production'
 # remember that the loadbalancing machines are not included
 # and will need to be manually updated
+
 - name: update the Operating System packages
   hosts: "{{ runtime_env | default('staging') }}"
   remote_user: pulsys
@@ -14,10 +14,14 @@
     - ../group_vars/all/vars.yml
     - ../group_vars/all/vault.yml
 
+  vars:
+    freebsd_reboot_after_pkg_upgrade: true
+
   tasks:
     - name: Ubuntu | refresh keys
       ansible.builtin.command: /usr/bin/apt-key adv --refresh-keys --keyserver keyserver.ubuntu.com
       become: true
+      changed_when: false
       when: ansible_os_family == "Debian"
 
     - name: Ubuntu | Upgrade all packages
@@ -25,6 +29,7 @@
         update_cache: true
         upgrade: dist
         force_apt_get: true
+      register: apt_update_status
       when: ansible_os_family == "Debian"
 
     - name: RedHat | Upgrade all packages with fallback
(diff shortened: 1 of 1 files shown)


playbooks/utils/os_updates.yml
```diff
@@ -45,12 +50,36 @@
...
+        name: "*"
+        state: latest
+      register: freebsd_pkg_update_status
+      when: ansible_os_family == "FreeBSD"
+
+    - name: FreeBSD | remove packages that are no longer required
+      ansible.builtin.command:
+        cmd: pkg autoremove -y
+      register: freebsd_pkg_autoremove_status
+      changed_when: "'Deinstallation has been requested' in freebsd_pkg_autoremove_status.stdout"
```
acozine, line 73: Love this!

playbooks/utils/os_updates.yml
```diff
@@ -45,12 +50,36 @@
...
+      when: ansible_os_family == "FreeBSD"
+
+    - name: FreeBSD | clear local package cache
+      ansible.builtin.command:
+        cmd: pkg clean -ay
+      register: freebsd_pkg_clean_status
+      changed_when: "'The cleanup will free' in freebsd_pkg_clean_status.stdout"
```
acozine, line 80: This too.

playbooks/utils/os_updates.yml
```diff
@@ -72,51 +101,116 @@
...
+    - name: FreeBSD | Mark hosts that need reboot
+      ansible.builtin.set_fact:
+        os_update_reboot_required: true
+        os_update_reboot_reason: "pkg updates"
       when:
-        - ansible_os_family == "Debian"
-        - reboot_required_file.stat.exists == true
+        - ansible_os_family == "FreeBSD"
+        - freebsd_reboot_after_pkg_upgrade | bool
+        - freebsd_pkg_update_status.changed | default(false)
```
acozine, line 129: Do we need both of these conditions?

playbooks/utils/os_updates.yml
```diff
@@ -72,51 +101,116 @@
...
+            | map(attribute='value')
+            | list
+          }}
+      when:
+        - os_update_reboot_required | default(false)
+        - not os_update_reboot_excluded
+
+    - name: Wait after rebooting sensitive service hosts
+      ansible.builtin.pause:
+        seconds: "{{ matching_reboot_stabilization_delays | max }}"
```
acozine: Hmmm. I don't think this will work. The reboot task has already run, which means it has run against all hosts that match its conditions. The delay needs to be worked into that task somehow, I think.

acozine (changes requested):
This is a great start toward the rolling reboot process we want, but I don't think it's ready to merge (though I haven't tested it . . . ). My bet is that the `pause` task comes too late. I don't have an immediate sense of how to adjust this - maybe make the reboot task run as a loop? Can we leverage inventory groups to define the order (one from each group, then go back and start again)?

playbooks/utils/os_updates.yml
kayiwa: Yes, I think we need both.

`freebsd_pkg_update_status.changed` prevents unchanged FreeBSD hosts from
being marked for reboot. `freebsd_reboot_after_pkg_upgrade` is the policy
switch that lets us disable automatic FreeBSD reboots even when `pkg` changes
packages.

Since FreeBSD does not give us the same `/var/run/reboot-required` signal
that Ubuntu does, this keeps the behavior both conservative and configurable.

playbooks/utils/os_updates.yml
kayiwa: Good catch.

kayiwa:
> This is a great start toward the rolling reboot process we want, but I don't think it's ready to merge (though I haven't tested it . . . ). My bet is that the `pause` task comes too late. I don't have an immediate sense of how to adjust this - maybe make the reboot task run as a loop? Can we leverage inventory groups to define the order (one from each group, then go back and start again)?



> This is a great start toward the rolling reboot process we want, but I don't think it's ready to merge (though I haven't tested it . . . ). My bet is that the `pause` task comes too late. I don't have an immediate sense of how to adjust this - maybe make the reboot task run as a loop? Can we leverage inventory groups to define the order (one from each group, then go back and start again)?

I don't understand inventory groups well enough to use that approach.

playbooks/utils/os_updates.yml
```diff
@@ -72,51 +101,143 @@
...
+      ansible.builtin.reboot:
+        msg: "Reboot initiated by Ansible because of {{ os_update_reboot_reason | default('OS package updates') }}."
+        connect_timeout: 5
+        reboot_timeout: 900
+        pre_reboot_delay: 0
+        post_reboot_delay: 15
+        test_command: whoami
+      when:
+        - os_update_reboot_required | default(false)
+        - not os_update_reboot_excluded
```
acozine: Doesn't the actual reboot task need to be the very last one? So it knows what the delay should be? Also, what is the difference between this task and the one that starts in line 219?

acozine (commented):
Sorry you've been waiting for a review on this for a while. I'm still not entirely sure this will work. Maybe I just need to see it run against staging one morning.

kayiwa:
> Sorry you've been waiting for a review on this for a while. I'm still not entirely sure this will work. Maybe I just need to see it run against staging one morning.

FTR. The last two patch Wednesdays have run from this branch. So here's the "working code" https://ansible-tower.princeton.edu/#/jobs/playbook/48640/output

playbooks/utils/os_updates.yml
kayiwa: Yes, this is a good catch. The current version has duplicate reboot and
delay-calculation tasks.

The first reboot task still uses the hardcoded 15 second post_reboot_delay,
so it could run before the per-host Solr delay is calculated. The duplicate
matching-delay task at the end is also leftover and should be removed.

The intended order is:

1. determine whether the host is excluded from automatic reboot
2. notify ops for excluded hosts
3. calculate any matching service stabilization delay from group_names
4. set os_update_post_reboot_delay
5. run the single reboot task using that calculated delay

I removed the earlier hardcoded reboot task and the duplicate delay task so
there is only one reboot task, and it runs after the delay value has been
calculated.

Issue #7106, opened 2026-05-11
[libstatic] playbook fails on deploy user permissions task


VickieKarasic:
### Expected behavior
`static.yml` playbook should succeed on all tasks when we make any changes to the repo 

### Actual behavior
Playbook is failing on the task to Grant permissions to deploy user 

### Error messages
```TASK [libstatic : Libstatic | Grant permissions for deploy user] ***************
[ERROR]: Error rendering template: unsupported operand type(s) for +: 'NoneType' and '_AnsibleLazyTemplateList'
Origin: /Users/karasicv/Desktop/Git_Repos/princeton_ansible/roles/libstatic/tasks/main.yml:25:15

23     recurse: true
24     mode: "0755"
25   with_items: '{{ named_sites + sites }}'
                 ^ column 15

fatal: [static-staging1.lib.princeton.edu]: FAILED! => {"msg": "Error rendering template: unsupported operand type(s) for +: 'NoneType' and '_AnsibleLazyTemplateList'"}

PLAY RECAP *********************************************************************
static-staging1.lib.princeton.edu : ok=62   changed=0    unreachable=0    failed=1    skipped=22   rescued=0    ignored=0
```
### Steps to replicate
How would someone else make this error happen?

Run the `static.yml` playbook on staging

acozine:
Closed by #7108

PR #7108, opened 2026-05-11
[libstatic] remove old cicognara references and named_sites vars

VickieKarasic:
- removes old cicognara references in group_vars for libstatic prod since that site is now hosted on Nomad

- removes references to the named_sites vars, since these no longer exist

related to #7106

### DIFF
--- a/roles/libstatic/tasks/main.yml
+++ b/roles/libstatic/tasks/main.yml
@@ -13,14 +13,6 @@
   tags:
     - install_keys
 
-- name: debug named_sites
-  ansible.builtin.debug: 
-    var: named_sites
-
-- name: debug sites
-  ansible.builtin.debug: 
-    var: sites
-
 - name: Libstatic | Grant permissions for deploy user
   ansible.builtin.file:
     path: "{{ item.repo_root }}"
@@ -74,17 +66,6 @@
   tags:
     - install_sites
 
-# - name: Libstatic | Add all the named sites
-#   ansible.builtin.template:
-#     src: "apache-named-sites.conf.j2"
-#     dest: "/etc/apache2/sites-available/{{ item.server_name }}.conf"
-#     mode: "0644"
-#   notify: restart apache
-#   become: true
-#   with_items: '{{ named_sites }}'
-#   tags:
-#     - install_sites
-
 - name: Libstatic | Add aliases and directories for the libstatic site
   ansible.builtin.template:
     src: "apache-static-sites.conf.j2"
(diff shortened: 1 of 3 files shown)


roles/libstatic/tasks/main.yml
```diff
@@ -13,6 +13,14 @@
...
     - install_keys
 
+- name: debug named_sites
+  ansible.builtin.debug: 
+    var: named_sites
+
+- name: debug sites
+  ansible.builtin.debug: 
+    var: sites
+
```
acozine: I'd remove these tasks now that they have given us the information we needed.

roles/libstatic/tasks/main.yml
```diff
@@ -61,21 +69,21 @@
...
+# - name: Libstatic | Add all the named sites
+#   ansible.builtin.template:
+#     src: "apache-named-sites.conf.j2"
+#     dest: "/etc/apache2/sites-available/{{ item.server_name }}.conf"
+#     mode: "0644"
+#   notify: restart apache
+#   become: true
+#   with_items: '{{ named_sites }}'
+#   tags:
+#     - install_sites
```
acozine: I'd go ahead and delete these.

roles/libstatic/tasks/main.yml
```diff
@@ -87,14 +95,14 @@
...
-  tags:
-    - install_sites
+# - name: Libstatic | enable apache2 sites
+#   ansible.builtin.command: 'a2ensite {{ item.site_name }}'
+#   become: true
+#   with_items: '{{ named_sites }}'
+#   register: apache2_enabled
+#   changed_when: '"Enabling site" in apache2_enabled.stdout'
+#   tags:
+#     - install_sites
```
acozine: Same thing here. The reason we originally commented them out was so we could easily get them back if it turned out we needed them.

acozine (commented):
Looks great - nice work.  I added a couple of suggestions for tidying up a bit.

roles/libstatic/tasks/main.yml
VickieKarasic: Done, thanks!

roles/libstatic/tasks/main.yml
VickieKarasic: Good catch - forgot we had commented that out

roles/libstatic/tasks/main.yml
VickieKarasic: Ditto for this one - thanks!

kevinreiss (approved):

PR #7112, opened 2026-05-11
[Nomad] Push logs to Loki & Signoz

tpendragon:
Use Alloy configuration to also ship OTLP logs to Signoz using the same infrastructure we already have.


kayiwa (approved):
🎉 Thanks for these @tpendragon

PR #7122, opened 2026-05-13
[gcploadbalancer] allow google scholar IPs

kayiwa:
Add the Google Scholar crawler CIDR ranges to the OpenResty ALTCHA bypass network list for the production GCP load balancers.

Google Scholar provided these ranges so their crawler can reach Dataspace content without receiving the ALTCHA verification challenge. This keeps the existing trusted user-agent bypass in place, but allows the Scholar crawler by source IP as well, which is less fragile than relying only on user-agent matching.

No behavior changes are expected for untrusted clients because requests outside the configured bypass networks and user-agent list still pass through the ALTCHA gate.


kayiwa:
I have run the playbook from this branch

kelynch (approved):

PR #7125, opened 2026-05-13
Updates postgresql role to support v16 on RHEL for Tower 2.6

acozine:
Before we can move Ansible Tower to Entra-ID we need to upgrade it. Version 2.6 of RHAAP (which we still call Tower locally) requires a separate database VM. It also supports postgreSQL 15, 16, or 17. We've chosen version 16, which is easily available for RHEL 9 today. 

This PR will eventually update the postgresql role to support postgreSQL 16 on RHEL. Right now it's a bit of a mess. Do not merge. Untested against any Ubuntu machines. Here be dragons.


roles/postgresql/tasks/main.yml
```diff
@@ -1,114 +1,20 @@
...
+   name: "{{ postgres_service_name }}"
+   state: started
+   enabled: true
+  # when:
+  #  - running_on_server
 
-- name: PostgreSQL | add a logrotation
+- name: PostgreSQL | add logrotate rules
   ansible.builtin.template:
     src: "../../common/templates/logrotate_rules.j2"
```
acozine, line 17: Check that this path is correct for both Ubuntu and RHEL

roles/postgresql/tasks/main.yml
```diff
@@ -120,87 +26,19 @@
...
-  notify: restart local postgres
-
 - name: PostgreSQL | add replication postgresql configuration
   ansible.builtin.template:
     src: replication.conf.j2
     dest: "/etc/postgresql/{{ postgres_version }}/main/conf.d/replication.conf"
     mode: "0644"
   when:
-    - running_on_server
+    # - running_on_server
```
acozine: We might not be able to run this in CI, because we don't have anywhere to write replication, so we probably need to keep the `running_on_server`

roles/postgresql/tasks/main.yml
acozine: Confirmed that `/etc/logrotate.d/` is the correct path for log rotation definitions on both OS flavors. 

At the same time we discovered that the logrotate status file has a different name per OS flavor. On Ubuntu it's `/var/lib/logrotate/status` and on RHEL/Rocky it's `/var/lib/logrotate/logrotate.status`. See #7172 for that work.

acozine:
Keeping this as is for history, but closing it in favor of #7184.

Issue #6479, opened 2025-09-02
[Monitoring] Check to see if our URLs are up.


tpendragon:
### User story

We should check all of our public URLs to make sure they're up. Right now some sites use Honeybadger for this, some use Datadog - if we're moving to CheckMK, then CheckMK should handle it.

Made because all of our Nomad sites went down when the sandbox for Nomad got deployed over it and we had no idea all our staging sites hosted there went down.

### Acceptance criteria

- [ ] When a site's not healthy (staging or prod) for a certain amount of time we get an alert.

hackartisan:
Configured checkmk to notify to `#dls-alerts` when DLS staging or prod sites are down.

hackartisan:
Reopening; digital collections had 43 minutes of downtime today and we didn't get an alert in dls-alerts

hackartisan:
Update: We reverted the conditions of the alert. Now it's back to being more sensitive than we would prefer.

hackartisan:
And I also turned off those notifications again because now they are noisy

hackartisan:
I've created a "delay service notification" rule so that the uptime checks don't trigger until two minutes after they're detected. here's the rule on staging: https://pulmonitor.princeton.edu/staging/check_mk/index.py?start_url=%2Fstaging%2Fcheck_mk%2Fwato.py%3Ffolder%3D%26host%3D%26item%3DTm9uZQ%253D%253D%26mode%3Dedit_ruleset%26service%3DTm9uZQ%253D%253D%26varname%3Dextra_service_conf%253Afirst_notification_delay

hackartisan:
Okay. The delay service notification rule made the UI show a service as up during that delay. What we really want is for the UI to show the service as down, but for the alert to be delayed. After 3 solid tries at making this work, I'm thinking maybe we consider trying again on signoz once we're clear we're keeping it. https://signoz.io/docs/monitor-http-endpoints/ works similar to the way we used to do it with datadog:
https://github.com/pulibrary/princeton_ansible/blob/f7a45a30ae7db21f925a75881deb2ebbe18115dd/group_vars/dpul/production.yml#L131-L155

I set up all the prod monitoring in honeybadger for now, with alerts going to #dls-alerts. And I went ahead and made a little [status page](https://status.hbuptime.com/BKASRM) just 'cuz it was easy

PR #7142, opened 2026-05-19
[DLS] Ensure all our apps are passing the status code from upstream

hackartisan:
follows the pattern set in #2227

work toward #6479


kayiwa (approved):
OK! Thanks @hackartisan

Issue #6478, opened 2025-09-02
[Monitoring] CheckMK should raise an alert if Nginx+ is pointed at a node with zero stable upstreams.


tpendragon:
### User story

One common thing we check when a site goes down is if nginx+ notices they're down. We should get an alert if it finds that a whole service is down.

### Acceptance criteria

- [ ] When nginx+ has no valid upstreams we should get an alert.

acozine:
I did a little nosing around today. Here's what I found:

To monitor NGINX Plus from CheckMK, you typically use the nginx_status plugin. While standard NGINX uses the stub_status module, NGINX Plus offers a more advanced API for monitoring.
Enable the Status Module:For basic metrics, ensure the NGINX Stub Status module is enabled.
For NGINX Plus specific metrics (like upstream health and advanced traffic stats), ensure the NGINX Plus API is active.
Deploy the Plugin:
Manual: Download the nginx_status script from your Checkmk server and place it in the /usr/lib/check_mk_agent/plugins/ directory on your NGINX host.Automatic (Enterprise)
Use the Bake Agent feature. Search for the "NGINX webservers (Linux)" rule, configure the connection details (e.g., port and protocol), and redeploy the agent.
Discovery: Once the plugin is in place and executable, run a service discovery on the host in Checkmk to see the new metrics.

aruiz1789:
We configured the Checkmk plugin for Nginx status monitoring.  This plugin uses the "stub status module" which does not provide detailed monitoring.  To monitor the health of upstream servers a local check script was placed on the Nginx Servers.

This provides an overview of the status of each server upstream. 

<img width="1224" height="488" alt="Image" src="https://github.com/user-attachments/assets/cdb57a11-c2a6-4542-9ee2-2f3832fb1dd0" />

PR #7147, opened 2026-05-20
[nginxplus] monitor upstream health in CheckMK

kayiwa:
Add a Checkmk local check for NGINX Plus load balancers that queries the local upstream API and reports service health per upstream. The check raises a critical alert when an upstream has no usable primary peers, which catches the operational case where NGINX Plus has marked an entire backend service unavailable.

related #6478


VickieKarasic (approved):
Thanks @kayiwa

PR #7159, opened 2026-05-26
[otel_collector] install versioned otelcol upgrades

kayiwa:
The otel_collector role did not honor changes to otel_version after the collector had already been installed. The unarchive task used the installed binary path as its creates guard, so Ansible skipped extraction whenever /opt/otelcol/otelcol-contrib already existed.

This changes the collector install flow to use versioned release directories under /opt/otelcol/releases/<version>. The role downloads the configured amd64 release archive, extracts it into the version-specific directory, and then copies that selected binary to /opt/otelcol/otelcol-contrib.

When otel_version changes, Ansible extracts the new release, replaces the active binary when the checksum differs, and restarts otelcol-contrib.


VickieKarasic (approved):
Ok - thanks @kayiwa

Issue #6234, opened 2025-06-12
[Log Rotation]: Define which VMs should send logs to the centralized logging service.


aruiz1789:
We are in the process of designing and building a real-time log management, analysis and visualization platform (Pulemetry).  That will help with collection, processing, indexing and analyzing large volumes of data from various sources.  

Identifying and defining the source of logs should be the initial result of this issue.

PR #7161, opened 2026-05-26
[gitlab] Add GitLab telemetry collection

kayiwa:
Configure the OpenTelemetry collector to scrape the local GitLab
Prometheus exporters and ship GitLab component logs to SigNoz.

This gives the GitLab staging host host, PostgreSQL, Redis, Puma,
Sidekiq, and GitLab application metrics without exposing exporter
ports outside the VM. It also adds log collection for the main
Omnibus GitLab services under /var/log/gitlab.

The configuration keeps the existing OTLP receivers in place so
future application traces can be accepted if GitLab or related
services emit them.

related to #6234


acozine:
Looks great, thanks @kayiwa!

PR #7162, opened 2026-05-27
[orangelight] disable ufw for catalog prod servers

### LINKED ISSUE #7160: Catalog isn't behind the Traefik bot wall.
### Expected behavior

https://catalog.princeton.edu/challenge redirects to a Cloudflare turnstile.

### Actual behavior

There's no turnstile in place.

### Impact of this bug

The next bot wave will cripple the catalog.

I think what happened was Traefik does health checks on apps, but can't access the newly built apps because we have UFW enabled on newly built boxes., but aren't allowing port 80 from the private network.

When I'm on `nomad-client-prod1.lib.princeton.edu` and `curl http://catalog1.princeton.edu` it hangs forever.

When I run `sudo ufw status` on `catalog1` I notice that only specific private IPs are allowed access to port 80, and that doesn't include the Nomad infrastructure.

My proposal is that we open port 80 to the private network.

kayiwa:
The catalog production webservers should not manage host firewall state with ufw during the orangelight deployment.

Add a post task to the orangelight playbook that disables ufw only when running against the production runtime environment and only for members of the orangelight_production_webservers group.

This keeps the change scoped to the catalog production web hosts and does not affect staging, QA, or indexer hosts.

Closes #7160 when we can successfully replace one of the catalog webservers and the firewall is disabled

### DIFF
--- a/playbooks/orangelight.yml
+++ b/playbooks/orangelight.yml
@@ -31,6 +31,14 @@
       when: runtime_env | default('staging') == "production"
 
   post_tasks:
+    - name: Production webservers | Disable ufw
+      ansible.builtin.command: ufw disable
+      when:
+        - runtime_env | default('staging') == "production"
+        - "'orangelight_production_webservers' in group_names"
+      register: ufw_disable_result
+      changed_when: "'Firewall stopped and disabled on system startup' in ufw_disable_result.stdout"
+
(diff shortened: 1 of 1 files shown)


VickieKarasic (approved):
This worked @kayiwa - thank you!

PR #7170, opened 2026-05-28
[Nomad] Disable DNSMasq caching.

### LINKED ISSUE #7168: Updating Machine IPs shouldn't break Nomad
### Expected behavior

Updating a machine's IP in DNS should propagate to Nomad. We had an [incident](https://docs.google.com/document/d/1nx0BBqVDiyn4HNA50sgUUKmt3kXMfScXOm5-fOpAH6Y/edit?tab=t.0) where we switched Postgres to the private network, and everything on Nomad didn't catch it because DNSMasq caches DNS entries.

### Actual behavior

DNSMasq kept the old IPs for several hours.

### Steps to replicate

Change a machine's IP, nslookup from any Nomad machine.

### Impact of this bug

Machines in the Nomad infrastructure may break and be difficult to diagnose.

Some possible options:

1. Stop caching in DNSMasq: https://netbeez.net/blog/linux-dns-caching-dnsmasq/ says we can do "cache_size=0"
2. Set max-cache-ttl option or flag (referenced here: https://serverfault.com/a/706517) to a much smaller number (minutes?)
3. Somehow know that we should restart dnsmasq when we update DNS..?

tpendragon:
We use it for routing, so don't really want to take over caching and it contributed to an outage. This might end up with a bunch more DNS requests to OIT infrastructure, we'll have to see.

The TTL of the DNS records is usually 12 hours.

Closes #7168

This has been deployed.

### DIFF
--- a/group_vars/nomad_sandbox.yml
+++ b/group_vars/nomad_sandbox.yml
@@ -1,5 +1,5 @@
 ---
-nomad_http_address: nomad-sandbox.lib.princeton.edu
+nomad_http_address: 'https://nomad-sandbox.lib.princeton.edu'
 nomad_server_group: "nomad_servers_sandbox"
 consul_version: '1.20.3'
 consul_gossip_encryption_key: '{{ vault_consul_gossip_encryption_key }}'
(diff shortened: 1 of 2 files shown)


kayiwa (approved):
🎉  thanks @tpendragon

PR #7173, opened 2026-06-01
[step ca] authentication via Entra ID

### LINKED ISSUE #7327: Restricted SSH Access to Proxmox Servers for Operations Team
The Operations Team requires secure, limited SSH access to our Proxmox VE cluster hosts to perform routine VM/container management tasks (start/stop/restart/status, list guests, view logs) without needing full root or administrative shell access. This ticket covers the implementation of a restricted access solution that enforces least privilege, logs all actions, and integrates with our existing authentication/authorization framework.

Current state: Only Ops team should has full SSH access using the default root used at initial setup. 

Goal: Provide an auditable, and restricted SSH entry point for Ops team members.

kayiwa:
related to #7025 

closes #7327 

create shortlived certificates for SSH auth. https://smallstep.com/docs/step-ca/provisioners/


group_vars/proxmox/sandbox.yml
```diff
@@ -0,0 +1,38 @@
...
+
+step_ca_oidc_enabled: true
+
+step_ca_oidc_admins:
+  - ac2754@princeton.edu
+  - ar1789@princeton.edu
+  - gpmenos@princeton.edu
+  - jkazmier@princeton.edu
+  - kayiwa@princeton.edu
+  - fkayiwa@princeton.edu
```
acozine, line 18: Did you mean to give yourself netID and alias, but give the rest of us netID only?

group_vars/proxmox/common.yml
```diff
@@ -0,0 +1,38 @@
+---
+step_ca_ssh_principals:
```
acozine, line 2: This is duplicated in sandbox.yml and here (and presumably would be in production if we had a prod environment).  I'd put any vars that apply across all environments into a common.yml, to prevent config drift and reduce maintenance overhead.

group_vars/proxmox/common.yml
```diff
@@ -0,0 +1,38 @@
...
+  - ac2754@princeton.edu
+  - ar1789@princeton.edu
+  - gpmenos@princeton.edu
+  - jkazmier@princeton.edu
+  - kayiwa@princeton.edu
+  - fkayiwa@princeton.edu
+  - vk4273@princeton.edu
+
+
+step_ca_oidc_listen_address: "localhost:10000"
```
acozine, line 22: Same here. I'd reserve the env-specific files for values that are specific to the one environment.

playbooks/ssh_crt_auth.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Configure Cert Auth
```
acozine: Can you add a comment to this describing what it does?

playbooks/step_ca.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Configure Step_CA
```
acozine: Same here - six months from now I'm going to want a comment here to make me confident about what this playbook is doing. I doubt we will run it often enough for the meaning to be loaded in memory.

roles/ssh_ca_trust/README.md
```diff
@@ -0,0 +1,60 @@
+# ssh_ca_trust
+
+Configures an SSH host to accept certificates signed by the `step_ca` SSH
```
acozine, line 3: In this context an "SSH host" is any VM or physical machine that we can SSH to?

roles/ssh_ca_trust/README.md
```diff
@@ -0,0 +1,60 @@
...
+that trusts a static public key. No secrets are involved, so this role is safe
+to apply broadly and is a good candidate to fold into your `common` role once
+the SSH-CA pilot graduates.
+
+## What it does
+
+1. Installs `openssh-server`.
+2. Writes the User CA public key to `/etc/ssh/trusted_user_ca_keys.pem`
+   (optionally asserting `ssh_ca_user_ca_fingerprint`).
+3. Creates local users and their `/etc/ssh/auth_principals/%u` files.
```
acozine, line 14: On any machine using this approach, each person SSH'ing in will have their own *nix user instead of using the shared `pulsys` user, correct?

roles/ssh_ca_trust/README.md
```diff
@@ -0,0 +1,60 @@
...
+   bad config never triggers a restart).
+6. Warns if `AllowUsers`/`AllowGroups` could silently block cert logins — this
+   was the final blocker on both pilot hosts.
+
+## Variables
+
+```yaml
+# Shared data — define once in group_vars/all (printed by the step_ca role):
+ssh_ca_user_ca_public_key: "ssh-ed25519 AAAA… ssh_user_ca_key"
+ssh_ca_user_ca_fingerprint: "SHA256:uFSXrhpueJZ1NT5j36Hf6BeAsV0Ssx+ZOR74oRymIqM"
```
acozine, line 26: Are these sample values or actual values? The first one looks like an example, but the second one looks real.

roles/ssh_ca_trust/README.md
```diff
@@ -0,0 +1,60 @@
...
+    principals:
+      - pulsys
+    # groups defaults to the OS admin group (sudo/wheel) + pulsys
+    # create_local_user defaults to true
+```
+
+With the `step_ca` `pulsys` certificate template, every Entra-authenticated
+user receives a cert whose only principal is `pulsys` (the cert `KeyID` records
+who they actually are). So a single `pulsys` user entry is usually all you
+need; the audit trail lives in the cert KeyID and sshd logs.
```
acozine, line 40: Okay, now I'm confused.

roles/step_ca/README.md
```diff
@@ -0,0 +1,96 @@
+# step_ca
+
+Builds and runs the Smallstep [`step-ca`](https://smallstep.com/docs/step-ca)
+certificate authority on `step-ca.lib.princeton.edu`, backed by Microsoft
+Entra ID OIDC, issuing short-lived SSH user certificates. Targets Ubuntu.
```
acozine, line 5: Very clear!

I might add a sentence describing the goal - the "so that we can" part of the conversation. Something like: "Using Entra ID authentication and short-lived SSH user certs instead of permanent SSH keys ensures that only currently authorized users can SSH to our infrastructure." That might not be correct, but that's the general direction I had in mind.

roles/step_ca/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+This role has a tight coupling with the [ssh_ca_trust](../ssh_ca_trust) role, which configures the SSH hosts that
```
acozine: I find "SSH hosts" confusing. Do we have any hosts that are NOT SSH hosts?

acozine (commented):
This is a fantastic solution to an immediate problem, and one we can roll out to more infrastructure in future.

I haven't read the entire PR, but here's my first round of questions and comments.

group_vars/proxmox/sandbox.yml
kayiwa: fkayiwa failed

roles/ssh_ca_trust/README.md
kayiwa: Yes, that is the intended direction. The role can support either model, but for this Proxmox/SSH certificate work we are moving toward one local Unix account per person. The cert principal is the Entra-backed identity, for example kayiwa@princeton.edu, and sshd maps that to the local Unix user through /etc/ssh/auth_principals/%u.

The shared pulsys approach was the pilot/default example, but I’ll clarify in the README that the long-term model is per-person Unix users rather than everyone SSH’ing as pulsys.

roles/ssh_ca_trust/README.md
kayiwa: These two playbooks use the same role to do two different things. 

Until proxmox we didn't have an individual user case. The role can do both. 

```sh

❯ cat playbooks/ssh_crt_auth.yml
---
# Configure hosts to trust the Library SSH User Certificate Authority.
#
# This playbook installs the CA public key into sshd, creates local Unix users,
# and writes each user's AuthorizedPrincipalsFile entry. After this runs,
# people can SSH to these hosts with short-lived certificates issued by
# step-ca/Entra instead of relying on long-lived personal SSH keys.

- name: Configure SSH certificate authentication
  hosts: ssh_crt_auth
  remote_user: pulsys
  become: true
  vars_files:
    - ../group_vars/step_ca/main.yml
    - ../group_vars/all/vars.yml
  roles:
    - role: ssh_ca_trust

❯ cat playbooks/proxmox.yml
---
- name: proxmox add host to CA
  hosts: step_ca
  remote_user: pulsys
  become: true
  vars_files:
    - ../group_vars/proxmox/{{ runtime_env | default('staging') }}.yml
    - ../group_vars/all/vars.yml
    - ../group_vars/step_ca/vault.yml

  roles:
    - role: step_ca

- name: proxmox add users to host
  hosts: proxmox_servers_{{ runtime_env | default('staging') }}
  remote_user: bootstrap
  become: true
  vars_files:
    - ../group_vars/proxmox/{{ runtime_env | default('staging') }}.yml
    - ../group_vars/all/vars.yml

  pre_tasks:
    - name: Install sudo
      ansible.builtin.apt:
        name: sudo
        state: present
        update_cache: true

    - name: Create Ops Group
      ansible.builtin.group:
        name: opsys
        state: present

    - name: Grant sudo privileges to opsys group
      ansible.builtin.copy:
        content: |
          %opsys ALL=(ALL:ALL) NOPASSWD:ALL
        dest: /etc/sudoers.d/opsys
        owner: root
        group: root
        mode: "0440"
        validate: /usr/sbin/visudo -cf %s
  roles:
    - role: ssh_ca_trust
```

roles/ssh_ca_trust/README.md
kayiwa: https://pypi.org/project/Faker/ (did it's job then.)

aruiz1789 (approved):

Issue #7030, opened 2026-04-13
Add Support for Yarn Berry (Yarn 2+) in Node.js Role


kayiwa:
Currently, our Nodejs role supports the installation and management of Node.js and Yarn Classic (Yarn 1.x). However, Yarn Berry (Yarn 2+) [introduces several improvements](https://medium.com/@bloodturtle/yarn-vs-yarn-berry-the-complete-comparison-guide-every-frontend-developer-needs-812c4e0db736) including, improved performance. To keep our role up-to-date with modern JavaScript development practices, we should add support for Yarn Berry.

## Additional Context:
Yarn Berry introduces breaking changes compared to Yarn Classic, so careful consideration must be given to ensure backward compatibility and smooth migration paths for users.

kayiwa:
If I am reading [this correctly](https://yarnpkg.com/corepack). yarn configurations needs to happen on a per-project basic (so at the capistrano level) 

```text
Corepack lets you define “which package manager version you want to use on a per-project basis,” and Yarn’s install page starts with installing or enabling Corepack rather than installing Yarn Berry from an OS package repository
```

Yarn’s own version-management docs also point to the repository as the control point. More [on that here](https://yarnpkg.com/cli/set/version):

```text
The `yarn set version` command says it sets the Yarn release to be used by Corepack, and by default it writes the packageManager field in the project root. When needed, it also stores Yarn inside the project and references it from .yarnrc.yml. The docs explicitly frame this as a way to ensure everyone on a team uses the same Yarn version in the same project, including deployment servers
```

acozine:
Discussed today in Ansible Open House. We will try to get TigerData-app to a stable yarn version with passing CI.

- select a specific version (e.g. 4.15)
- set this in all the places
  - Prancible
    - in group_vars for the project
    - in the playbook 
  - in the application repo
    - in DevBox.json
    - package.json

PR #7174, opened 2026-06-02
[nodejs] refactor nodejs role to install Yarn Berry via Corepack

kayiwa:
Add opt-in Yarn Berry support to nodejs role

Keep the existing Yarn Classic installation path through the upstream Yarn APT
repository for backward compatibility. Add a new nodejs_yarn_berry_enabled flag
that switches Yarn management to Corepack and installs the pinned Yarn Berry
version 4.15.0.

Update role defaults, tasks, and Molecule verification so existing consumers
continue to receive Yarn Classic unless they explicitly opt into Yarn Berry.

related to #7030

### DIFF
--- a/roles/nodejs/defaults/main.yml
+++ b/roles/nodejs/defaults/main.yml
@@ -1,13 +1,16 @@
 ---
 # Enable or disable support for Yarn upstream APT repository.
 nodejs__yarn_upstream: true
+
 # URL for the Yarn GPG key
 nodejs__yarn_key_url: "https://dl.yarnpkg.com/debian/pubkey.gpg"
-# Path for the Yarn GPG keyring (Modern Ubuntu/Debian standard)
+
+# Path for the Yarn GPG keyring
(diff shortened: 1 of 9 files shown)


roles/nodejs/defaults/main.yml
```diff
@@ -17,3 +20,11 @@ nodejs_release_url: "https://pulmirror.princeton.edu/mirror/nodejs/release/"
 
 # use yarn rather than npm as package manager
 use_yarn: true
+
+# Use Yarn Classic from the Yarn APT repository by default.
+# Set this to true to install Yarn Berry with Corepack instead.
+nodejs_yarn_berry_enabled: false
+
+# Pinned Yarn Berry version
+nodejs_yarn_berry_version: "4.15.0"
```
carolyncole, line 29: @kayiwa is this going to upgrade everyone by default?

kayiwa: line 26 makes sure that it doesn't.

bess (approved):
Thank you! 🤩

PR #7176, opened 2026-06-03
[nginxplus] make upstream check backup-aware

kayiwa:
The existing local check went CRITICAL whenever all primary peers in an upstream were unhealthy, regardless of whether backup peers were up and serving traffic. This produced false alarms for any upstream using the traefik-wall pattern on pdc-discovery-prod when an active health_check probes through highchallenge, the wall returns a challenge page instead of the app's /health.json, the primary is marked unhealthy, and nginx correctly falls over to the direct VM backups. The site stays up but the check still pages red.

Split the per-upstream tally into primary and backup groups. WARNING now covers the "all primaries down, backups serving" case with a message naming both sides ("serving from N/M backup peers"). CRITICAL is reserved for upstreams with no usable peers at all.

Dropped the "NGINX Plus upstream summary" service. Per- upstream services give checkmk everything it needs to roll up host status, and the summary added noise without adding signal.

### DIFF
--- a/roles/nginxplus/files/check_nginx_plus_upstreams.py
+++ b/roles/nginxplus/files/check_nginx_plus_upstreams.py
@@ -57,11 +57,6 @@ def main():
         emit(3, "NGINX Plus upstreams", "-", "Unexpected API response format")
         return
 
-    total_upstreams = 0
-    critical_upstreams = 0
-    warning_upstreams = 0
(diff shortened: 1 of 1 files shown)


kayiwa:
Successfully run on prod load balancers.

bess (approved):

Issue #7181, opened 2026-06-03
[PostgreSQL] Ubuntu databases should have UTF-8 encoding


acozine:
### Expected behavior
The postgresql role should create a working postgres database server on Ubuntu or on RHEL.

### Actual behavior
On Ubuntu (for sure on Noble/24, probably on others also), the role fails when we try to create the database, becasue the default encoding is set to LATIN1 instead of UTF8.

### Error messages

```
[ERROR]: Task failed: Module failed: Changing database encoding is not supported. Current encoding: LATIN1
Origin: /Users/alicia/projects/princeton_ansible/roles/postgresql/tasks/create_db.yml:23:3

21   run_once: true
22
23 - name: PostgreSQL | create postgresql server database
     ^ column 3

failed: [172.20.80.212] (item={'name': 'postgres'}) => {"ansible_loop_var": "item", "changed": false, "item": {"name": "postgres"}, "msg": "Changing database encoding is not supported. Current encoding: LATIN1"}
```

### Steps to replicate
Run the postgresql.yml playbook against a new Ubuntu VM.

### Impact of this bug
We cannot recreate our database environment on demand.

### Relevant links and code snippets, if applicable
Manual fix: log into postgres and from the psql command line run:

```
UPDATE pg_database SET datistemplate = FALSE WHERE datname = 'template1';
DROP DATABASE template1;
CREATE DATABASE template1 WITH TEMPLATE = template0 ENCODING='UNICODE' LC_COLLATE='en_US.UTF8' LC_CTYPE='en_US.UTF8';
UPDATE pg_database SET datistemplate = TRUE WHERE datname = 'template1';
```
Repeat these steps, switching `template0` to `template1` and vice versa. 
Then drop the main postgres database from the main command line with `dropdb postgres`, recreate it with `createdb postgres`. Once the new main database exists, go back to the postgres command line and modify both templates to reject connections:

```
UPDATE pg_database SET datallowconn = FALSE WHERE datname = 'template0';
UPDATE pg_database SET datallowconn = FALSE WHERE datname = 'template1';
```

VickieKarasic:
We found that we define `ctype_parts` here but we don't implement it: https://github.com/pulibrary/princeton_ansible/blob/ea11f0033a20c123c74abe3ed5acf4dc3a0a0466/roles/postgresql/defaults/main.yml#L72

kayiwa:
[...]

Issue #7136, opened 2026-05-17
Postgresql warm standby not enabled


tpendragon:
I noticed our postgres warm standbys weren't working. I think it's because `postgresql_cluster_member` isn't set.

acozine:
For the configuration task, we use the `postgresql_cluster_member` var:

```
- name: PostgreSQL | add replication postgresql configuration
  ansible.builtin.template:
    src: replication.conf.j2
    dest: "/etc/postgresql/{{ postgres_version }}/main/conf.d/replication.conf"
    mode: "0644"
  when:
    - running_on_server
    - ansible_os_family == "Debian"
    - postgresql_cluster_member == true
  notify: restart postgresql
```

For the task file includes, we use `postgresql_cluster` plus conditions on the inventory hostnames. For example:

```
- name: PostgreSQL | configure warm standy follower
  ansible.builtin.include_tasks: warm_standby_follower.yml
  when:
    - postgresql_cluster != false
    - inventory_hostname in postgresql_cluster.followers
```

We define `postgresql_cluster` in group_vars [here](https://github.com/pulibrary/princeton_ansible/blob/df1a96e71060aea4b53ce801dfca13ba78ea6420/group_vars/postgres_staging_cluster.yml#L2-L5) and elsewhere in the same directory.  But we set `postgresql_cluster_member` to false by default and don't set it anywhere else. 

Can we use the `postgresql_cluster` var instead of `postgresql_cluster_member`?

PR #7184, opened 2026-06-03
Updates postgresql role to support psql >=15 on RHEL

acozine:
We need to upgrade Ansible Tower to RHAAP version 2.6, which requires a separate, dedicated database VM. It also supports postgreSQL 15, 16, or 17. We've chosen version 16, which is easily available for RHEL 9 today.

This PR updates the postgresql role by:
- splitting the tasks into RHEL and Debian tasks files
- splitting the vars into RHEL and Debian vars files
- updating tasks/main.yml to load both tasks and vars based on the OS flavor ansible fact
- using OS-agnostic var names for flexibility
- updating paths, etc., on RHEL to support PostgreSQL >=15

It has been tested on both Ubuntu and RHEL for building standalone postgres servers. It doesn't change the clustered build tasks, but we will look at those again as part of working on #7181 and #7136.


VickieKarasic (approved):
This looks good @acozine and glad it's working!

PR #7191, opened 2026-06-04
[Postgresql] Divide client tasks from server tasks

### LINKED ISSUE #7190: Dynamic parent include path error when running approvals playbook
### Expected behavior

The approvals playbook can run

### Actual behavior

It [hits an error in the postgresql role](https://ansible-tower.princeton.edu/#/jobs/playbook/49775/output):

```
ERROR! Error when evaluating variable in dynamic parent include path: install_{{ ansible_os_family | lower }}.yml. When using static imports, the parent dynamic include cannot utilize host facts or variables from inventory
```

### Steps to replicate
Run the approvals playbook (and probably other playbooks)

### Impact of this bug

Can't run the approvals playbook

* #7184 may be related

acozine:
Closes #7190.

In #7184 we added conditions to the postgresql role that looked at the OS flavor and used that to load variables and tasks. This worked well for server builds, and supported installing postgreSQL 16 on RHEL for our upgraded Ansible Tower. However, it broke automation for client projects like Approvals. Those playbooks call the postgresql role, but they run primarily against the VMs that run the code, not the postgresql server VMs, and only check that the relevant database exists and is usable. Those playbook runs do not know (or care) what the OS flavor of the postgreSQL server might be.

To address this failure, this PR adds logic to the main tasks file of the postgresql role to divide client runs from server runs. Client playbooks should now run very few tasks from this role.


acozine:
I didn't make this change on the first go-round, but I'm wondering . . . 

Would the postgresql role be less confusing if, instead of using the boolean var `postgresql_is_local` we used something like `postgres_type` and set the value to `client` or `server`? 

Advantages: 

- We'd avoid that minute of thinking "Wait, in this case, IS postgres local? What does that mean?" 
- With a non-boolean var the conditions become easier to set.

Disadvantages:

- That var is set in a whole boatload of places, so it would be a big changeset.

acozine:
Another question: do we ever need to reload or restart the postgreSQL server if changes are made to a client DB? If this PR gets merged, we will never do that . . .

kayiwa (approved):
🥇 OK! Thanks @acozine

tpendragon (approved):
Seems better!

I wonder if the "client" tasks should really be in their respective notebooks? It's almost like we're writing a module via a role sub-task or something here. Like if we have to pass all the variables anyways, is it simpler than just having the postgresql_db task in the associated playbook? Not blocking though, this is def clearer.

Feel free to hit the green button, I just didn't because I wanted you to read that and I always miss comments on closed PRs. 😀

sandbergja (approved):
Thanks, @acozine !

acozine:
@tpendragon thanks for the comment. I see the appeal of "avoid the sub-task", but if we move those tasks into the associated client playbooks or roles it would mean a lot of duplication - and if the tasks ever needed to change, it would be a lot of maintenance.

PR #7195, opened 2026-06-04
[Postgres] Configure pg_hba as a client task

acozine:
In #7191 we divided the server tasks from the client tasks. However, we missed the task that configures pg_hba for client access. This won't break existing systems, but it will cause a failure the next time we rebuild a server from scratch.

This PR reinstates the task to configure the pg_hba file as part of a client-side run of the postgresql role. 

It also removes the `PostgreSQL` prefix from task names, since Ansible now prints out the role name when it runs, so the prefix is unnecessary.

### DIFF
--- a/roles/postgresql/tasks/client.yml
+++ b/roles/postgresql/tasks/client.yml
@@ -1,5 +1,5 @@
 ---
-- name: PostgreSQL | create remote postgresql database
+- name: Create remote postgresql database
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
Thanks, @acozine , good catch!

acozine:
I re-ran the DSS playbook on staging from this branch, and it succeeded.

PR #7196, opened 2026-06-04
[Postgres] Use non-boolean vars

acozine:
Follows up on #7191.

This PR explores what it would look like to use a var like `postgres_db_type` instead of `postgresql_is_local` to control the logic in the postgresql role.

We currently have a few places that use `postgres_is_local` instead of `postgresql_is_local`, and we have the var set in `group_vars/all`, which surprised me. We also have some duplication across environments.

This PR:
- changes the variable from the boolean `postgresql_is_local` to `postgres_db_type`
- sets the default in the postgresql role to `server`
- sets the value in most group vars files to `client`
- removes the var from `group_vars/all` because the same value does not apply across all our inventory
- where applicable, moves the var from environment-specific group vars files into the `shared` or `main` or ` common` group vars file for a project
- removes conditions on handlers, which should run any time they are called
- removes conditions on other tasks where the condition is applied to an include

### DIFF
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -114,8 +114,6 @@ maria_db_cluster_host: "{{ vault_maria_db_cluster_host }}"
 # change the libwww role to use this???
 percona_xtra_cluster_host: "maria-staging"
 running_on_server: true
-postgresql_is_cloud: false
-postgresql_is_local: false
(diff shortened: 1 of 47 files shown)


acozine:
Closing in favor of #7217. Remaining work covered by #7220.

PR #7201, opened 2026-06-05
[Postgresql] remove misleading/obsolete postgresql vars from group_vars/all/

acozine:
While trying to install a database for the Ansible upgrade, I ran into a failure that referenced `lib-postgres3.princeton.edu`. That's a long-gone machine and I couldn't figure out where the reference was coming from. It turned out the `group_vars/all/vault.yml` still pointed to `lib-postgres3` (note: not `lib-postgres-prod3` but the original `lib-postgres3`) and in `group_vars/all/vars.yml` the postgres host pointed to that vaulted variable. 

We've treated `group_vars/all` a bit like a set of defaults, but values there it take higher precedence than role defaults, and this may cause unexpected behavior. Also, we haven't maintained the file diligently - the postgres version in this file was still set to `10`!

This PR changes both group_vars/all files:

- in vault.yml it removes the `vault_postgres_host` variable - this one pointed to `lib-postgres3`
- in vars.yml it removes four postgres-related variables:
  - deletes `postgres_version` - we set that to `15` in `roles/postgresql/defaults/main.yml`
  - deletes `postgres_host` - this was pointed at the vaulted var, so it's worse than useless
  - deletes `postgresql_is_cloud` - this var only appeared in this file, so we are not using it
  - deletes `postgresql_is_local` - we set that to `false` in `roles/postgresql/defaults/main.yml`

My gut says we could also remove the user and password vars from group_vars/all, but I'm not sure if they are being used by inheritance. Leaving those alone for now.


sandbergja (approved):
Thanks, @acozine !

Issue #7197, opened 2026-06-05
Can we migrate observability stack from Datadog to SigNoz?


kayiwa:
We are hoping to move our entire observability platform from Datadog to SigNoz to  lower costs of our telemetry data. SigNoz is an open‑source, OpenTelemetry‑native platform that will serve our metrics, traces, and logs.

In order to estimate the cost of how much this will impact our storage we need to look at the rate our production infrastructure hits our file system

### Scope

- Metrics – currently in Datadog (custom metrics, host metrics, container(?) metrics)
- Traces – APM traces sent to Datadog
- Logs – log ingestion and storage in Datadog
- Dashboards – all critical dashboards used
- Monitors / Alerts – all active alerts and notification rules

### Acceptance criteria

- All production services send metrics, traces, and logs to SigNoz (and not to Datadog).
- The dashboards are available in SigNoz and match Datadog data
- All active alerts from Datadog are replicated in SigNoz and firing correctly.
- A 1‑week parallel run shows no gaps in observability coverage (no incidents missed because of missing data).
- Datadog ingestion has been fully stopped and no billing continues.

PR #7202, opened 2026-06-08
[signoz] configure smtp for signoz

kayiwa:
In order to use alerts configure signoz helm to use the SMTP relay

related to #7197


VickieKarasic (approved):
Thanks @kayiwa

PR #7204, opened 2026-06-08
[EZProxy] add internal IP range to allow list

VickieKarasic:
Add internal IP range to princeton_allow.txt so that on campus workstations can connect to resources without the proxy

Related to ServiceNow ticket: TKT0448806


tpendragon (approved):

kayiwa (approved):
OK! thanks @VickieKarasic

PR #7207, opened 2026-06-08
[PDC Discovery] Switching production to solr9

carolyncole:


kayiwa (approved):
🎉  THANKS! @carolyncole

PR #7212, opened 2026-06-09
[Allsearch API] make passive health checks more permissive on load balancer

sandbergja:
Yesterday, we saw a scenario where passive health checks were failing frequently for both allsearch api boxes, taking the application down for a few seconds at a time.

We have not yet figured out why this is happening: there is not a backlog of requests in the queue on the upstreams.  However, we can make the passive health checks a bit more permissive in the meantime while we investigate.


kevinreiss (approved):
Thanks @sandbergja

PR #7214, opened 2026-06-10
[signoz] synthetic monitoring

kayiwa:
use the checkmk servers to monitor and/alert for http checks

### DIFF
--- a/inventory/all_projects/_orphans.ini
+++ b/inventory/all_projects/_orphans.ini
@@ -6,10 +6,10 @@ analyze-catalog.lib.princeton.edu
 kraken.lib.princeton.edu
 
 [physical_boxes]
-lib-adc1.princeton.edu # obsolete physical LB
-lib-adc2.princeton.edu # obsolete physical LB
+# lib-adc1.princeton.edu # obsolete physical LB
+# lib-adc2.princeton.edu # obsolete physical LB
 # new LBs (VMs) are in nginxplus file
-libserv144.princeton.edu # veridian server
+# libserv144.princeton.edu # veridian server
 # libserv144 is the old postgres4
 # it now hosts Princeton Periodicals and Blue Mountain
 
--- a/inventory/all_projects/cdh.ini
+++ b/inventory/all_projects/cdh.ini
@@ -38,8 +38,10 @@ cdh-test-web2.princeton.edu
 [cdh_shared_production]
 # these groups allow us to add ops, lib, & CDH keys
 # for the pulsys user
-lib-adc1.princeton.edu
-lib-adc2.princeton.edu
+# lib-adc1.princeton.edu
+# lib-adc2.princeton.edu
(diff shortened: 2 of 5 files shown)


group_vars/checkmk/production.yml
```diff
@@ -20,3 +20,114 @@ checkmk_agent_server_protocol: http
...
+
+  httpcheck/standard:
+    collection_interval: "{{ otel_httpcheck_standard_interval }}"
+    targets:
+      - endpoint: https://library.princeton.edu/
+        method: GET
+      - endpoint: https://findingaids.princeton.edu/
+        method: GET
+
+  httpcheck/tls:
```
hackartisan: Discussed: we maybe don't need to check tls here, since we are doing it with checkmk?

group_vars/checkmk/production.yml
```diff
@@ -20,3 +20,114 @@ checkmk_agent_server_protocol: http
...
+otel_httpcheck_service_name: synthetic-httpcheck
+otel_httpcheck_environment: production
+otel_httpcheck_location: pul-production
+
+otel_httpcheck_critical_interval: 30s
+otel_httpcheck_standard_interval: 300s
+otel_httpcheck_tls_interval: 3600s
+
+otel_receivers:
+  httpcheck/critical:
```
hackartisan: do we need critical and standard? It's a nice feature to have available but we might want to consolodate unless / until we have different types of checks we want

group_vars/checkmk/production.yml
```diff
@@ -20,3 +20,114 @@ checkmk_agent_server_protocol: http
...
+        method: GET
+      - endpoint: https://dpul.princeton.edu/health
+        method: GET
+
+  httpcheck/standard:
+    collection_interval: "{{ otel_httpcheck_standard_interval }}"
+    targets:
+      - endpoint: https://library.princeton.edu/
+        method: GET
+      - endpoint: https://findingaids.princeton.edu/
```
hackartisan: this app does also have a health endpoint

inventory/all_projects/synthetic_signoz.ini
```diff
@@ -0,0 +1,4 @@
+[synthetic_signoz_production]
+pulmonitor-prod1.lib.princeton.edu
```
hackartisan: Discussed: let's move these groups to checkmk.ini

hackartisan (changes requested):
This looks great and I'm happy to merge once the things we discussed are done.

Nonblocking! I can see two questions, both about alerts:
- is it possible to require x number of failed checks before alerting? I guess this would be set up on the alert side, which isn't really what this PR is doing.
- @carolyncole mentioned a scenario where during some hours you might want no alerts, during other hours you might want alerts after 3 minutes, and during some other hours you might want an alert after 30 minutes.

PR #7215, opened 2026-06-10
[solr8] remove all references to solr8

kayiwa:
remove references to solr8 on our inventory


acozine:
We also need to remove the production groups from production.ini. That can be a separate PR, though.

PR #7216, opened 2026-06-11
Enable nomad storage namespace mapping

JaymeeH:
ref [#2726](https://app.zenhub.com/workspaces/rdss-61a4f1a12a399b001730f65a/issues/gh/pulibrary/tigerdata-app/2726)


JaymeeH:
The work here is redundant/unneeded since we configure the filesystem permissions and network for the top level pod at runtime with `--rm --security-opt label=disable --security-opt unmask=ALL -p 8888:8888 --user podman --device /dev/net/tun --device /dev/fuse`

PR #7217, opened 2026-06-11
[PostgreSQL] adds a role for postgreSQL client actions

acozine:
Follow up to #7184, #7191, and #7195.

The current `postgresql` role includes tasks for building a postgres server, building a postgres cluster, and creating a database for use by a client application. This PR continues the work of separating out those tasks.

This PR:
- creates a psql-client role
- tests the psql-client role by creating two containers: one for the client application and one for the database server
- removes the extra dependency on the postgresql role from the approvals role
- uses FQDNs instead of IPs for enabling connections to our postgresql servers

We agreed to try moving one client application over to the `psql-client` role at a time. However, this will be challenging, because the `ruby_app` role lists a dependency on the `postgresql` role. If we change that dependency from `postgresql` to `psql-client`, all applications that call the `ruby_app` role from their playbooks will switch at the same time. 

Also, right now the tests for those client applications fail if we switch them over. The `psql-client` tests require two containers and a network, which are created by settings in `roles/psql-client/molecule/default/molecule.yml` and tasks in `roles/psql-client/molecule/default/prepare.yml`. Any other role that calls the `psql-client` role as part of its testing currently only runs the tasks in the `psql-client` role, skipping the `prepare.yml`, which means the tests fail for lack of preparation.

### DIFF
--- a/.github/workflows/molecule_tests.yml
+++ b/.github/workflows/molecule_tests.yml
@@ -60,6 +60,7 @@ jobs:
           - php
           - postfix
           - postgresql
+          - psql-client
           - pulfalight
           - pul_nomad
           - pulmap
@@ -131,4 +132,3 @@ jobs:
         run: |
           docker ps -a
           docker logs $(docker ps -aq --filter name=instance) || true
-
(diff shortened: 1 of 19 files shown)


acozine:
If we added the `prepare.yml` to `roles/ruby_app/molecule/default` and updated the `molecule.yml` there to include the changes from the `psql-client` role, we might get passing tests after the switch. We'd still have to switch all the ruby apps at once, though.

PR #7221, opened 2026-06-12
[tigerdata-app] update Ruby version to 3.3.2

jrgriffiniii:
- Update ruby_version_override in group_vars/tigerdata/common.yml to ruby-3.3.2
- Add Python/uv-related workspace configuration files (.python-version, Pipfile, main.py, mise.toml, pyproject.toml, uv.lock)


bess (dismissed):
@jrgriffiniii It looks like this PR changes more than the ruby version for tigerdata... it's making changes to princeton_ansible itself. If those are actual changes you want to propose, would you please put them in a separate PR? 

I'm going on vacation, no need to tag me for review again. This looks great to me as long as the princeton_ansible changes are removed. I don't object to them, maybe they are a great idea! But let's please make one logical change per PR to make our own lives easier.

PR #7224, opened 2026-06-15
[microk8s] Allow routed MicroK8s pod traffic through UFW

kayiwa:
Add UFW route rules for Calico VXLAN and pod CIDR traffic so pod-to-pod
forwarded traffic is not blocked by the host firewall. This prevents DNS
traffic such as vxlan.calico -> cali+ UDP/53 from being dropped.

Also keep explicit MicroK8s interface rules for vxlan.calico and cali+.

We suspect this to have impacted the staging hosts
Screenshot 2026-06-15 at 8 24 48 AM

### DIFF
--- a/group_vars/pulmap/staging.yml
+++ b/group_vars/pulmap/staging.yml
@@ -166,9 +166,9 @@ otel_exporters:
     resolver:
       static:
         hostnames:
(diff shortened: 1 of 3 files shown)


JaymeeH (approved):
Thanks @kayiwa !

Issue #7237, opened 2026-06-18
Upgrade towerdeploy Ruby runtime from EOL Ruby 3.1


kayiwa:
The towerdeploy Molecule test currently runs with Ruby 3.1.0 in the Ubuntu 22.04 test image. Ruby 3.1 is end-of-life, and newer gem releases are beginning to assume Ruby 3.2+ APIs.

This surfaced in CI when cap --version failed after installing Capistrano 3.19.2 and the latest i18n gem:

```sh
undefined method `[]' for Fiber:Class (NoMethodError) current = Fiber[:i18n_config] || self.config = I18n::Config.new
```

Ruby 3.1 does not provide Fiber.[], which newer i18n releases use. We applied a short-term workaround by pinning i18n to a Ruby 3.1-compatible version, but this should not be the long-term state.

kayiwa:
This work will be complete if it includes a README on what this role actually does.

kayiwa:
https://github.com/pulibrary/princeton_ansible/pull/7238#pullrequestreview-4525554931 (also make it possible to update versions)

kayiwa:
by way of @acozine 

- to test the Ruby upgrade on TowerDeploy:
  - create a new towerdeploy_staging VM, add it to inventory and merge
  - update the Prancible project in Tower
  - update the Prancible inventory in Tower
  - in a branch, make the ruby changes
  - in the same branch, change the ‘hosts’ line in the towerdeploy playbook to run against towerdeploy_staging
  - edit the Tower Template for 1 Cap Deploy and check “prompt on launch” for Source Control Branch
  - test the changes by running the Tower Template and selecting the  branch - changes should run against the staging VM

PR #7238, opened 2026-06-18
[towerdeploy] Pin i18n for Ruby 3.1 compatibility

kayiwa:
Pin the i18n gem to 1.14.8 for towerdeploy because newer 1.15.x releases call Fiber.[], which is only available in Ruby 3.2+.

Replace shell-based gem installs with community.general.gem so the role is idempotent and can pin versions cleanly.

### DIFF
--- a/roles/towerdeploy/tasks/main.yml
+++ b/roles/towerdeploy/tasks/main.yml
@@ -1,32 +1,30 @@
 ---
 # tasks file for towerdeploy
 
-- name: towerdeploy | install capistrano
-  ansible.builtin.command: gem install capistrano --version "{{ cap_version }}"
-  register: gemoutput
-  # note: upgrading the cap version will not report 'changed'
-  changed_when: gemoutput.rc != 0
-  failed_when: gemoutput.rc !=0
-
-- name: towerdeploy | install capistrano dependencies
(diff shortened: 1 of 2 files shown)


roles/towerdeploy/tasks/main.yml
```diff
@@ -1,32 +1,30 @@
 ---
 # tasks file for towerdeploy
 
-- name: towerdeploy | install capistrano
-  ansible.builtin.command: gem install capistrano --version "{{ cap_version }}"
-  register: gemoutput
-  # note: upgrading the cap version will not report 'changed'
-  changed_when: gemoutput.rc != 0
-  failed_when: gemoutput.rc !=0
```
acozine, line 9: Do we need these for future upgrades of capistrano?

acozine (commented):
This looks great! It's much cleaner and simpler. 

Just one question, see below.

roles/towerdeploy/tasks/main.yml
kayiwa: can this be addressed here? https://github.com/pulibrary/princeton_ansible/issues/7237

acozine: yep, sounds good

PR #7240, opened 2026-06-18
Separate BigFix install from Rapid7 and Cortex XDR

acozine:
We are auditing our infrastructure for the presence of the two OIT-mandated security tools, Rapid7 and Cortex XDR. We found a number of hosts that did not have Rapid7 installed. This was caused by failures in the BigFix install tasks, which meant the rest of the security playbook did not run on those failed hosts.

BigFix is an additional automation tool, so it's only a partial fit for a group of security tools. Also, we download the BigFix agent from an on-prem server with no external access, which means cloud resources currently fail these tasks.

To avoid blocking security tool installation while we figure out the BigFix installation, this PR:
- separates the BigFix install tasks into a separate playbook

We can define which machines need the BigFix client and work to install it as needed.


kayiwa (approved):
OK! thanks @acozine

Issue #7099, opened 2026-05-07
Isolate maintenance failures


acozine:
### What maintenance needs to be done?
We have a few VMs that consistently cause our maintenance playbooks to fail. We should isolate these from the rest of the inventory temporarily, allowing us to see unexpected failures more easily. 

One host can be in multiple groups. If we create a "usual suspects" or "sad trombones" group in inventory, and add these VMs to it, we can run the main maintenance playbooks with `!sad_trombones` and have a separate run for the new group. As we fix whatever issues are causing the failures on these VMs, we can remove them from the group so we will see regressions clearly too.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
We need to ensure all VMs are reachable and maintainable with our automation tools.

### Acceptance criteria
 - [ ] we have a new group in inventory
 - [ ] VMs that have failed maintenance in the last month are added to this group
 - [ ] playbooks in Tower run against `hosts: usual_group:!problem_group`
 - [ ] we add a template that only runs against the problem group

VickieKarasic:
We looked through recent os_updates failures in staging and think the ones that are still failing are fixable manually (we fixed 3 out of 5 that failed recently). See example [Tower output](https://ansible-tower.princeton.edu/#/jobs/playbook/58919/output) for recent staging failures. 

In production, the lib-vm-bitcur{1-13} consistently fail because they are not running a supported OS, so it would make sense to update the playbook to not run against these so that os_updates failures are actionable.

Issue #3915, opened 2023-05-18
Epic: separate maintenance tasks from on-demand tasks


acozine:
Most of the tasks in the `common` role need to be done when we build a new VM and as regular maintenance on all VMs. Right now the `common` role runs as a dependency of almost every other role we have. This means that any failures in the `common` role, such as an inability to install a utility on a particular VM, become blockers for on-demand tasks such as deploying code.

Can we develop maintenance playbooks that we can run on a schedule - similar to the apt update playbook - to maintain all utilities across our infrastructure? This would separate maintenance from other tasks and allow us to debug any issues with maintenance without interfering with on-demand tasks.

Related to #2685 because that ticket also involves removing the `common` role from the dependency chain for roles.

acozine:
Related to / includes #4120

acozine:
This work ties in with a plan for two-part automation for virtual infrastructure:
- part one: build a generic VM
  - select levels of CPU, memory, storage
  - install shared tools and packages
  - configure authentication and authorization (SSH keys or other solution)
  - install and run security tools, including the local firewall
  - connect to monitoring platform
- part two: specialize the VM for its intended purpose
  - install application-specific tools and packages
  - confgure database and indexing connections
  - deploy code

Maintenance tasks should re-run part one against existing infrastructure regularly, ensuring that all the generic tools and connections get updated.

PR #7241, opened 2026-06-18
adds a playbook that just tests connectivity

acozine:
As part of updating OS packages, installing security tools, and other tasks that apply across our infrastructure, we have struggled with unreachable hosts.

This PR adds a playbook that just checks whether Ansible can reach all hosts in our inventory. This should help us identify and correct connection issues.

Related to #7099 #3915


kayiwa (approved):
🔌  thanks @acozine

PR #7245, opened 2026-06-22
[Security] Installing XDR on FreeBSD is not possible, stop trying

acozine:
The XDR tool will not run on FreeBSD. This PR acknowledges that fact, and adjusts the security-tools installation playbook so it will not fail against FreeBSD machines. The absence of XDR will still be noticeable on the security tools audit whenver we run that.

This PR:
- puts the XDR installation tasks into a block which only runs when the service is not present and the OS is not FreeBSD
- uses a standard syntax for the conditions that check the presence or absence of services: rapid7, rapid7 Scan Assistant, and XDR

### DIFF
--- a/playbooks/utils/security_install.yml
+++ b/playbooks/utils/security_install.yml
@@ -21,80 +21,84 @@
     ansible.builtin.service_facts:
 
   tasks:
-  - name: Download the XDR deb file (Ubuntu)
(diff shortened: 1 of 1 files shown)


acozine:
Rapid7 won't run on BSD either. Added the same changes for those tools.

PR #7250, opened 2026-06-23
change CDH DNS names

kayiwa:


roles/nginxplus/files/conf/http/cdh_prod_web.conf
```diff
@@ -5,8 +5,10 @@ proxy_cache_path /var/cache/nginx/prod_cdhweb/ keys_zone=prod_cdhwebcache:10m;
 upstream prod_cdhweb {
     least_time header 'inflight';
     zone prod_cdhweb 64k;
+    # server cdh-web1.princeton.edu;
+    server cdh-web2.princeton.edu;
     server cdh-web3.princeton.edu;
-    server cdh-web4.princeton.edu;
+    # server cdh-web4.princeton.edu;
```
rlskoeser, line 11: Don't you want to remove cdh-web3 as well?  Any reason to keep 3 and 4 around in comments?

You should probably also update or remove the comment on line 4, which is no longer accurate.

rlskoeser (commented):
Thanks for shuffling the htr boxes around and cleaning up the old cdh web VMs.

Unless I'm misreading the change, you still have cdh-web3 listed - I think it should be removed entirely.

rlskoeser (approved):
OK

Issue #7232, opened 2026-06-17
Request new staging VMs for CDH eScriptorium project


tanhaow:
### Relevant deadlines
If the new VMs must be ready by a certain date, please enter it here.
- [ ] We need these VMs by: MM/DD/YYYY

### Describe the VM(s) you need

If you do not know the answers to some of these questions, that's okay. Please add the information you do have, and we will work together to gather any other information we need. You can also look at the [documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/new_vms_for_devs.md) in pul-it-handbook.

Which project are these VMs for?
- [ ] Project: [Princeton-CDH/htr2hpc](https://github.com/Princeton-CDH/htr2hpc/milestones)

How many VMs do you need?
- [ ] Number of VMs: 2

What should the FQDN(s) be? (for example, our-new-project-staging1.princeton.edu):
- [ ] FQDN: We'd like to reuse `cdh-test-htr1.lib.princeton.edu` and `cdh-test-htr2.lib.princeton.edu`, but the current VMs with those names haven't been decommissioned yet (we're monitoring production for ~2 weeks before decommissioning). Do you have any suggestions? Ideally we want to stay consistent with the `cdh-test-` naming convention used by our other staging VMs.

What virtual resources does each VM need?
- [ ] Number of CPUs per VM: 2
- [ ] Amount of memory per VM (in GB): 16
- [ ] Amount of storage/disk space per VM (in GB): 50 (current VMs have 28GB root disk which has been insufficient)

What firewall changes should be made? (for example, port 22 for SSH, etc.)
- [ ] Port number(s): 22 (SSH), 80 (HTTP), 443 (HTTPS)
- [ ] Service(s): nginx, Passenger

What else should we add to each VM? (for example, mount Isilon drives, add NFS server access)
- [ ] Additional resources: NFS mount (same as current staging: `lib-fs-staging.princeton.edu`), access to staging PostgreSQL (`lib-postgres-staging1.princeton.edu`)

### Implementation notes, if any

These are replacement staging VMs for eScriptorium. We have successfully migrated to production (`htr.cdh.princeton.edu`) and are now hope to up a new staging environment. Configuration should match the current staging VMs (`cdh-test-htr1/2.lib.princeton.edu`).

rlskoeser:
[...]

PR #7254, opened 2026-06-24
add new htr machines to NFS

kayiwa:


acozine:
Part of #7232. The legacy VMs will be decommissioned once we validate their replacementns.

PR #7265, opened 2026-06-29
I7260 r7 checksum

aruiz1789:
Updated the package download checksum. If the package version changes frequently, we can remove the checksum or internally host the package.


kayiwa (approved):
OK! Thanks @aruiz1789

PR #7266, opened 2026-06-29
Updating the ruby version to hopefully get a newer default stringio

carolyncole:
refs #7267 

This PR was run on the servers with no harm done.  That said it did not change the stringio version to `3.2.0`.  Only running an update to the gem solved the issue.


carolyncole:
the default version of stringio was `3.1.1` after running this PR

carolyncole:
This is old.  I hpoe we fixed this a different way

PR #7269, opened 2026-06-29
switch to certinext/emsign CA API

kayiwa:
move away from sectigo API


roles/ezproxy/tasks/main.yml
```diff
@@ -105,7 +105,7 @@
   when: running_on_server
 
 - name: Ezproxy | update acme certificates for {{ domain_name }}
```
acozine, line 106: Good catch, I'm glad we got this one fixed before the migration.

roles/ezproxy/tasks/main.yml
```diff
@@ -105,7 +105,7 @@
...
-  ansible.builtin.command: /snap/bin/certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme.sectigo.com/v2/InCommonRSAOV --eab-kid {{ vault_acme_eab_kid }} --eab-hmac-key {{ vault_acme_eab_hmac_key }} --domain *.{{ domain_name }}.princeton.edu --cert-name {{ domain_place_name }}
+  ansible.builtin.command: /snap/bin/certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme-us.certinext.io/v1/directory --eab-kid {{ vault_acme_eab_kid }} --eab-hmac-key {{ vault_acme_eab_hmac_key }} --domain {{ domain_name }}.princeton.edu --domain *.{{ domain_name }}.princeton.edu --cert-name {{ domain_place_name }}
```
acozine: Do we need both `ezproxy.princeton.edu` AND `*.ezproxy.princeton.edu`? That seems like overkill (and more than we currenlty have). What did I miss?

acozine (commented):
Good thinking.

I just want to double check on the domain names.

VickieKarasic (approved):
Thanks @kayiwa - looks good to me after @acozine's comments are resolved.

roles/ezproxy/tasks/main.yml
VickieKarasic: I don't think we do, and it's breaking the way ezproxy is working. I pushed some changes that remove the domain_name without the *. at the beginning, but now the playbook is failing on test at the update acme certificates task.

kayiwa: I actually think we need both when you authenticate to use the admin UI we use ezproxy.princeton.edu once that's done the ezproxy application generates the wildcards for db connections.

PR #7270, opened 2026-06-29
[bibdata] bibdata role needs psql-client role

kayiwa:
dependency was not on the meta so we add it.
unbreaks main


acozine:
Needs the changes to molecule.yml and a prepare.yml playbook, and probably the psql variables in the defaults/main.yml, to make the psql-client test pass.

acozine (approved):
Looks good. I don't understand how the bibdata role depends on ruby_app, but obviously it does, so I'm going to run with it.

kayiwa:
> Looks good. I don't understand how the bibdata role depends on ruby_app, but obviously it does, so I'm going to run with it.

It definitely does.  https://github.com/pulibrary/princeton_ansible/blob/996468fd1c4b88b3f7ebb33011217c9acdb17c5c/playbooks/bibdata.yml#L37

PR #7272, opened 2026-06-30
[orangelight] role needs psql-client role

kayiwa:
allow molecule to pass on CI

this unbreaks main


regineheberlein (approved):
Thanks @kayiwa !

PR #7273, opened 2026-06-30
lib-storage exemption from autoreboot

kayiwa:
while @regineheberlein is working on this long running process this
      server will require manual reboots


regineheberlein (approved):
Thanks so much @kayiwa !

PR #7275, opened 2026-06-30
rdss zero config observability

kayiwa:
ship production logs traces and metrics to production

related #7197


JaymeeH (approved):
Thanks @kayiwa

PR #7276, opened 2026-06-30
update postgres host to point to v18 server

JaymeeH:
ref [#347](https://app.zenhub.com/workspaces/rdss-61a4f1a12a399b001730f65a/issues/gh/pulibrary/orcid_princeton_hanami/347)


JaymeeH:
We needed to manually install additional psql packages from here: https://wiki.postgresql.org/wiki/Apt

PR #7257, opened 2026-06-24
[postgres] Switch ruby app role to use psql-client for database tasks

acozine:
Follows up on https://github.com/pulibrary/princeton_ansible/pull/7217.

Closes https://github.com/pulibrary/princeton_ansible/issues/7220.

This PR simplifies the psql-client role:

- removes sub-task files that only had one task
- removes the database dump step
- delegates user and database creation to the postgresql server so we do not need to install postgresql libraries on client VMs

The PR also updates the ruby_app role:

- changes the role's dependency from postgresql to psql-client
- adds a postgresql container to the tests
- creates a network for molecule to test within
- adds default values for the postgresql testing
- runs the converge steps on the main container (not on the postgresql container) so we are testing delegation correctly

Finally , the PR also changes the molecule testing for other roles that depend on ruby_app, either directly or indirectly. These roles are:

- approvals
- blacklight_app
- dpul
- dss
- figgy
- lae
- lib_jobs
- lockers_and_study_spaces
- oawiaver
- pulfalight
- pulmap
- repec
- resque_worker
- sidekiq_worker
- sneakers_worker

PR #7277, opened 2026-06-30
[postgresql] Focus on server tasks in postgresql role

acozine:
Now that the psql-client role is in use for automating client-side tasks, we can focus the postgresql role on server-side tasks.

This PR:
- removes the `postgres_is_local` var and all its uses
- removes the handler for restarting the remote database service (that is now done on the client side) 
- removes the `client.yml` tasks file
- renames the `server.yml` tasks file to `main.yml`, simplifies the logic for choosing which tasks to execute


acozine:
Related o #7217 and #7257

kayiwa (approved):
OK! Thanks @acozine

Issue #7267, opened 2026-06-29
StringIO version broke passenger


carolyncole:
Not entirely sure what caused passenger to break, but we started getting the following errors:
```
Error: The application encountered the following error: You have already activated stringio 3.1.0, but your Gemfile requires stringio 3.2.0. Since stringio is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports stringio as a default gem. (Gem::LoadError)
```

To solve this @JaymeeH ran by hand `gem update stringio`

We need to integrate this update into our ansible playbook so that if we rebuild the machine it will come back up with the correct stringio as a part of it.

carolyncole:
I could not in a reasonable amount of time get tigerdata to upgrade to 4.0 https://github.com/pulibrary/tigerdata-app/pull/2737 Circle Ci kept failing.  I'm going to add the gem update to our playbook.

PR #7279, opened 2026-07-01
[tigerdata] Update the stringio gem to make sure passenger keeps running

carolyncole:
refs #7267


leefaisonr (approved):
Thanks @carolyncole

PR #7284, opened 2026-07-02
Update orangelight.yml to use psql-client role

acozine:
The Orangelight playbook has been failiing its daily run.

Althogh the role started testing the psql-client functionality  in #7272, the playbook was still calling the postgresql role. This updates the playbook.


christinach (approved):
Thanks @acozine !

PR #7285, opened 2026-07-02
require single reboots of vms

kayiwa:
prevent all our k8s hosts from rebooting all at once.


acozine:
Love the alphabetical order - thanks!

PR #7302, opened 2026-07-10
[cdh] add test-htr.cdh serverblock

kayiwa:


tanhaow (approved):
Looks good. Thank you, Francis!

roles/nginxplus/files/conf/http/dev/cdh_test_htr.conf
```diff
@@ -19,13 +19,22 @@ map $http_upgrade $connection_upgrade {
...
 
+server {
+    listen 80;
+    server_name test-htr.lib.princeton.edu;
+
+    location / {
+        return 301 https://test-htr.cdh.princeton.edu$request_uri;
+    }
+}
+
```
rlskoeser: We would prefer to retire the pilot instance url and not redirect it - in case any pilot users have it bookmarked, we don't want them finding our staging test instance.

PR #7303, opened 2026-07-13
[nginx] add conf file for proxmox

VickieKarasic:
adds new nginxplus conf file for ProxMox LB alias name lib-px-staging.lib.princeton.edu


acozine (approved):
I'm a little confused about the difference between sandbox and staging in this context. But if y'all are happy, I'm happy.

PR #7310, opened 2026-07-14
Remove lib-screen01L from inventory

acozine:
We have seen failures for maintenance playbooks against `lib-screen01L` - that's because we decommissioned it. This VM was a test firewall device from 2024.

This PR removes it from inventory.


VickieKarasic (approved):
Thanks @acozine

PR #7316, opened 2026-07-15
[psql-client] Fix PGDG package priority in psql-client role

kayiwa:
Add an APT preference for PostgreSQL client and libpq packages so PGDG packages take precedence over the Princeton Ubuntu mirror. This prevents dependency conflicts when installing versioned PostgreSQL clients while remaining compatible with PostgreSQL 15 and 18.


leefaisonr (approved):
Thanks @kayiwa !

PR #7319, opened 2026-07-16
[ezproxy] add logrotation

### LINKED ISSUE #7315: [EZproxy] We should rotate the EZproxy logs
### Expected behavior
Logs rotate regularly, keeping clear space on the server.

### Actual behavior
The EZproxy logs are in a weird location. On the production server, we have logs dated back to January, 2026. We are seeing disk-usage warnings as a result.

### Steps to replicate
Look at ezproxy-prod1.princeton.edu, in the `/var/local/ezproxy/logs` directory.

### Impact of this bug
EZproxy may stop working if the disk gets too full.

Path for the logs is `/var/local/ezproxy/logs/` - the naming pattern is also odd.

kayiwa:
use common role pattern to add logrotation

closes #7315

### DIFF
--- a/group_vars/ezproxy/common.yml
+++ b/group_vars/ezproxy/common.yml
@@ -5,19 +5,21 @@ shib_admin_user:
   - kr2@princeton.edu
   - sc9047@princeton.edu
   - vk4273@princeton.edu
-logrotate_rules:
+# ezproxy logrotate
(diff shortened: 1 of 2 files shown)


kevinreiss (approved):
Thanks @kayiwa

PR #7320, opened 2026-07-16
[ORCID] Update postgres host name

JaymeeH:
fixes typo in the production group vars for the postgres host


kayiwa (approved):
OK! thanks @JaymeeH

PR #7325, opened 2026-07-17
[Security] streamlines install and audit of mandated security tools

acozine:
We were getting disk-usage alerts on lib-redis-staging1. I deleted the package file for Cortex XDR to make a little room. Then I thought, "Hey, we could just always do this."

Also, our security audit playbook kept reporting that the mandated security tools are not installed on our FreeBSD machines. Which is true, but not actionable, since neither tool will run on FreeBSD.

This PR:
- adds tasks to delete the package (.deb/.rpm) files after installing Cortex XDR 
- adds a task to report FreeBSD machines separately in our security audit
- adds conditions to two other tasks to skip FreeBSD machines when reporting non-compliance


kayiwa (approved):
OK! thanks @acozine

Issue #7333, opened 2026-07-20
Use alias to mount OIT Samba servers


acozine:
### What maintenance needs to be done?
OIT has migrated the alias `dms-smbserve` to a new server, `aissmb226l`. However, some library resources are still connecting to the old server. This is because we hard-coded the IP for the old server in several of our mounts.

Related to #7271 
Related to #4029 
Related to #1794

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
The old Samba server is running an obsolete OS and needs to be retired.

### Acceptance criteria
 - [x] We no longer mount the HR share for bibdata
 - [x] We mount the Samba share using the alias, to prevent this problem from cropping up in future
 - [x] Kevin Hoffman confirms that no connections are hitting the old Samba server from library resources

### Implementation notes, if any
The hr_share role hard-codes the IP for its share [in its main tasks file](https://github.com/pulibrary/princeton_ansible/blame/7b36ef71f3f8a49a36e2f9fca8277e5cd0e888cf/roles/hr_share/tasks/main.yml#L19).
The lib_jobs role does the same, in [role defaults](https://github.com/pulibrary/princeton_ansible/blob/7b36ef71f3f8a49a36e2f9fca8277e5cd0e888cf/roles/lib_jobs/defaults/main.yml#L27)  

The following machines are still connecting to the old Samba server as of today:

- bibdata-prod1
- bibdata-prod2
- bibdata-worker-prod1
- biodata-worker-prod2
- bibdata-staging1
- bibdata-staging2
- bibdata-worker-staging`
- bibdata-worker-staging2
- bibdata-qa1
- bibdata-qa2
- bibdata-worker-qa1
- bibdata-worker-qa2
- lib-approvals-prod1
- lib-approvals-prod2
- lib-approvals-staging1
- lib-approvals-staging2
- lib-jobs-prod1
- lib-jobs-prod2
- lib-jobs-staging1
- lib-jobs-staging2
-

acozine:
Also related to #2293 and #2294?

acozine:
@jkazmier-PUL suggested the original blocker may have been issues with the VMs' DNS servers. If the mount is still problematic, check `ifconfig -all` and check all the DNS servers.

acozine:
[...]

PR #7334, opened 2026-07-20
bibdata does not need the HR share any more

### LINKED ISSUE #4029: Remove Samba Mount for HR Server
The mount at  /mnt/dms-smbserve/bi-library-hr is no longer needed. At one time we loaded information on who was eligibile to access campus during the covid period but that is long deprecated.

### LINKED ISSUE #7271: Bibdata does not need to Mount the HR Share
We no longer need to mount the shared folder (mnt/dms-smbserve/bi-library-hr/prod) referenced in the BIBDATA_ACCESS_DIRECTORY environment variable. This share is [mounted](https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/bibdata.yml#L43) through the HR role.

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

- Bibdata no longer uses this share. 

### Acceptance criteria
 - [ ] Remove Variable references for BIBDATA_ACCESS_DIRECTORY
 - [ ] Remove reference to [hr share](https://github.com/pulibrary/princeton_ansible/blob/main/playbooks/bibdata.yml#L43)
 - [ ] run against staging env
 - [ ] run against qa env
 - [ ] run against production env.

While bibdata no longer uses this share lib_jobs and approvals still use it for production tasks so we only need to remove it from bibdata.

acozine:
Closes #7271.
Closes #4029.
Related to #7333.

We no longer need to mount the HR shares for bibdata, which is one of the systems that has been connecting to an obsolete Samba server that OIT wants to retire.

This PR:
- removes the BIBDATA_ACCESS_DIRECTORY variables from the group vars for all three bibdata environments
- removes the hr_share role from the bibdata playbook
- asks in a comment whether we still need to send `bibdata_access` logs to our logging service - if we do not, we can remove those variable as well

### DIFF
--- a/group_vars/bibdata/production.yml
+++ b/group_vars/bibdata/production.yml
@@ -76,8 +76,6 @@ rails_app_vars:
     value: '/opt/bibdata/shared/hathi/input'
   - name: HATHI_OUTPUT_DIR
     value: '/opt/bibdata/shared/hathi/output'
-  - name: BIBDATA_ACCESS_DIRECTORY
-    value: '/mnt/dms-smbserve/bi-library-hr/prod'
   - name: SCSB_S3_PARTNER_ACCESS_KEY
     value: '{{ vault_scsb_s3_access_key_prod }}'
   - name: SCSB_S3_PARTNER_SECRET_ACCESS_KEY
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @acozine ! 🐴

PR #7341, opened 2026-07-23
[proxmox] credentials for proxmox are baked into tower

kayiwa:
we remove these credentials since the primary place this playbook will be run in on AAP


VickieKarasic (approved):
Ok - thanks @kayiwa

Issue #7219, opened 2026-06-11
[Tower] Integrate the ChecKMK playbook into the replace-a-vm workflow


acozine:
### User story
As a developer, I want to replace a VM in two steps: run the replace-a-vm workflow in Tower, then run the application-specific playbook. 

### Acceptance criteria

- [ ] When I run the replace-a-vm workflow, I get a new VM that is ready to have an application installed - it has our security tools installed, our SSH public keys added, and an updated host and monitoring in CheckMK 

### Implementation notes, if any
The current blocker is that the CheckMK playbook requires some variables we do not yet pass to the workflow, including the folder where the VM should be listed. We recently review VMware so the folder structure there reflects the folder structure in CheckMK - could we pull that data and use it in the CheckMK playbook? Otherwise, look at the Tower template to add a CheckMK host and see how we can integrate it.

hackartisan:
I sort of wonder if the checkmk tasks are part of application setup, rather than replacing a vm. Like, maybe checkmk stuff should update every time I run an application setup playbook. I think recently some changes were made to a custom monitor script (the ruby one) and they mostly didn't get distributed to machines because you have to run the checkmk playbook separately to get them in place.

I'm not familiar enough with how the checkmk playbook is written to know, but I assume it's sort of like the postgres ones -- there's tasks that need to happen for the server but other things that need to get set up on clients. Running postgres client tasks is part of each application playbook.

acozine:
@hackartisan I believe the local scripts are installed and activated by the roles that require them, not by the CheckMK agent playbook. We can test this, though.

acozine:
I put a bunch of comments onto the wrong ticket, folks. This is what I get for trying to "just finish this one thing" on a Friday afternoon. Moving them now.

acozine:
@kayiwa @VickieKarasic and I worked on this today. Putting in a few notes . . . 

- we made a copy of the workflow template in Tower
- we re-ordered the playbooks in the copy of the workflow template so that OS updates come last to ensure that reboots happen AFTER BigFix is installed
- we added the CheckMK template to the copy of the workflow
- we added the necessary survey questions for the CheckMK agent playbook
- the CheckMK agent template in Tower failed
  - but it worked from Francis' laptop
  - we discovered that the version of the CheckMK collection in Tower was too old for our agent version
  - the CheckMK agent template was running on an EE called CheckMK EE
  - we were going to rebuild the CheckMK EE, but we couldn't find the build instruction file, so
[...]

Issue #7343, opened 2026-07-24
Rebuild tigerdata QA environment


bess:
### Expected behavior
I expect to be able to [replace and rebuild a VM](https://github.com/pulibrary/pul-it-handbook/blob/main/services/replace_rebuild_vm.md). 

### Actual behavior
When I run the Replace a VM playbook for tigerdata-qa1 it tells me it "Failed to find virtual machine tigerdata-qa1", even though I know it must exist because I can ssh to it. 

### Error messages

<img width="1352" height="717" alt="Image" src="https://github.com/user-attachments/assets/b0157f3b-3be0-48d2-9495-95bba460c578" />

<img width="695" height="125" alt="Image" src="https://github.com/user-attachments/assets/3ed9ddec-094b-43e9-8d0c-bf8ff6050617" />

### Steps to replicate
I suspect that this VM might be in a weird state because earlier I attempted to use the other option on tower for replacing a VM. I'd like to build a workflow for tower to automate more of this rebuild process, and Francis encouraged me to look at the work Angel had done. I tried to run it, it didn't seem to do anything, but afterwards I was unable to replace this VM via the other system. 

### Impact of this bug
I can't release this week's Tiger Data code until I can rebuild these systems.

acozine:
The tigerdata-QA1 VM runs in our production VMware environment. This job in Tower couldn't find the VM because it was looking in the staging VMware environment (`"runtime_env": "staging",` shows up in the Variables section of the Details tab of job 53934). 

You can check which environment a VM is in by running the `View VM details` template (which is fairly quick) in your best-guess VMware environment. If the VM is not found, run the same template in the other VMware environment.

acozine:
@leefaisonr the newest workflow Template in Tower should take care of this now.

PR #7347, opened 2026-07-24
uses a different var for the CheckMK agent playbook

acozine:
Partially addresses #7219.

Traditionally we have used `runtime_env` to supply values for two different things:
- the VMware platform a VM runs on (options are  `staging` and `production`)
- the larger inventory group in which a VM can be found (originally `staging` and `production`, plus, more recently, `CI` and `QA`)

When we added files for CI and QA to the `by_environment` section of our inventory, we broke this connection, because QA and CI VMs run in our production VMware environment but they are not included in our production inventory group.

This PR uses a new variable in the CheckMK agent playbook. This allows us to include this playbook in the integrated workflow for replacing a VM - we can pass, for example, `production` as the value for `runtime_env` for VMware and `QA` as the value for `inventory_group` for the CheckMK playbook.

If we merge this PR, we must update the Survey questions in Tower for the CheckMK agent and replace-a-VM Templates so they provide the correct variable name and the correct set of choices.

### DIFF
--- a/playbooks/utils/checkmk_agent.yml
+++ b/playbooks/utils/checkmk_agent.yml
@@ -9,10 +9,20 @@
 # checkmk folder names must be all lower case and should not begin with a slash
 # checkmk site names are: 'production', 'staging', 'aws', and 'gce'
 #
-# by default this playbook runs against inventory from the staging environment
-# to run against a production group or host, pass '-e runtime_env=production'
+# by default this playbook runs against the staging inventory group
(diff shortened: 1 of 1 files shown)


acozine:
Related to #7343.

acozine:
We discussed this today in the Ansible Open House. The consensus was that we'd prefer the simplicity of passing one variable value for both things - the infrastructure environment and the inventory group. 

Closing this PR in favor of #7351

PR #7351, opened 2026-07-27
adds ci and qa to staging and production envs respectively; adds an ADR

### LINKED ISSUE #7219: [Tower] Integrate the ChecKMK playbook into the replace-a-vm workflow

acozine:
Closes #7219.

Supersedes #7347.

Amends our inventory so that every group is part of either `staging` or `production`, reflecing the underlying infrastructure. 

Adds an ADR about where CI and QA resources should run and why we are doing this.

### DIFF
--- a/architecture-decisions/0006-environtments.md
+++ b/architecture-decisions/0006-environtments.md
@@ -1,0 +1,30 @@
+# 1. Align inventory environments with infrastructure
+
+Date: 2026-07-27
+
+## Status
+
+Draft
+
+## Context
+
+We have two infrastructure environments: staging and production.
+
+We have four inventory environments: ci, qa, staging, and production. It's possible we could have more in future.
+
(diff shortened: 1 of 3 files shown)


acozine:
Reviewed in the CheckMK open house, ready for review.

bess (approved):

Issue #7361, opened 2026-07-28
[Security] Rapid7 install glitches


acozine:
### Expected behavior
Rapid7 installs, runs, and updates cleanly on all our infrastructure.

### Actual behavior
The Remediation team notified us that many of our VMs had not connected to the Rapid7 platform in a very long time.

Also, because we download `latest` for installing, but we hard-code the Rapid7 version in `group_vars/all/vars.yml`, every time the version updates, the playbook will fail on new VMs. We should derive the version from the packages instead of hard-coding it.

### Error messages
On a stale system, in the `/opt/rapid7/ir_agent/components/insight_agent/common/insight_agent.stderr` logfile:
```
File "/home/jenkins/agent/workspace/endpoint-installer-linux/bootstrap_common.py", line 43, in check_deps
agent.exceptions.AgentInitializationError: Missing initialization dependencies - ['/opt/rapid7/ir_agent/components/bootstrap/common/ssl/client.key', '/opt/rapid7/ir_agent/components/bootstrap/common/ssl/client.crt', '/opt/rapid7/ir_agent/components/bootstrap/common/ssl/cafile.pem']
```

The `/opt/rapid7/ir_agent/components/bootstrap/common/ssl` directory was empty. 

### Steps to replicate
Connect with the Remediation team, check the log, or check the SSL directory.

### Impact of this bug
We are not getting updated information on the security of our systems and we cannot clear old Remediation tasks.

### Relevant links and code snippets, if applicable

In `/playbooks/utils/security_install.yml`, the task to `Configure the agent with token` was commented out. My guess is that that task started to fail on new VMs when the version updated, and we commented it out temporarily and then forgot to go back and fix it.

acozine:
We may want to look for a different condition for the rapid7 install tasks. On the stale VMs the `ir_agent`service was installed, so the condition "when `ir_agent` not in ansible_facts.services" returned `false` and all Rapid7 install tatsks got skipped. However, the service was not healthy and not connecting to the platform.

PR #7362, opened 2026-07-28
Updates rapid7 version, adds token to vault file, updates configuration

acozine:
Partially addresses #7361.

Creates a clean install process for Rapid7. 

Adds the Rapid7 token to  our vaulted vars.
Adds tasks to derive the version number from the package files when we install, to avoid breakages caused by new releases.
Removes the version number from our variables.
Uncomments the task to configure the Rapid7 agent, which works now that we supply the token and the version number. 

Temporarily removes conditions so we can update existing infrastructure.


acozine:
I ran this successfully against all our staging inventory today. This is ready for review - we can open a separate ticket for finding the right condition for when to skip the Rapid7 install tasks.

playbooks/utils/security_install.yml
```diff
@@ -207,7 +215,7 @@
     - name: Install and configure the Rapid7 Scan Assistant Service except on FreeBSD
       when:
         - ansible_distribution != "FreeBSD"
-        - "'R7ScanAssistant.service' not in ansible_facts.services"
+        # - "'R7ScanAssistant.service' not in ansible_facts.services"
```
carolyncole: Should this be deleted instead of commented out?

playbooks/utils/security_install.yml
```diff
@@ -135,30 +135,28 @@
...
-            - ansible_os_family == "RedHat"
+          loop:
+            - "/opt/{{ cortex_install }}.deb"
+            - "/opt/{{ cortex_install }}.rpm"
 
     - name: Install, configure, and start Rapid7 except on FreeBSD
       when:
         - ansible_distribution != "FreeBSD"
-        - "'ir_agent.service' not in ansible_facts.services"
+        # - "'ir_agent.service' not in ansible_facts.services"
```
carolyncole: Should this be deleted instead of commented out

carolyncole (changes requested):
@acozine looks ok to me, but I am not sure what it is doing.  A couple of clean up questions

acozine:
Per [this KB article](https://princeton.service-now.com/service?id=kb_article&sys_id=KB0013869), we can diagnose issues with Rapid7 using this command: `./ir_agent -diagnose -region us-east-1`.

playbooks/utils/security_install.yml
acozine: Good question. The condition needs to be reinstated (or replaced with a more thorough diagnostic check) after we update our infrastructue and confirm that it's actively connecting. Otherwise we will re-run the installation on every machine every time, which is time-consuming. I left it in, commented, as a guidepost for Future Us, but maybe removing it is better . . .

carolyncole (approved):
Thanks @acozine

PR #7367, opened 2026-07-29
Standardize how we run mailcatcher

bess:
I'd like to try keeping the defaults for mailcatcher whenever possible. Let's not override anything we don't have to, so we keep our config files easier to understand.

Screenshot 2026-07-29 at 3 19 49 PM


carolyncole (approved):
Thanks @bess

PR #7378, opened 2026-07-31
[hr_share] replace IP with alias for the dms sambaserver

acozine:
Partially addresses #7333.

Since the Alma migration we have used a hardcoded IP to mount shares on the central Samba server for sharing data with other departments. The server using that IP is obsolete and needs to be replaced. 

This PR substitutes an alias, so future changes to the IP or to the underlying hardware for this service will not interrupt the mounts for library services.


christinach (approved):
Thanks @acozine !

PR #7379, opened 2026-08-03
[common] fix stop apt duplicate-repository warnings on noble hosts

kayiwa:
Newly provisioned noble machines printed hundreds of "configured
  multiple times" warnings on every apt update, because base provisioning
  always added its own Ubuntu archive list even though 24.04 already ships
  those archives in the newer [deb822 sources format](https://repolib.readthedocs.io/en/latest/deb822-format.html).

  Provisioning now detects the existing archive definitions and only adds
  its own when they are actually missing, removing the redundant file

related to https://github.com/pulibrary/vm-builds/issues/68

### DIFF
--- a/roles/common/molecule/default/verify.yml
+++ b/roles/common/molecule/default/verify.yml
@@ -56,3 +56,25 @@
         name: "{{ common_ubuntu_packages }}"
         state: present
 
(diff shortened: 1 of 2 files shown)


bess (approved):
Thank you, @kayiwa !

PR #7399, opened 2026-08-04
[Nomad] Update Nomad/Consul

tpendragon:
May as well do it while it's easy.

This has been run on sandbox & prod.


kayiwa (approved):
OK! Thanks @tpendragon

PR #7409, opened 2026-08-06
[proxmox] feat(nginx): add redirect to port 8006 for lib-px-staging

kayiwa:
Adds two server blocks:
- listen 80 → return 301 to https://$server_name:8006
- listen 443 → return 301 to https://$server_name:8006

Allows users to reach the service without manually specifying the port.


VickieKarasic (approved):
thanks @kayiwa!

PR #7410, opened 2026-08-06
Add missing health check to DSS

tpendragon:


kayiwa (approved):
OK! thanks @tpendragon

PR #7417, opened 2026-08-08
[mflux] add configurable otel internal metrics endpoint

kayiwa:


bess (approved):
@kayiwa If I understand this correctly, this is the upstream work to what we need in the private `tigerdata-config` repo, which I have seen working. 🏆 
Is that right?

PR #7421, opened 2026-08-10
[tigerdata] Upgrade nodejs and yarn

bess:
Ref https://github.com/pulibrary/tigerdata-app/issues/2779


leefaisonr (approved):
Thanks @bess

PR #7429, opened 2026-08-11
[otel_collector] Fix format of extensions config

sandbergja:
Before this change, we got the following error which prevented the solr playbooks from running:

```
$ /opt/otelcol/otelcol-contrib validate --config /opt/otelcol/config.yaml
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

'extensions' expected a map or struct, got "slice"
```


christinach (approved):
Thanks @sandbergja 🍄 !

PR #7436, opened 2026-08-12
[catalog] Node 26.7

sandbergja:
See also sibling PR https://github.com/pulibrary/orangelight/pull/6100


christinach (approved):
Thanks @sandbergja !

PR #7438, opened 2026-08-12
[Certs] removes Sectigo vars

acozine:
We have migrated our SSL certificates from Sectigo to CertiNext. We added the CertiNext vars when we started the migration, so we could revoke Sectigo certs and create CertiNext certs. 

Now we can remove the Sectigo vars - all future interactions will be with CertiNext.


kayiwa (approved):
OK! 🔏  Thanks @acozine

PR #7443, opened 2026-08-14
Only BSD machines should report that they are running BSD

acozine:
Copy-pasta meant that our security audit playbook was not correctly checking for BSD as an OS.

This PR corrects the logic on our BSD task for the security audit.


sandbergja (approved):
Thanks, @acozine ! 🐴

PR #7445, opened 2026-08-15
[ezproxy] remove unnecessary maxmind ppa from exproxy role

kayiwa:
- install geoipupdate from Ubuntu repositories
- remove redundant apt cache clean/update tasks
- use Ansible apt cache handling for package dependencies
main is failing CI


christinach (approved):
Thanks @kayiwa 🦒

PR #7450, opened 2026-08-17
[firewall] add brownfield firewall

kayiwa:
this allows firewall rule enforcement on brownfield VMs and makes it possible for ubuntu to use UFW and RHEL/Rocky to use firewalld

related https://gitlab.lib.princeton.edu/ops/team-handbook/-/work_items/394


group_vars/zookeeper/common.yml
```diff
@@ -0,0 +1,22 @@
...
+# Additional host firewall rules for ZooKeeper servers.
+firewall_additional_rules:
+  - port: 21047
+    protocol: tcp
+    source: 10.6.48.214
+    comment: Rapid7 Scan Assistant
+
+  - port: 2181
+    protocol: tcp
+    source: 172.20.80.0/22
```
tpendragon: Do we need to be port specific? Can the generic rule just be allow 172.20.80.0/22?

tpendragon (changes requested):
My only question here is can we be less specific. We assumed the rule before was "if you're from the private network you can talk on the private network", can we just enforce that?

roles/firewall/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
...
+allowed_admin_cidrs:
+  - 10.249.64.0/18
+  - 10.249.0.0/18
+  - 128.112.0.0/16
+  - 172.20.80.0/22
+  - 172.20.95.0/24
+  - 172.20.192.0/19
+  - 172.24.0.0/19
+  # Required for local QEMU user networking.
+  - 10.0.2.0/24
```
kayiwa, line 11: we don't need this

roles/firewall/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
...
+ssh_allowed_cidrs:
+  - 10.249.64.0/18
+  - 10.249.0.0/18
+  - 128.112.0.0/16
+  - 172.20.80.0/22
+  - 172.20.95.0/24
+  - 172.20.192.0/19
+  - 172.24.0.0/19
+  # Required for local QEMU user networking.
+  - 10.0.2.0/24
```
kayiwa, line 45: this needs to go

roles/firewall/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
...
+fail2ban_ignoreips:
+  - 127.0.0.1/8
+  - ::1
+  - 10.249.64.0/18
+  - 10.249.0.0/18
+  - 128.112.0.0/16
+  - 172.20.95.0/24
+  - 172.20.192.0/19
+  - 172.24.0.0/19
+  - 10.0.2.0/24
```
kayiwa, line 60: this can go?

tpendragon (approved):

PR #7451, opened 2026-08-17
[CircleCI/Nomad] Reduce deployer resources.

tpendragon:
Not as required since Tigerdata has its own workers now.


kayiwa (approved):
OK! Thanks @tpendragon

PR #7456, opened 2026-08-18
[Tigerdata/Nomad] Tigerdata CI on Docker instead of Podman

tpendragon:
This runs on docker. One REALLY nice thing about this is it really simplifies running commands in CircleCI.


kayiwa (approved):
OK! Thanks @tpendragon

PR #7458, opened 2026-08-19
[Nomad/DaviesProject/Milberg/Cicognara] Host all the static sites from a single job. 

tpendragon:
This simultaneously migrates all the Nomad static sites to Docker (instead of podman), updates nginx for all of them, makes it so we don't have to build Docker images for them, and moves management of it to princeton_ansible instead of their code repository. It also locks the containers up - its root user is a nonsense user on the host, the user it runs as is non-root, the root file system is read-only, and it's not allowed to mess with CAP privileges.

The job registers the same services that nginxplus was looking for, so no nginx config changes necessary.


kayiwa (approved):
NICE! Thanks @tpendragon

PR #7460, opened 2026-08-19
Snow gitlab tls two

kayiwa:
Harden our gitlab staging instance. related to this sNow ticket. https://princeton.service-now.com/now/nav/ui/classic/params/target/incident.do%3Fsys_id%3Db4bce21dc37a039cfd677f75e0013101


roles/gitlab/tasks/main.yml
```diff
@@ -89,20 +88,15 @@
...
-    --cert-name {{ inventory_hostname }}
-  changed_when: false
-  when: running_on_server | default(false)
+# Install a snap with classic confinement
+- name: Gitlab | Install certbot with option --classic
+  ansible.builtin.command: snap install --classic certbot
+  become: true
+  when: running_on_server
+
+- name: Gitlab | update acme certificates for {{ domain_name }}
```
acozine, line 97: Does this mean that GitLab will have certs on the VMs themselves, as well as on the LBs? If so, we should add an entry to the "special certs" list on pul-it-handbook.

roles/nginxplus/files/conf/http/dev/gitlab_staging.conf
```diff
@@ -1,9 +1,62 @@
...
+
+limit_req_zone $gitlab_staging_auth_limit_key
+               zone=gitlab_staging_auth:10m
+               rate=30r/m;
+
+#
+# Log $uri rather than $request_uri because $request_uri includes
+# query-string arguments. GitLab APIs can accept authentication tokens
+# in query parameters, so omitting arguments reduces the risk of
+# credentials being written to access logs.
```
acozine, line 29: That's a nice feature. Are there other configs we'd want to add this to?

acozine (approved):
Thanks for extending the security to our staging env. I see a couple of minor questions, but I'm happy to merge.

PR #7467, opened 2026-08-20
Migrate PCDM.org to Docker

tpendragon:
This moves over the pcdm-staging/prod into the new static container.

It also fixes up the prod config for static sites to be closer to staging, to get it ready for sending logs.


kayiwa (approved):
🎉 thanks @tpendragon

PR #7471, opened 2026-08-21
[Nomad] Improve logs sent to Signoz, send logs for Production

tpendragon:
After this, anything with the `logger` tag should send logs to Signoz.

Also there's an OTEL endpoint set up on all nodes, noticed that Traefik logs were getting sent thanks to that.


kayiwa (approved):
OK! thanks @tpendragon

PR #7472, opened 2026-08-21
[Nomad] Configure firewall as part of the play.

tpendragon:
Some of the nodes weren't connecting to each other for Elixir networking
because of UFW, this fixed that.


kayiwa (commented):
🔥 thanks @tpendragon

kayiwa (approved):

PR #7474, opened 2026-08-21
Secure Docker w/ UFW

tpendragon:
Unfortunately Docker punches through UFW without this: https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-5a-be-careful-when-mapping-container-ports-to-the-host-with-firewalls-like-ufw

Before this change I could query the port from my computer while on VPN, after this change I no longer can.


kayiwa (approved):
🔒  thanks @tpendragon

PR #7477, opened 2026-08-24
Add firewall to the Common Only playbook.

tpendragon:
We should run this on a schedule.

Done as part of Ansible Open House


kayiwa (approved):
OK. Thanks

PR #7481, opened 2026-08-25
Nomad should only be accessible on campus.

tpendragon:
We'll open up two paths generally later.


kayiwa (approved):
OK! thanks @tpendragon

PR #7482, opened 2026-08-25
Update bibdata-staging allowlist to include catalog-dedupe1 and catalog-dedupe2

christinach:
Add catalog-dedupe1 and catalog-dedupe2 in the
bibdata-staging allowlist


christinach:
I ran the playbook on bibdata-staging and now I'm able to curl bibdata-staging patron url.

sandbergja (approved):
Thanks, @christinach ! 🐬

PR #7488, opened 2026-08-27
[Allsearch api] ship application logs to signoz

sandbergja:
We previously were shipping only the nginx logs.

I ran this today on allsearch-api-prod2.princeton.edu, and its app logs are displaying.  They are not yet displaying for allsearch-api-prod1.princeton.edu, which I have not run this on.

It could be nice to filter these so we only ship logs with severity WARN and above to signoz, but I don't know how to do that yet!


christinach (approved):
Thanks @sandbergja 🍄🍄

PR #7493, opened 2026-08-28
Use consistent log type names, and don't bring in extra log files

### LINKED ISSUE #7490: Name the log_type consistently across application configs
We don't want the application name in the log type; that's the service name and we can already filter on that.

log types should be:
- error
- access
- app-log

maybe something else will come up as we figure these out, but we can deal with outliers later

Image

### LINKED ISSUE #7491: Don't send gz files to signoz
We really only need to send the primary log file; nothing that has been rotated needs to be sent because those should be getting sent when they are first written.

hackartisan:
fixes #7490
fixes #7491

### DIFF
--- a/group_vars/allsearch_api/production.yml
+++ b/group_vars/allsearch_api/production.yml
@@ -75,7 +75,7 @@ otel_receivers:
       http:
         endpoint: 127.0.0.1:4318
   filelog/nginx-access:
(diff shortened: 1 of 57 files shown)


kayiwa (approved):
OK! Thanks @hackartisan 🧵

PR #7549, opened 2026-09-01
[nginxplus] log upstream timing metrics as numbers for analysis

kayiwa:
allow our telemetry to display upstream response time

This PR adds values to our logs that we get from [the otel module](https://docs.nginx.com/nginx/admin-guide/dynamic-modules/opentelemetry/). Then it maps them to include the units in the variable names, so we know what we're looking at when they come through to signoz


bess (approved):
Thank you, @kayiwa !

PR #7561, opened 2026-09-01
manage python dependencies with uv so renovate can update them

kayiwa:
renovate cannot maintain hard-pinned flat requirements. use uv instead


bess (approved):
Thank you, @kayiwa 🎉

PR #7593, opened 2026-09-02
[signoz] add clickhouse fsgroupchangepolicy to signoz

kayiwa:
Configure ClickHouse with fsGroupChangePolicy=OnRootMismatch to avoid recursively resetting ownership across the large local ClickHouse volume during pod restarts and node maintenance.


hackartisan (approved):
needs to be rebased? but looks good.

hackartisan:
doesn't look like it needs a rebase after all

PR #7606, opened 2026-09-02
fix molecule tests failing to find shared roles (unbreak main)

### LINKED ISSUE #7604: Fix  Molecule CI syntax failure after migration to uv
CI testing for the `lib_jobs` (possibly all roles) Ansible role fails during the Molecule syntax-check phase when invoked through uv

The available CI output does not include the underlying Ansible error, making it difficult to determine whether the failure is caused by a missing dependency, role resolution, environment configuration, or a version change introduced by the uv lockfile.

kayiwa:
Every role's test run broke because the newest Molecule (added by renovate) no longer adds the repository's shared roles directory to Ansible's role search path, so any test playbook referencing a sibling role failed its syntax check. Set the search path explicitly for CI, the test runner, and local shells.

https://github.com/ansible/molecule/issues/4391

closes #7604

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -21,7 +21,7 @@
     "UV_PROJECT_ENVIRONMENT":      "$DEVBOX_PROJECT_ROOT/.venv",
     "UV_PYTHON":                   "3.11",
     "UV_PYTHON_PREFERENCE":        "only-system",
-    "ANSIBLE_ROLES_PATH":          "$DEVBOX_PROJECT_ROOT/roles",
+    "ANSIBLE_ROLES_PATH":          "$DEVBOX_PROJECT_ROOT/external_roles:$DEVBOX_PROJECT_ROOT/roles",
     "ANSIBLE_VAULT_PASSWORD_FILE": "$DEVBOX_PROJECT_ROOT/bin/lastpass-ansible",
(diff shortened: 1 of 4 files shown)


run_molecule.py
```diff
@@ -55,7 +55,16 @@ def roles_dependant_on_changed_roles():
     return out_roles
 
 
+def roles_path():
```
tpendragon, line 58: What's this for?

devbox.json
```diff
@@ -21,6 +21,7 @@
     "UV_PROJECT_ENVIRONMENT":      "$DEVBOX_PROJECT_ROOT/.venv",
     "UV_PYTHON":                   "3.11",
     "UV_PYTHON_PREFERENCE":        "only-system",
+    "ANSIBLE_ROLES_PATH":          "$DEVBOX_PROJECT_ROOT/roles",
```
tpendragon: Oh I saw people saying if you did this that it'd install galaxy roles there. Is that fine?

tpendragon (changes requested):

run_molecule.py
kayiwa: It sets the role search path so a scenario can resolve sibling roles.  roles/lib_jobs 's converge playbook
has  - role: samba , which lives in  roles/samba .

Molecule ≤6 added the repo root to the search path itself, in  provisioner/ansible.py :


  util.abs_path(os.path.join(self._config.project_directory, os.path.pardir)),

devbox.json
kayiwa: Oh... I didn't think of that. :-(

run_molecule.py
tpendragon: But doesn't the env variable handle this so we don't gotta add code?

kayiwa:
superseded by https://github.com/pulibrary/princeton_ansible/pull/7611

PR #7611, opened 2026-09-03
Fix: make role tests find shared roles again

### LINKED ISSUE #7604: Fix  Molecule CI syntax failure after migration to uv

kayiwa:
Test runs for every role broke because the current Molecule release no longer adds the repository's shared roles directory to the search path Ansible uses to resolve roles by name, and it replaces the repository configuration with its own generated file

https://github.com/ansible/molecule/issues/4391

closes #7604

### DIFF
--- a/roles/ad_join/molecule/default/molecule.yml
+++ b/roles/ad_join/molecule/default/molecule.yml
@@ -33,6 +33,10 @@ platforms:
     pre_build_image: true
 provisioner:
   name: ansible
+  # Molecule no longer puts the repository's roles directory on the role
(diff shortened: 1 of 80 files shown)


christinach (approved):
Thanks @kayiwa 🦒

PR #7612, opened 2026-09-03
Signoz - ship QA services

kayiwa:
we were not shipping our QA services. 

related to #7197


bess (dismissed):
Discussed as a team 
* We will not ship QA telemetry at this time
* However, changes to the playbook which add the otel tags should be resubmitted in a separate PR 
Thank you!

hackartisan (approved):

PR #7614, opened 2026-09-03
updated staging nginx config files to remove the restrict.conf template.

### LINKED ISSUE #7192: [nginxplus] Remove restrict.conf, we no longer need it
Our staging are now all served from the dev/staging load balancers. And the dev/staging load balancers have IPs in the private network. This means the individual site config files do not need to restrict traffic by including the `restrict.conf` template, which lists IP ranges that should be allowed access to staging sites. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

Our nginxplus configuration should be as simple as possible.

### Acceptance criteria
We only use `restrict.conf` if needed.

 - [x] we remove the lines that include `restrict.conf` from the config for one staging site, push the new config to the dev/staging load balancers, and confirm that we can still only reach that site from campus IPs
 - [x] we audit the staging sites to make sure they are all running on the dev/staging LBs and move any that are still running on the production LBs
 - [ ] we remove the lines including `restrict.conf` from all staging config files
 - [ ] we audit the production sites to make sure none of them use the `restrict.conf` file
 - [ ] if it's completely unused, we delete `restrict.conf` entirely

aruiz1789:
Removed file restric.conf, updated all configs in staging not to reference the file.
This is related to moving the hosts to the private network and dev LB.
closes #7192


acozine (approved):
Looks good, thanks @aruiz1789

PR #7616, opened 2026-09-03
Certbot playbook - use nginx instead of standalone when creating/revoking certs

kayiwa:
Use nginx plugin. We were getting 

```sh
sudo certbot certonly --standalone --non-interactive --agree-tos --email lsupport@princeton.edu --server https://acme-us.certinext.io/v1/directory --eab-kid redacted --eab-hmac-key redacted --domain nomad-ingress-staging.lib.princeton.edu --cert-name nomad-ingress-staging.lib Saving debug log to /var/log/letsencrypt/letsencrypt.log
Requesting a certificate for nomad-ingress-staging.lib.princeton.edu
Could not bind TCP port 80 because it is already in use by another process on this system (such as a web server). Please stop the program in question and then try again.
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
```

this fixes that


playbooks/utils/incommon_certbot_revoke.yml
```diff
@@ -1,53 +0,0 @@
----
-# you MUST run this playbook on a single host with '--limit' for example `ansible-playbook -e domain_name=orcid-staging --limit adc-dev2.princeton.edu playbooks/utils/incommon_certbot_revoke.yml`
-# to run in production, you must add '-e runtime_env=production'
-# Run the playbook on the inactive load balancer first
-# if the 'nginx -t' check fails, we probably didn't delete an obsolete config file
-# run 'playbooks/nginx.yml' (or nginx_staging.yml) following the comments for removing obsolete config
-
-- name: revoke incommon acme for {{ domain_name }}
```
acozine, line 8: This PR replaces the two of the three `incommon_*` playbooks with equivalent `certinext_*` playbooks. But there's no `certinext_incommon_certbot_revoke.yml` playbook, even though the PR deletes the `incommon_certbot_revoke.yml` one. Is that on purpose? Do we have another way to revoke certs?

acozine (changes requested):
One concern, see below.

I'm also generally curious why we're renaming the playbooks - both sectigo and certinext use incommon, right? What's the advantage of renaming? Is it for auto-complete somehow?

playbooks/utils/incommon_certbot_revoke.yml
kayiwa: incommon was always there. We were using both for a little bit and lazy.

acozine: We looked at this today - the revoke playbook was added in #7223, so we'll still have a revoke playbook once this goes in.

acozine: But . ..  the certinext revoke playbook is still using `standalone` instead of `nginx` - do we need to update it?

Issue #7494, opened 2026-08-28
Ship application logs consistently to signoz


bess:
### What maintenance needs to be done?
All applications that send logs to signoz need to:
* Send nginx logs
  * Probably this is set up already!  
* Send application logs
  * Maybe you need a separate .json formatted log file, like [this](https://github.com/pulibrary/pdc_discovery/commit/0db928bdc3e7c5ed11cc359d99fa76667e6ad3b3)  
  * And then you tell signoz to process that log file, like [this](https://github.com/pulibrary/princeton_ansible/blob/5fe77d2197c3c7cd6af6d70f5ef23db8b4d3a964/group_vars/pdc_discovery/production.yml#L120-L121)

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
In order to know across the department when applications are experiencing problems with solr, we need to report all of the application errors to the same place. Nginx web logs are not sufficient to know about solr connection problems that an application is experiencing.

### Acceptance criteria
Each team needs to ensure that all of our applications are sending application logs, not just nginx logs, to signoz. Your friendly neighborhood solr performance tiger team has made a start on this work:

 - [ ] DLS
   - [x] figgy
   - [x] pulmap
   - [x] digital collections
   - [x] pulfalight
   - [x] dpul
 - [ ] DACS
   - [x] allsearch-api
 - [ ] RDSS
   - [x] pdc_discovery
   - [x] pdc_describe
   - [x] tigerdata-app
   - [x] mflux-ci1
   - [x] orcid

### Implementation notes
There are lots of ways to send application logs. We prefer json formatted logs, because otherwise multi-line stack traces dirty our data.

  * An example from [allsearch-api](https://github.com/pulibrary/princeton_ansible/pull/7488), using semantic logger
  * An example from [pdc_discovery](https://github.com/pulibrary/princeton_ansible/pull/7492), using [rails_semantic_logger](https://github.com/pulibrary/pdc_discovery/commit/0db928bdc3e7c5ed11cc359d99fa76667e6ad3b3)

PR #7620, opened 2026-09-03
[zookeeper] make zookeeper write json logs

kayiwa:
move from plaintext to json

advances #7494


sandbergja (approved):
This is great, thanks @kayiwa ! 🦒

PR #7622, opened 2026-09-04
[ezproxy] send ezproxy logs to signoz as structured json

kayiwa:
migrate logs from common log format to json

advances #7494


VickieKarasic (approved):
Thanks @kayiwa!

PR #7623, opened 2026-09-04
[otel-collector] stop re-downloading the collector tarball on every run

### LINKED ISSUE #7618: otel role should only download tarball if necessary
The otel_collector role fails sometimes on the [task to download the otel-contrib tarball](https://github.com/pulibrary/princeton_ansible/blob/97b4d955834f39fce1c0e750c3a12717c9e3b229/roles/otel_collector/tasks/main.yml#L16), usually becasue the server is busy. This happens even on servers where the file is already present in `/tmp`.

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

We should streamline our automation and eliminate failures that interrupt longer playbooks. Ultimately we want to run the otel_collector role as part of our standard maintenance, so it should be as efficient as possible.

### Acceptance criteria
 - [ ] if the otel-contrib tarball is present in the destination directory, we should skip downloading it again

The get_url module does not support a `creates` parameter, but the [documentation](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/get_url_module.html#notes) suggests that by using a file name for `dest` and keeping `force` set to `false` (which is the default), we should get the behavior we want. We can probably streamline the variables in this role.

kayiwa:
long playbooks were failing in part because github but also this sorta wasn't helping github

closes #7618

### DIFF
--- a/roles/otel_collector/README.md
+++ b/roles/otel_collector/README.md
@@ -6,6 +6,9 @@ You define receivers/processors/exporters/pipelines in `group_vars` or `host_var
 ## What it does
 
 - Downloads and unpacks a specific `otelcol-contrib` version to `/opt/otelcol`.
+- Keeps the release tarball under `/opt/otelcol/releases` and skips the download
+  entirely when that file is already there, so re-running the role does not
+  depend on reaching the internet.
 - Renders `config.yaml` from your vars (no hardcoding).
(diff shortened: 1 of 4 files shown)


christinach (approved):
thanks @kayiwa 🦒

PR #7625, opened 2026-09-04
[ezproxy] report ezproxy rule trips to Signoz

kayiwa:
- poll the EZproxy security database for newly tripped rules
- ship structured security events through the OpenTelemetry Collector
- map block and log actions to error and warning severities
- rotate the security trip log and document the alerting workflow


VickieKarasic (approved):
This looks great @kayiwa - thank you!

PR #7627, opened 2026-09-04
adds tip for deleting devbox and starting over

acozine:
We kept asking on slack "how do I delete devbox and start over?"

This PR changes a section title from "Troubleshooting Setup" to just "Troubleshooting" and adds a section with instructions for deleting devbox and starting over.

We may want this somewhere else instead, or in addition, but I wanted to get it out of my brain for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ devbox run init
 If your DevBox environment is acting wonky, or you're unsure if it might be causing some weird behavior, delete the local virtual environment and start over:
 
 ```bash
+# Delete the old python venv if you still have one
+rm -rf .venv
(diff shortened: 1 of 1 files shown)


kayiwa (dismissed):
can we add removing the python virtual environment. `.venv` with `rm -rf .venv`

hackartisan (approved):
Thank you!

acozine:
> can we add removing the python virtual environment. `.venv` with `rm -rf .venv`

Good idea, thanks. Added in 2cf1893a2750beaef9910d3e3f5dbebc2c7be5d7.

PR #7629, opened 2026-09-04
[dacs] Parse severity from semantic logger log lines for signoz

sandbergja:
Semantic logger includes the severity like so in its json:

```
{"level":"info"...
```

This adds this level to the otel severity fields, so that we can use it as a filter in signoz.

Copied more or less from the solr otelcol config


christinach (approved):
thanks @sandbergja 🍄

PR #7633, opened 2026-09-04
Set the connect timeout to the ceiling, it's more explicit

hackartisan:
see https://www.netdata.cloud/guides/nginx/nginx-upstream-timed-out/


kayiwa (approved):
thanks @hackartisan 🧵

PR #7638, opened 2026-09-07
[Figgy] Remove restrict.conf from stream proxy.

tpendragon:
The file isn't on the dev lb anymore.


kayiwa (approved):
OK! thanks @tpendragon

PR #7641, opened 2026-09-08
[orangelight] Rotate the Rails production log daily

kayiwa:
The application log at /opt/orangelight/current/log/production.log was   never rotated and could grow until it filled the disk. It now rotates   daily with the same retention as our other logs, truncating in place so   the running app keeps writing without a restart. Also fixes the reload   step, which previously never detected config changes.

related #7494


sandbergja (approved):
Thanks, @kayiwa ! 🦒

PR #7642, opened 2026-09-08
[orangelight] Remove datadog

sandbergja:
We are using signoz instead

I removed the agent from the boxes with the `remove_datadog` playbook and manually deleted `/opt/datadog-agent`


regineheberlein (approved):
Thanks @sandbergja !

PR #7648, opened 2026-09-08
[ruby_app] Rotate application logs for every Ruby app

### LINKED ISSUE #7645: Rails and Hanami applications should have their application logs rotated
### Expected behavior

Application logs for our ruby applications (Hanami, Rails, and Rack) are rotated on a daily basis (or more often if necessary).

### Actual behavior

Some applications do (e.g. Figgy, Orangelight, Pulmap), others don't (Orangelight prior to #7641)

### Error messages

Disk full alert

### Steps to replicate

Check the logs and confirm that the files have been rotated (e.g. production.log.[date].gz, etc.)

### Impact of this bug

The disk fills up and brings down our app servers :-( 

* Check if this is a prod vs. staging thing?  For example, pulfalight-staging does not rotate while pulfalight-prod does
* Make sure that logs that have file extenstion `.json` also get rotated (for example pdc-discovery logs), not just `*.log`

kayiwa:
Rotation now lives in the shared Ruby app role so every app gets it, covering both plain text and structured JSON logs, and   truncating in place so running apps keep logging without a restart.

closes #7645


kayiwa:
```bash
pulsys@pdc-discovery-prod1:~$ ls -al /opt/pdc_discovery/current/log/
total 2714324
drwxrwxr-x 2 deploy deploy       4096 Aug 28 13:40 .
drwxr-xr-x 7 deploy deploy       4096 Apr  7 17:53 ..
-rw-rw-r-- 1 deploy deploy  777127869 Sep  8 20:43 production.json
-rw-rw-r-- 1 deploy deploy 2002318504 Sep  8 20:43 production.log
pulsys@pdc-discovery-prod1:~$ date
Tue Sep  8 08:43:29 PM UTC 2026
pulsys@pdc-discovery-prod1:~$ exit
logout
Shared connection to pdc-discovery-prod1.princeton.edu closed.
```

after playbook run 

```bash
pulsys@pdc-discovery-prod1:~$ ls -al /opt/pdc_discovery/current/log/
total 2714888
drwxrwxr-x 2 deploy deploy       4096 Sep  8 20:49 .
drwxr-xr-x 7 deploy deploy       4096 Apr  7 17:53 ..
-rw-rw-r-- 1 deploy deploy       2231 Sep  8 20:49 production.json
-rw-rw-r-- 1 deploy deploy  777399235 Sep  8 20:49 production.json-20260908-1788900557
-rw-rw-r-- 1 deploy deploy       8225 Sep  8 20:49 production.log
-rw-rw-r-- 1 deploy deploy 2002611663 Sep  8 20:49 production.log-20260908-1788900557
pulsys@pdc-discovery-prod1:~$ date
Tue Sep  8 08:49:42 PM UTC 2026
pulsys@pdc-discovery-prod1:~$
```

bess (approved):

roles/ruby_app/defaults/main.yml
```diff
@@ -18,18 +18,28 @@ app_otel_env: "{{ ruby_app_otel_env | default([]) }}"
...
+# variable name on purpose: the shared logrotate_rules is already claimed
+# further up the dependency chain, so rules placed there never reach the hosts
+# that run an application.
+ruby_app_logrotate_rules:
+  - name: "{{ app_name }}"
     paths:
       - "/opt/{{ app_directory }}/current/log/*.log"
+      # Apps that emit structured logs for our observability stack write a
+      # second file alongside the plain text one, and it grows just as fast.
+      - "/opt/{{ app_directory }}/current/log/*.json"
```
hackartisan, line 31: Several of the rules removed include *.log; what will happen to those?

hackartisan (changes requested):

hackartisan (commented):
Which apps are running hanami? Have we verified the logs get written to the same place?

roles/ruby_app/defaults/main.yml
kayiwa: Good catch to check. Nothing loses coverage: every removed rule pointed at `/opt/<app>/current/log/*.log` , which is exactly what the new shared rule matches, and both `approvals`  and  `lockers_and_study_spaces`  already depend on  ruby_app , so they get it automatically.

The rule is also named after the app, so it writes the very same file as before, `/etc/logrotate.d/approvals`  and  `/etc/logrotate.d/lockers_and_study_spaces` , not a generic `ruby_app `. I confirmed this by running the real dependency chain for both roles (on staging) and inspecting the result:

```
/etc/logrotate.d/approvals
    /opt/approvals/current/log/*.log
/etc/logrotate.d/lockers_and_study_spaces
    /opt/lockers_and_study_spaces/current/log/*.log
```

hackartisan: sorry, I must have overlooked that line! Thanks

hackartisan (approved):

PR #7650, opened 2026-09-08
[Figgy] Highly Available Staging Redis in Nomad

tpendragon:
The goal is to add highly available redis to Nomad.

Work towards https://github.com/pulibrary/figgy/issues/7359

I recommend reading https://github.com/pulibrary/princeton_ansible/blob/12b0c4926576874f2b27dff5333350b0ba272409/nomad/redis/README.md

Redis is a nightmare. It works by launching three nodes, and then another three nodes to watch those redis nodes, then those "sentinel" nodes just override the config file in both redis & sentinel. There's no real way to separate "runtime" configuration, so the job has to seed a basic config and then never touch it again.

There's some tricky business in here to make sure when it boots up the first time everything points at one primary, so they discover each other.

Any config that might change sometimes (like password) is passed in through command line arguments, because those get written to the config file on boot-up.

Things I tested so far:

1. Delete the whole infrastructure and rebuild from scratch, make sure it spins up.
2. Delete the primary, make sure the others recognize it.
3. Bring back the primary, make sure it becomes a replica.
4. Adding another node.
5. That Figgy can connect to it. https://github.com/pulibrary/figgy/pull/7358
6. Copying data from another redis: (run as pulsys on dpul-collections-staging1.lib.princeton.edu) `sudo docker run --rm --network host --userns host riotx/riotx replicate redis://figgy-web-staging1.princeton.edu:6379/0 redis://172.20.80.36:6379/0 --target-pass <passhere>`
7. Hard reboot the whole thing, make sure it comes together.

I did my best to document in the README all the various things folks would need to use this, and what I'd done to test it.


hackartisan (commented):
does wiping the cluster lose all the data?

tpendragon:
> does wiping the cluster lose all the data?

Turning the job off does not wipe the data, purging the job...I think doesn't, I'll test that.

Doing the step that's "stop the job and then delete every volume" on every node that's in the README does.

tpendragon:
Confirmed. Purging the job without wiping the directories does not lose the data. Re-deploying lets it pick up all the old data.

tpendragon:
If you wiped the whole Nomad VM cluster, the data would be lost. If you did one node at a time over time and make sure the services are healthy before doing the next, it would not lose data.

eliotjordan (approved):

PR #7658, opened 2026-09-10
[ssh_ca_trust] let named service accounts keep password logins

kayiwa:
The Veeam backup appliance signs in to the Proxmox nodes with a local   password and cannot present an SSH certificate. Hosts that trust the SSH   certificate authority turn passwords off for everyone, so the only way to   get backups working was to re-enable password logins for the whole host,   which is far more than we want to open up.


aruiz1789 (approved):
Thanks @kayiwa

PR #7659, opened 2026-09-10
[Nomad] Looking for feedback - deploy as part of playbook run?

tpendragon:
The deploys for stateful services don't work without the volumes being built first, which happens in the playbook. Maybe deploy while we're at it?

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -10,7 +10,8 @@
     "podman@5.8.6",
     "ruby@4.0.6",
     "ruby_3_4@3.4.8",
-    "step-cli@0.30.6"
+    "step-cli@0.30.6",
+    "nomad_2_0@latest"
   ],
   "env": {
     "PYTHON_VERSION":              "3.13",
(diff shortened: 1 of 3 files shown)


hackartisan (commented):
Are there documentation changes that would go with this PR?

It seems like it will sometimes make 2 steps into 1 step; I'm trying to see a potential downside; I think for these kinds of services building the machine and deploying the application are usually done in the same playbook, so it seems to align with convention.

tpendragon:
> Are there documentation changes that would go with this PR?

Good question! Probably...? I dunno where exactly. I also don't know if still having the HCL in the `nomad` directory makes sense, but I dunno where else to put it. I don't wanna write a whole role for it, when the playbook's working so well.

What kinda docs would you like to see?

hackartisan:
@tpendragon do you want to chat about this some time?

tpendragon:
@hackartisan Sure. So far I've heard from you and Francis "this is obviously better", I just haven't prioritized looking into where I should update docs.

bess (approved):
+1 for this change

Thank you!

PR #7661, opened 2026-09-11
Load the dev environment automatically on entering the repo

kayiwa:
Entering the project directory now activates everything automatically and re-syncs Python, Ruby, and Ansible dependencies only when the lock files change, so the common case stays fast. Manual activation still works for anyone who prefers it.


bin/devbox-bootstrap
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
tpendragon, line 1: I guess I'm curious why a bunch of this wouldn't just happen as part of the init_hook?

bin/devbox-bootstrap
```diff
@@ -0,0 +1,63 @@
...
+# Called automatically by .envrc (direnv) every time you enter the project
+# directory, and available as `devbox run bootstrap`. Work is skipped when the
+# dependency lock files have not changed since the last successful run.
+
+set -uo pipefail
+
+root="${DEVBOX_PROJECT_ROOT:-$(cd "$(dirname "$0")/.." && pwd)}"
+cd "$root" || exit 1
+
+state_dir=".devbox/bootstrap"
```
tpendragon, line 13: Is all this state stuff necessary?

bin/first-time-setup.sh
```diff
@@ -4,6 +4,18 @@ echo "==================================="
 echo "First Time Setup - Devbox Migration"
 echo "==================================="
 
+# Figure out which shell profile to update
```
tpendragon, line 7: Why'd this move?

bin/first-time-setup.sh
```diff
@@ -43,6 +46,26 @@ echo ""
...
 devbox install
 
+# Install direnv so this environment loads automatically when you cd here
+if ! command -v direnv &>/dev/null; then
+  echo ""
+  echo "Installing direnv..."
+  devbox global add direnv
+fi
+
+if ! grep -q "direnv hook" "$SHELL_PROFILE" 2>/dev/null; then
```
tpendragon, line 56: I don't usually like it, personally, when setup steps put stuff in my shell config. Can we just set up the `envrc` and info to the docs, or a separate file about installing it? For instance in my setup all my normal customizations to my shell go into a `~/.zshrc.local` file.

bin/first-time-setup.sh
```diff
@@ -43,6 +46,26 @@ echo ""
...
+  echo "" >>"$SHELL_PROFILE"
+  echo "# direnv" >>"$SHELL_PROFILE"
+  echo "eval \"\$(direnv hook $SHELL_NAME)\"" >>"$SHELL_PROFILE"
+  echo "Added the direnv shell hook to $SHELL_PROFILE"
+fi
+
+# Approve this project's .envrc so direnv is allowed to load it
+if command -v direnv &>/dev/null; then
+  (cd "$(dirname "$0")/.." && direnv allow) &&
+    echo "Approved .envrc for this project"
```
tpendragon, line 66: I definitely would prefer this to be optional. It warns you when you come in.

I've found that `direnv` makes navigating to the directory just to check the source code or whatever quite slow actually, I've turned it off on most of my projects in favor of explicit `devbox shell`, but recognize that's a personal decision.

.envrc
```diff
@@ -0,0 +1,28 @@
...
+  exit 0
+fi
+
+use_devbox() {
+  eval "$(devbox shellenv --init-hook --install --no-refresh-alias)"
+  watch_file "$DEVBOX_PROJECT_ROOT/devbox.json" "$DEVBOX_PROJECT_ROOT/devbox.lock"
+}
+
+# The .venv here is managed by uv, not by Devbox's Python plugin. This marker
+# answers the plugin's "overwrite it?" prompt once, so loading never blocks.
```
tpendragon, line 20: Yeah, speaking of which, why is this? Devbox has UV and virtualenv support, why do we have a separate virtualenv?

tpendragon (commented):
I have a few questions and comments as a regular direnv/devbox user, but this isn't my every day repo, so they're not blocking.

kayiwa:
Closed this PR as it felt geared to how I work and having to justify how I work was both a burden and an imposition on others if seems.

PR #7666, opened 2026-09-14
[beyla] stop re-downloading the Beyla release on every run

kayiwa:
similar to work we did with Otel Collector https://github.com/pulibrary/princeton_ansible/pull/7623

update version of Beyla to 3.35 which addresses a fundamental coexistence question when some services use an SDK while Beyla instruments others. The eBPF layer should continue an existing trace rather than silently create a competing trace lineage. It is not directly a Solr fix, but it affects Go-based gateways or services downstream from NGINX

https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3156


sandbergja (approved):
So cool -- we had a playbook fail because it couldn't download beyla just this morning. I'm excited for the new beyla version too.  Thanks, @kayiwa ! 🦒

PR #7679, opened 2026-09-16
[orangelight] Don't ship rotated compressed logs to signoz

sandbergja:
Before this commit, we inadvertently shipped compressed logs (e.g. a file with filename `staging.log-20260915-1789430400.gz`) to signoz, where they are not of much use.

This commit makes sure that we are only sending the current production.log or staging.log.

I ran this on staging today (15 September)


kayiwa (approved):

kayiwa (approved):
OK! Thanks @sandbergja 🍄

PR #7680, opened 2026-09-16
[postgresql18] Back up production databases nightly to cloud storage

kayiwa:
Postgres18 had no automated offsite backup. A nightly systemd timer now dumps every database to a Google Cloud Storage bucket mounted with gcsfuse, keeping 7 daily, 5 weekly, and 3 monthly copies plus the cluster roles and tablespaces needed for a full restore.

Dumps are written to local disk and moved into the bucket only once complete, so monitoring never sees a half-written file. A Checkmk check reports whether the day's backups arrived, warning separately when the per-database dumps are present but the roles dump is missing.

Backups must never disrupt the database they protect, so a dump is aborted if free space nears a floor rather than risk filling the disk that holds the data directory, a dump whose connection is dropped by a restart is retried, and any pending restart is applied before a dump starts. Enabled for production only.

Copied from solr9cloud role


carolyncole (approved):
@kayiwa seems like something we should do...

PR #7684, opened 2026-09-17
[postgresql] ensure service name vars load and template safely

### LINKED ISSUE #7683: postgres_service_name undefined in PostgreSQL backup service task
Running the PostgreSQL role fails on postgres18-prod1.lib.princeton.edu because the variable postgres_service_name is not defined when rendering the backup service template.

```text
[ERROR]: Task failed: 'postgres_service_name' is undefined

Task failed.
Origin: /Users/kayiwa/projects/pulibrary/princeton_ansible/roles/postgresql/tasks/backup_schedule.yml:36:3

34   when: running_on_server
35
36 - name: PostgreSQL | Create backup service
     ^ column 3

<<< caused by >>>

'postgres_service_name' is undefined
Origin: /Users/kayiwa/projects/pulibrary/princeton_ansible/roles/postgresql/templates/postgres-backup.service.j2

fatal: [postgres18-prod1.lib.princeton.edu]: FAILED! => {"changed": false, "msg": "Task failed: 'postgres_service_name' is undefined"}
```

## Steps to Reproduce
Run the postgresql role (or playbook that includes it) against postgres18-prod1.lib.princeton.edu

Task PostgreSQL | Create backup service in roles/postgresql/tasks/backup_schedule.yml fails during template rendering

## Expected Behavior
The task should render postgres-backup.service.j2 successfully and create the backup service.

Actual Behavior
Template rendering fails because postgres_service_name is undefined in the Ansible variable scope at the time backup_schedule.yml runs.

kayiwa:
The backup service task failed on postgres18-prod1 with 'postgres_service_name' is undefined when the play was limited to a subset of tags. The OS-specific vars file was being skipped, so the service/package name variables that downstream tasks and templates depend on were never loaded.

- Tag the OS-specific include_vars with `always` so the service and package path vars are loaded regardless of tag selection
- Default `postgres_service_name` to `postgresql` in the backup service template as a defensive fallback

Fixes: backup_schedule.yml failing at "PostgreSQL | Create backup service"

closes #7683


carolyncole (approved):
Thanks @kayiwa

PR #7686, opened 2026-09-17
[nginxplus] stop the loadbalancer from logging bogus file lookups

### LINKED ISSUE #7681: Reduce error.html/index.html log noise from the load balancer

We have many log entries in the load balancer's error.log that are similar to this:

```
"/var/local/www/default/error.html/index.html" is not found -- Not a directory
```

You can find examples in the signoz log explorer with the following query:

```
log_type in ['nginxplus_error'] AND errtext = 'Not a directory' AND body CONTAINS '/var/local/www/default/error.html'
```

We should address the underlying issue or otherwise remove these from the logs.

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

- Reduce the noise as we try to find useful information in the logs

### Acceptance criteria
 - [ ] We no longer get `"/var/local/www/default/error.html/index.html" is not found` messages in the logs
 - [ ] If this message is related to the 001-default.conf file, we also evaluate this file to see if we even need the 001-default config anymore

[This line seems suspicious](https://github.com/pulibrary/princeton_ansible/blob/658f8ffafb16f50e867f2c865a9109d3f107c7bd/roles/nginxplus/files/conf/http/001-default.conf#L64) -- `root` typically refers to a directory, but `/var/local/www/default/error.html/index.html` is not a directory. Also, I'm not sure what role the 001-default config plays, do we even need this config file?

kayiwa:
The catch-all server pointed its document root at an HTML file instead   of a directory, so every request without a matching host name (mostly   scanners hitting the IP address) made nginx search for a path that can   never exist and log "is not found -- Not a directory". Those entries   drowned out useful information in the error log.

Unmatched requests now serve the existing landing page from a real   directory, the catch-all is declared as the default server so its   purpose is explicit rather than a side effect of config load order, and   its HTTP-to-HTTPS redirect uses the requested host instead of a literal   name that does not resolve in production.

  Also removes a duplicate copy of this config that was overwritten on   every run, plus an unused HTTP status-text mapping, leaving one source   of truth.

closes #7681


roles/nginxplus/templates/etc_confd_default.conf.j2
```diff
@@ -1,66 +1,36 @@
...
-  507 'Insufficient Storage';
-  508 'Loop Detected';
-  510 'Not Extended';
-  511 'Network Authentication Required';
-  default 'Something is wrong';
-}
+# Catch-all server. Because this is the first config nginx loads, it answers
+# every request whose Host header matches none of the sites we publish, such as
+# scanners hitting the load balancer by IP address. It serves the static
+# "nothing to see here" landing page and never proxies to a backend.
```
sandbergja, line 5: Thanks for adding this documentation!

sandbergja (approved):
Thanks, @kayiwa ! 🦒

PR #7690, opened 2026-09-17
[mailcatcher] Don't attempt to install mailcatcher if install_mailcatcher=false

sandbergja:
We were missing some indentation!  We noticed that, prior to this commit, it attempted to install mailcatcher on a fresh catalog VM.


christinach (approved):
thanks @sandbergja 🍄🐬

PR #7691, opened 2026-09-17
[postgresql] exclude updating the apt cache from idempotence check

sandbergja:
This step is not idempotent, and the apt cache could totally have updates between molecule checks (as it did here: https://github.com/pulibrary/princeton_ansible/actions/runs/35277192243/job/105390532946?pr=7690)

Use the molecule-idempotence-notest tag so that we don't get a CI failure due to the apt packages changing.


kayiwa (approved):
OK! Thanks @sandbergja 🍄

PR #7692, opened 2026-09-17
[bind9] Make the local resolver IPv4-only and actually used

kayiwa:
Our network does not route IPv6, so named spent time waiting on AAAA   lookups to time out before falling back to A records, showing up as slow   resolution and intermittent resolution errors. Pin named to IPv4 for both   listening and outbound queries

The role previously only installed the package and pointed the host at the   systemd-resolved stub, so bind was never consulted. It now forwards to the   campus resolvers, caches answers, and points the host at itself, keeping   the campus resolvers as fallbacks so a named outage cannot take host DNS   down with it.


roles/bind9/defaults/main.yml
```diff
@@ -1,2 +1,36 @@
...
 running_on_server: false
+
+bind9_packages:
+  - bind9
+  - bind9-utils
+
+# Our network does not route IPv6. Keeping named on IPv4 avoids AAAA
+# lookups that can only time out before falling back to A records.
+bind9_ipv4_only: true
+bind9_named_options: "-u bind{{ ' -4' if bind9_ipv4_only else '' }}"
```
sandbergja, line 11: 👍🏻

roles/bind9/defaults/main.yml
```diff
@@ -1,2 +1,36 @@
...
+  - localhost
+bind9_recursion: true
+bind9_allow_recursion: "{{ bind9_allow_query }}"
+bind9_dnssec_validation: auto
+
+# Point the host at its own named instance, keeping the campus resolvers
+# as a fallback so a named outage cannot take DNS down with it.
+bind9_manage_resolv_conf: true
+bind9_resolver_address: 127.0.0.1
+bind9_resolv_nameservers: "{{ [bind9_resolver_address] + bind9_forwarders }}"
```
sandbergja, line 33: 👍🏻

roles/bind9/tasks/main.yml
```diff
@@ -1,35 +1,70 @@
...
+- name: Bind9 | Configure named
+  ansible.builtin.template:
+    src: named.conf.options.j2
+    dest: "{{ bind9_options_file }}"
+    owner: root
+    group: bind
+    mode: "0644"
+  become: true
+  notify:
+    - Restart bind
```
sandbergja, line 17: When `named` is restarting, will the box use the OIT nameservers (per `/etc/resolv.conf`) instead?

roles/bind9/tasks/main.yml
```diff
@@ -1,35 +1,70 @@
...
+    owner: root
+    group: root
+    mode: "0644"
+  become: true
+  when: bind9_legacy_defaults.stat.exists
+  notify:
+    - Restart bind
+
+- name: Bind9 | Verify named configuration is valid
+  ansible.builtin.command: named-checkconf
```
sandbergja, line 48: 😎

roles/bind9/tasks/resolver.yml
```diff
@@ -0,0 +1,25 @@
...
+  ansible.builtin.stat:
+    path: /etc/resolv.conf
+  register: bind9_resolv_conf
+
+# systemd-resolved regenerates the file it links to, so a symlink has to go
+# before we can pin the host to named.
+- name: Bind9 | Resolver | Drop the systemd-resolved symlink
+  ansible.builtin.file:
+    path: /etc/resolv.conf
+    state: absent
```
sandbergja: this seems kind of scary -- will we not be able to resolve DNS between this step and the next one?  Also, I'm curious how systemd-resolved what role `systemd-resolved` will play in the future -- will it still be caching things?  Any chance that it would clobber /etc/resolv.conf back to the way it likes it?

roles/bind9/defaults/main.yml
```diff
@@ -1,2 +1,36 @@
...
+bind9_resolv_search:
+  "{{ [ansible_facts.domain] if ansible_facts.domain | default('') else [] }}"
```
sandbergja: What is the value of`ansible_facts.domain` -- `princeton.edu` or a subdomain?  Our current resolv.conf has `search princeton.edu`, which is kind of nice

sandbergja (commented):
Thanks, @kayiwa ! 🦒  I'm so excited about this!  I had a few questions to improve my own understanding.

roles/bind9/tasks/resolver.yml
kayiwa: You were right, and I've removed that task. Good catch

roles/bind9/defaults/main.yml
kayiwa: This question found a bug. Thank you

roles/bind9/tasks/main.yml
kayiwa: Yes, and the failover is effectively instant. The reason is that nothing is listening on  127.0.0.1:53 , so the kernel returns an ICMP port-unreachable │ and the client gets  ECONNREFUSED  immediately.

hackartisan (commented):
Thank you!! I would love to have documentation of how long the cache is held and how to reset it if needed. Not a blocker for this PR, though.

roles/bind9/README.md
```diff
@@ -1,38 +1,157 @@
...
+See the [BIND cache settings reference](https://bind9.readthedocs.io/en/stable/reference.html#namedconf-statement-max-cache-ttl).
+Check `named -V` when troubleshooting a host running a different BIND version.
+
+### Inspect the remaining TTL
+
+Query the local resolver directly on the affected host (`dig` is provided
+by Ubuntu's `dnsutils` package):
+
+```bash
+dig @127.0.0.1 lib-solr-prod9.princeton.edu A +noall +answer
```
sandbergja, line 72: Thanks so much for these `dig` commands!

sandbergja (approved):
Thanks, @kayiwa ! 🦒

PR #7695, opened 2026-09-18
[solr] Use system default for number of gc threads

hackartisan:
We have been limiting gc to 4 threads on 8 core machines; java will by
default use threads equal to the number of cores. That seems like a good
default. Load testing revealed at worst no impact from this change.

Also explicitly set the default g1gc garbage collector, for clarity.

Remove old commented-out code

Pull gc-related config value out of large system config block and into
"solr_gc_tune" variable so its with other gc-related configs.


sandbergja (approved):
Thanks, @hackartisan ! 🧵 🍄

PR #7698, opened 2026-09-18
updates playbook docs for certinext playbooks

acozine:
@VickieKarasic and I ran the certinext playbooks and discovered that the docs in the comments were not copy-paste-able. 

This PR updates the playbooks to use the full playbook names and adds more detail to the revoke playbook.


sandbergja (approved):
Thanks, @acozine ! 🐴 ☕

PR #7700, opened 2026-09-18
[mysql] Back up MariaDB databases to Google Cloud Storage

kayiwa:
Production MariaDB had no offsite backups, so a lost server meant lost   data. This mirrors the arrangement already used for PostgreSQL: a GCS   bucket is mounted with gcsfuse and a nightly systemd timer dumps every   database into it, keeping 7 daily, 5 weekly and 3 monthly copies


christinach (approved):
Thanks @kayiwa 🦒 !

PR #7702, opened 2026-09-18
new certinext key that allows SAN certs

kayiwa:
use a certinext API key that allows use of SAN certificates.


christinach (approved):
Asante @kayiwa 🦒 !

PR #7705, opened 2026-09-20
[otel/beyla] version update

kayiwa:
Collector pod that does not terminate promptly can exceed Kubernetes’ grace period and be killed. That complicates MicroK8s rolling upgrades and increases the chance that queued telemetry does not drain normally. (we have experience with this in the past)
https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.161.0

and jvm-garbage collection duration metrics. If this works it nullifies the hard work we put into making this work on our dashboards but we learned things in the process. https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/3237


bess (approved):
Thanks, @kayiwa !

